PR #47, opened 2026-07-09
Scrum-883-fix-issue-45

### LINKED ISSUE #45: Dex not picking up id columns that contain word "key" during relationship inference
### What happened

Dex doesn't pick up id colums that contain the word "key".
The infer_relationships function should be more generic and customizable.

### Steps to reproduce

Try out on TPCH data in Snowflake samples.
Automatic relationship inference came back empty because it looks for _id-style column names and TPC-H uses _KEY suffixes, aka the real FK structure (ORDERS→CUSTOMER, LINEITEM→ORDERS/PART/SUPPLIER, SUPPLIER/CUSTOMER→NATION→REGION)

### Connector

Not connector-specific

### dex-core version

1.0.1

### Environment

Claude Code

### Relevant output

```shell

```

emnarahmouni-exmergo:
_fk_stem` only recognized `_id`-style column names, so warehouses that use `_key` conventions produced zero inferred relationships. This includes dimensional models (`customer_key` surrogate keys) and, as reported in #45, TPC-H's real FK structure (`O_CUSTKEY`, `L_ORDERKEY`, `N_REGIONKEY`, ...), which uses no separator at all before `KEY`.
- Generalized the suffix list to `("id", "key")` so a future convention is a one-line addition, not a rewrite of the matching logic.
- Added a new match branch for TPC-H's alias-prefixed FK names (`L_ORDERKEY` on LINEITEM referring to `O_ORDERKEY` on ORDERS) , the existing entity-name match can't see this since the FK is named after the *child's* table alias, not the parent's entity name. Guarded so it doesn't degrade to matching on a bare `id`/`key` suffix alone.

## Why bare `KEY` (but not bare `ID`) is accepted without a separator `ID` is deliberately skipped without a separator (`HOSTID`) because it's the tail of ordinary English words (`PAID`, `VALID`, `GRID`). `KEY` isn't a realistic false positive in warehouse column names, and TPC-H's own PK/FK convention concatenates it directly (`CUSTKEY`, not `CUST_KEY`).

image
after fix: 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- `explore relationships` only recognized `_id`-style foreign key columns, so
+  warehouses using a `_key` convention (dimensional surrogate keys, and
+  TPC-H's own FK structure: `O_CUSTKEY`, `L_ORDERKEY`, `N_REGIONKEY`, ...)
+  inferred zero joins. `_fk_stem` now recognizes `key` alongside `id`, and a
+  new alias-stripping match handles TPC-H's convention of naming a foreign
+  key after the child table's own alias rather than the parent's entity name
+  (`L_ORDERKEY` on `LINEITEM` referring to `O_ORDERKEY` on `ORDERS`). (#45)
+
 ## [1.0.1] - 2026-07-06
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (commented):
LGTM!

Can you add a summary of these changes to the CHANGELOG.md?
Under "Unreleased".

After this we can approve and merge

emnarahmouni-exmergo:
> add a summary

Done ✅

marcociav-exmergo (approved):
LGTM!

PR #57, opened 2026-07-10
SCRUM-895-Fix transform build surfacing dbt deprecation warnings

### LINKED ISSUE #50: `transform build` surfaces a dbt deprecation warning instead of the actual error
## Summary

When `dbt build` fails, the dex envelope's `errors[]` reports a dbt **deprecation warning** rather than the actual failure. The real cause is recoverable only by grepping `logs/dbt.log`.

This is a regression against the BigQuery field-feedback (2026-07-04) regression item 4, which recorded "Failure-path envelope: IMPROVED — the real message is in `errors`." That does not hold on the Snowflake/dbt path.

## Repro

1. Point the generated `profiles.yml` at a database that does not exist (e.g. `database: NOPE_MISSING_DB`).
2. Run `transform build --target dev --confirm --budget 60`.

## Actual

```json
{
  "status": "error",
  "data": {"success": false, "returncode": 2},
  "errors": ["dbt build failed: [WARNING][PropertyMovedToConfigDeprecation]: Deprecated functionality"]
}
```

The real cause, found only in `logs/dbt.log`:

```
Runtime Error
  Database error while listing schemas in database "NOPE_MISSING_DB"
  Database Error
    002043 (02000): SQL compilation error:
    Object does not exist, or operation cannot be performed.
```

## Expected

`errors[]` contains the `MainEncounteredError` payload (the "Object does not exist" message), not a `[WARNING]` line.

## Diagnosis

The extractor appears to take the first stderr line rather than dbt's `MainEncounteredError` event. Because dbt 1.11 emits deprecation warnings (`PropertyMovedToConfigDeprecation`, `MissingArgumentsPropertyInGenericTestDeprecation`) on every run of a normally-authored project, **the error channel is reliably poisoned** on any build failure.

## Suggested fix

Filter lines matching `[WARNING]` out of error extraction, and surface dbt's structured `MainEncounteredError` event (available in the JSON log stream, since dex already invokes dbt with `--log-format json`).

See also #55 (surfacing deprecations at plan time), which would remove the warnings that poison this channel in the first place.

---

**Environment**

emnarahmouni-exmergo:
Fixes #50: _collect_messages picked messages[0] in raw dbt log order, so a [WARNING] deprecation notice (logged during parsing) beat the actual failure (logged during execution) into errors[0] on every build failure.

Now deprecation-tagged lines sink to the end, and dbt's MainEncounteredError event (the real failure summary) is promoted to the front when present. Deprecations still surface in warnings, just no longer mistaken for the cause.

Before : 
Screenshot 2026-07-10 153457
After : 
Screenshot 2026-07-10 153526

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- `transform build` surfaced a dbt deprecation warning (`[WARNING]
+  PropertyMovedToConfigDeprecation`) as the failure cause instead of the real
+  error. dbt 1.11 logs these notices before the actual failure on every
+  normally-authored project, so the notice reliably won the `errors[0]` slot.
+  `_collect_messages` now promotes dbt's own `MainEncounteredError` event (the
+  structured summary of what actually failed) to the front and sinks
+  `[WARNING]`-tagged lines to `warnings` instead. (#50)
+
 ## [1.1.0] - 2026-07-09
 
 ### Added
--- a/packages/dex-core/src/exmergo_dex_core/transform/build.py
+++ b/packages/dex-core/src/exmergo_dex_core/transform/build.py
@@ -408,17 +408,30 @@ def _default_runner(
     return run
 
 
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (approved):
LGTM

PR #58, opened 2026-07-10
SCRUM-897-surface dbt deprecation warnings at plan time

### LINKED ISSUE #55: Surface dbt deprecation warnings at plan time (`transform plan` / `semantic define`)
## Summary

`transform plan` and `semantic define` run dbt's own parser as a validation gate, and reported `warnings: []`. The very same YAML then emitted deprecation warnings at `transform build`.

## Observed

Plan-time (both `transform plan` for staging/marts and `semantic define` for the semantic layer):
```
status: ok
warnings: []
```

Build-time, from the identical files:
```
[WARNING][MissingArgumentsPropertyInGenericTestDeprecation]: Deprecated
[WARNING][PropertyMovedToConfigDeprecation]: Deprecated functionality
[WARNING][DeprecationsSummary]: Deprecated functionality
```

(The first is triggered by the conventional inline `relationships: {to: ..., field: ...}` generic-test form, which dbt 1.11 wants expressed via an `arguments:` property.)

## Expected

Since dex already invokes dbt's parser at plan time specifically to guarantee "a plan that validates will build," it should surface dbt's deprecation warnings in the plan envelope's `warnings[]`. The author can then fix them **in the diff**, before anything is applied.

## Why it matters beyond tidiness

These deprecation warnings are exactly what poisons the build error channel in #50 — the error extractor picks up a `[WARNING]` line as the failure reason. Surfacing (and thereby eliminating) them at plan time partially subsumes that bug.

Also relevant if `transform plan --scaffold` generates the deprecated generic-test form: dex would be emitting YAML that trips dbt's own deprecation path while reporting it clean.

---

**Environment**

emnarahmouni-exmergo:
Fixes #55: `semantic define`/`update`/`plan` reported `warnings: []` for a plan that parsed cleanly but whose YAML would go on to log dbt deprecation notices (e.g. `PropertyMovedToConfigDeprecation`) at `transform build`.

`shadow_parse` only collected dbt's log messages when the parse failed. It now collects them on a clean parse too, and the caller surfaces them as plan-time warnings instead of letting the author discover them for the first time at build.

Scoped to `semantic define/update/plan`, the only command that runs dbt's real parser today; `transform plan` (model_sql/schem21a_yml) has no such gate and is out of scope here.

Screenshot 2026-07-10 161324

 after fix: 

Screenshot 2026-07-10 161049

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- `semantic define`/`update`/`plan` reported `warnings: []` for a plan that
+  parsed cleanly but whose YAML would go on to log dbt deprecation notices
+  (e.g. `PropertyMovedToConfigDeprecation`) at `transform build`. `shadow_parse`
+  only collected messages on a failed parse; it now collects them on a clean
+  parse too, and the caller surfaces them as plan-time warnings instead of
+  letting the author discover them for the first time at build (where they
+  also poisoned the failure-error channel, #50). (#55)
+
 ## [1.1.0] - 2026-07-09
 
 ### Added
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
Nice

PR #68, opened 2026-07-13
Amazon Redshift Connector

### LINKED ISSUE #39: AWS Redshift Connector
Add the AWS Redshift connector.
NOTE: connectors must be heavily dogfooded in a coding agent before they can be added to `exmergo-dex-core`

spascarella:
## What this changes

Adds Amazon Redshift as dex's sixth connector, Serverless-first and
provisioned-compatible. The mental model: Postgres wire and catalog mechanics
plus Snowflake compute-time gating, billed in seconds with an RPU-hours (and
optional USD) translation. Dogfooded end to end against a live Serverless
workgroup throughout: every empirical claim below (error shapes, wake billing,
HLL limits, SVV semantics, dbt IAM builds) was verified live, not assumed.

### Cost model (the part to review hardest)

- `Paradigm.COMPUTE_TIME`, budgets in seconds, ledger field `billed_seconds`.
  RPU-hours translate via the workgroup's `baseCapacity` (free control-plane
  call); dollars only when `redshift.rpu_price_usd` is set.
- Serverless bills a 60-second minimum per wake, and warmth is unknowable
  without spending. Estimates include the floor exactly once per command
  (latched, so multi-probe maintain sweeps do not multiply it: a 12-probe
  sweep used to quote 738s of which 720 were floors), and execution charges
  it exactly once, by whichever billed statement runs first.
- The budget is hard-enforced regardless of estimate quality: before every
  billed statement, `SET statement_timeout` is wound down to the remaining
  budget. A server kill surfaces as SQLSTATE 57014 and is translated to the
  over-ceiling refusal; other cancellations (WLM rules, admin kills) pass
  through untranslated so the user is never told to raise a budget that is
  not the problem.
- Estimates are byte-rate heuristics from SVV_TABLE_INFO (labeled
  `estimate_quality: "heuristic"`); Redshift has no dry run and EXPLAIN costs
  are relative units. No sampling knob: Redshift has no TABLESAMPLE, so the
  budget is the bound.

### Auth (discover, don't ask)

Discovery order: `redshift.workgroup` (Serverless IAM via the AWS default
chain, endpoint and database resolved from the control plane, temporary DB
credentials minted at connect time), `cluster_identifier` (provisioned IAM),
`REDSHIFT_*` env, committed host config plus `REDSHIFT_PASSWORD`, then a dbt
profiles.yml target (env_var references rendered like dbt would; unrenderable
templates and `method: iam` outputs are skipped). Only a coarse method string
ever crosses the envelope; every failure names all the fixes.

### Where to look

- `packages/dex-core/src/exmergo_dex_core/adapters/redshift.py`: the adapter
  (catalog census via pg_class + SVV_TABLE_INFO + SVV_COLUMNS, HLL profiling,
  wake floor, the single billed door, error translation).
- `connect.py`: discovery chain, control-plane resolution, CostGate wiring.
- `transform/{init,commands,dev_target}.py`: dbt profile rendering (IAM and
  env_var password variants, never a secret), COMPUTE_TIME build accounting,
  dev-schema privilege preflight (skipped for `method: iam` profiles, whose
  DB user is minted at runtime).
- `explore/relationships.py`: the relationship verify probe was rewritten
  from a projected NOT EXISTS to a LEFT JOIN against a deduplicated key set,
  for all connectors; Redshift refuses projected correlated subqueries.
  Verified semantically equivalent on every dialect (NULL FKs, non-unique
  parents, empty parents).
- `tests/fakes/redshift.py` + `tests/adapters/test_connect_redshift.py`
  (66 tests) + safety-spine additions: a stateful fake with the driver's real
  error shapes, injectable clock, and server-side timeout simulation.
- `references/redshift.md`: the connector reference, including the honest
  Serverless notes (metadata is cheap but not free; one dev schema per
  building user or view swaps break).

### Review hardening

A full multi-angle review ran before this PR; all confirmed findings are
fixed in-branch with regression tests. Highlights: the escalation scan now
charges the wake floor; the dev-target preflight keys on the profile's
`method: iam` instead of a magic user name; WLM cancels are no longer blamed
on the budget; billed seconds are recorded even when a statement dies
mid-flight; scope is pushed into the census SQL; the CI seed script is
genuinely idempotent. Known follow-ups (out of scope here): Snowflake has the
same per-estimate resume-floor overcounting fixed here for Redshift, and the
Postgres/Snowflake dbt-profile readers share the Jinja-password gap fixed
here for Redshift.

## How to test

CI: unit jobs just need the `redshift` extra (already wired). The scheduled
live job needs the `redshift-integration` GitHub environment with
`DEX_TEST_REDSHIFT_{WORKGROUP,DATABASE,HOST,ROLE_ARN,REGION}` variables and the
`DEX_TEST_REDSHIFT_DEV_PASSWORD` secret. `scripts/setup_redshift_ci.sh`
provisions all of it (workgroup, usage-limit backstop, OIDC role, seed, users,
the `dex_ci_reader` CI grant, and the GitHub environment), is safe to re-run,
and needs an operator with AWS admin over Redshift Serverless, IAM, and Secrets
Manager plus `psql` and `gh` on PATH.

## Related Issues
This implements https://github.com/exmergo/dex/issues/39#issue

## Checklist

- [x] Engine tests pass (`uv run pytest` in `packages/dex-core`) - 795 passed
- [x] Eval scoring-core tests pass (`uvx pytest evals`) - 26 passed
      (per-connector safety-spine section added; live suite 10/10)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
       # The cloud extras are included so their unit tests and safety families
       # run here (they are offline against fake clients, but they importorskip
       # on the client libraries; without the extras they would silently skip).
-      - name: Sync (duckdb + bigquery + snowflake + databricks + postgres + dev extras)
-        run: uv sync --extra duckdb --extra bigquery --extra snowflake --extra databricks --extra postgres --extra dev --python ${{ matrix.python-version }}
+      - name: Sync (duckdb + bigquery + snowflake + databricks + postgres + redshift + dev extras)
+        run: uv sync --extra duckdb --extra bigquery --extra snowflake --extra databricks --extra postgres --extra redshift --extra dev --python ${{ matrix.python-version }}
       - name: Tier-1 unit tests
         run: uv run pytest -q
 
@@ -48,8 +48,8 @@ jobs:
       - uses: astral-sh/setup-uv@v5
         with:
           enable-cache: true
-      - name: Sync (duckdb + bigquery + snowflake + databricks + postgres + dev extras)
-        run: uv sync --extra duckdb --extra bigquery --extra snowflake --extra databricks --extra postgres --extra dev
+      - name: Sync (duckdb + bigquery + snowflake + databricks + postgres + redshift + dev extras)
+        run: uv sync --extra duckdb --extra bigquery --extra snowflake --extra databricks --extra postgres --extra redshift --extra dev
       - name: Safety-critical assertions
         run: uv run pytest -q tests/test_safety_spine.py
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -6,8 +6,8 @@ name: Integration
 # fork-runnable; this workflow answers "does the connector still work against
 # the real service" on main and on a schedule. Each suite reads shared/public
 # sample data, writes only to its scratch scope, and caps every statement
-# (bytes on BigQuery, seconds on Snowflake, Databricks, and Postgres), so a
-# worst-case run costs cents.
+# (bytes on BigQuery, seconds on Snowflake, Databricks, Redshift, and
+# Postgres), so a worst-case run costs cents.
 #
 # One-time setup (documented in CONTRIBUTING.md):
 #   BigQuery: scripts/setup_bigquery_ci.sh (Workload Identity Pool provider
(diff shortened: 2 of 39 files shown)


spascarella:
Pushed `97dc739`: reworked the CI setup script and validated the scheduled
integration job end to end.

- Seed now runs as the namespace admin via a Secrets-Manager-managed password
  (the IAM-seed path could not create users on a fresh namespace, since a
  Redshift superuser cannot have the disabled password IAM identities carry).
- Added a `dex_ci_reader` role granted to the CI identity's minted DB user
  `IAMR:<role>`, so the job connects with read access (`dex_ro`'s grants never
  reached it).
- `CREATE USER`/`CREATE ROLE` are idempotent and the seed waits out a cold
  Serverless resume.

Validated by dispatching `integration.yml` on this branch (with a temporary,
now-reverted `redshift-integration` branch-policy allowance): the redshift job
ran the full live suite green, **10 passed** in ~38s, exercising the real
OIDC-federated path (assume role -> GetCredentials -> `IAMR:<role>` -> read via
`dex_ci_reader`, plus the dbt build via the password secret). The environment
is back to `main`-only.

marcociav-exmergo (approved):
LGTM! 
Let's let it rip

PR #66, opened 2026-07-12
Explore commands are now dbt project aware with optional "--use-project" flag

marcociav-exmergo:
## What this changes

Introduce a new `--use-project` flag in explore commands to keep dbt project into consideration when exploring warehouses.

Commands
- `explore relationships --use-project`: folds in the dbt project's declared foreign keys at confidence 1.0
- `explore profile --use-project`: lets a semantic model's declared primary entity override the heuristic grain
- `explore map --use-project`: applies declared grain and ranks metric-backing models higher

## Related issues

Closes #65
Internally SCRUM-917

PR #69, opened 2026-07-13
Persist explore profile and relationships results to the .dex/ cache

### LINKED ISSUE #53: `explore profile` does not persist to the `.dex/` cache, forcing a duplicate paid scan to use `explore query`
## Summary

Only `explore map` writes the `.dex/` catalog; `explore profile` returns its profiles and discards them. Since `explore query` **requires** the catalog, a user who has already profiled the exact tables they care about must re-scan them via `map` — paying twice for identical work.

This is documented behavior, but on a metered connector the cost is real.

## Observed

Same 8 tables (`SNOWFLAKE_SAMPLE_DATA.TPCH_SF1`), one session:

| Command | Statements | Warehouse-seconds | Credits |
|---|---|---|---|
| `explore profile` (8 tables) | 16 | 15.89 | 0.00596 |
| `explore map` (same data, for the catalog) | 28 | 38.63 | 0.01449 |

The `map` was needed *solely* to obtain a catalog so `explore query` would run.

## Expected

One of:
- a `--persist` flag on `explore profile` that writes its results through to the `.dex/` cache, or
- `profile` writes through by default (it already produces exactly the per-column payload `map` caches), or
- `explore query` accepts a catalog assembled from prior `profile` results.

## Impact

Charging twice for the same scan is a poor look on the connector whose central thesis is cost governance. It also creates a perverse incentive to skip `profile` and go straight to `map`, which is the *more* expensive command when the user only cares about a few tables.

Interacts with #48: had schema-scoped `map` worked, this would have been a smaller annoyance. Together the two turn "explore one schema cheaply" into a path with no correct route.

---

**Environment**

spascarella:
## What

`explore profile` and `explore relationships` now persist their results to
`.dex/cache.json` by default, merging into any existing cache. Previously only
`explore map` wrote the cache.

Closes #53.

## Why

The connector's whole thesis is cost governance, and today it charges twice for
the same scan. `profile` computes per-column profiles, PII flags, grain, and
data-quality notes, returns them in its envelope, then throws them away.
`relationships` does the same after profiling every object and inferring joins.
Because `explore query` requires the cache, a user who has already profiled the
exact tables they care about must re-scan them via `map` just to make `query`
work.

The contract was already advertised in code: the query firewall's own refusal
messages tell users to run `explore profile <table>` so its "columns and PII
flags are known" (`guards/query_firewall.py`). That promise was false until now.
This change makes the code honor a contract it already prints.

## What changed

- **`cmd_profile`** annotates grain (now project-aware, consistent with `map`)
  and persists the profiled datasets. Prior relationships are preserved, because
  `profile` runs no inference pass.
- **`cmd_relationships`** annotates grain before persisting (so its cached
  datasets carry `candidate_keys`/`grain` like `map`'s do) and writes the
  profiled datasets plus its relationship set. Because it inventories and
  profiles the full object set and infers across all of it, its relationships
  are authoritative for the run and replace the prior set. This composes with
  the `--use-project` work from #66: declared joins from the dbt project are
  folded in and persisted too.
- **New `_merge_profiles` helper**, keyed by exact identifier:
  - Refreshed dataset (profiled this run, present in prior): fresh measurement
    wins, except `rank_score`, which is carried forward so `map`'s connectivity
    ranking survives.
  - New dataset: inserted with `rank_score=None`.
  - Untouched prior dataset: kept verbatim, including its older `profiled_at`
    age marker.
  - `created_at` is preserved from a reusable prior; `updated_at` is bumped by
    `save_cache`.
  - **Connector mismatch** (prior built for A, now profiling B): the prior is
    dropped wholesale and a fresh cache scoped to B is written, with a loud note
    naming the dropped connector. Mixing connectors would poison the PII policy
    and the maintain baseline, and `.dex/` is non-canonical scratch that one
    `explore map` rebuilds. It never refuses, since refusing would break the
    firewall promise this change exists to keep.

## Tests

- `test_profile.py`: fresh-cache write, merge-preserves-relationships-and-rank,
  stale-profile refresh advances `profiled_at`, new-dataset insert,
  connector-mismatch replacement, and an in-memory `_merge_profiles` unit test
  (carry-rank, preserve-relationships, mismatch, and replace variants).
- `test_query.py`: profile-then-query with no prior `map` succeeds; querying an
  unprofiled table is still refused (partial-cache scoping); PII flags flow from
  a profile-built cache into the firewall.
- `test_relationships.py`: datasets and relationships persist in the annotated
  shape; relationships-then-query succeeds.
- `conftest.py`: an autouse `chdir(tmp_path)` fixture, because the repo root
  defaults to the CWD and several existing CLI tests omit `--repo-root`; without
  it these now-persisting commands would write `.dex/` into the checkout.

Full `dex-core` suite: 751 passed, 41 skipped. Explore suite: 126 passed.

## Verification

Manually confirmed end to end against a fixture DuckDB:

1. `explore profile RAW_HOSTS` in an empty repo writes `.dex/cache.json`, then
   `explore query "SELECT COUNT(*) FROM RAW_HOSTS"` is allowed with no prior
   `map` (previously refused with "run explore map first").
2. A PII-projecting query stays refused from a profile-built cache, so the flags
   survive the profile -> cache -> firewall path.
3. `map`, then `profile <one-table>`: relationships and all other datasets stay
   intact, only the profiled table is refreshed, its prior `rank_score` and the
   cache's `created_at` are retained, and `updated_at` advances.

## Known follow-up (out of scope)

`map` folds same-lineage/replica duplicate edges before caching
(`fold_replica_relationships`); `relationships` does not. A cache last written by
`relationships` may therefore contain replica-duplicate edges a `map` run would
fold. This mirrors the two commands' existing envelope behavior; duplicating the
fold logic would need dev-schema config plumbing that `relationships` lacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`explore profile` and `explore relationships` now persist their results
+  to `.dex/cache.json`**, merging into any existing cache instead of
+  discarding the scan they just paid for. Previously only `explore map` wrote
+  the cache, so `explore query` on an already-profiled table demanded a
+  second, redundant warehouse scan via `map`, and the query firewall's own
+  refusal messages ("run `explore profile <table>` first") promised a path
+  that did not work. The merge is keyed by identifier: refreshed datasets
+  carry forward `map`'s rank score, untouched prior datasets keep their older
+  `profiled_at` (and `profile` preserves prior relationships, while
+  `relationships` replaces them with its authoritative full-set inference),
+  `provenance.created_at` survives, and a prior cache built for a different
+  connector is replaced wholesale with a loud note rather than poisoned by
+  mixing. `relationships` also annotates candidate keys and grain before
+  persisting, so its cached datasets match `map`'s shape. Known asymmetry:
+  `relationships` does not fold same-lineage replica edges the way `map`
+  does, mirroring the two commands' existing envelope behavior.
+
 ### Added
 
 - **`--use-project`: explore can read the dbt project, on request.**
--- a/packages/dex-core/src/exmergo_dex_core/explore/commands.py
+++ b/packages/dex-core/src/exmergo_dex_core/explore/commands.py
@@ -4,7 +4,8 @@ Each ``cmd_*`` opens the adapter, drives the explore engine, and shapes the resu
 into the sanitized envelope. Keeping this here (not in ``cli.py``) keeps dispatch
 thin and keeps ``map``'s composition (it runs inventory, profile, and relationships
 together) out of the CLI layer. These are the only explore commands that hold an
-adapter; ``map`` is the only one that writes, and only to the ``.dex/`` cache.
+adapter; ``map``, ``profile``, and ``relationships`` all persist what they learned,
+and only to the ``.dex/`` cache, so a scan is never paid for twice.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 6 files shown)


marcociav-exmergo (approved):
Nice.

I'm creating an issue for the follow-up task you outlined at the end of the summary

PR #72, opened 2026-07-14
Fold replica edges in explore relationships and fix dev-schema matching

### LINKED ISSUE #70: Add missing duplicate edge folding logic to `explore relationships` command
`explore map` folds same-lineage/replica duplicate edges before caching (fold_replica_relationships); relationships does not. 
A cache last written by `explore relationships` may therefore contain replica-duplicate edges a map run would fold. 
This mirrors the two commands' existing envelope behavior; duplicating the fold logic would need dev-schema config plumbing that `explore relationships` lacks.

spascarella:
# Fold replica edges in `explore relationships`; fix dev-schema matching

## Summary

`explore map` folds same-lineage/replica duplicate relationship edges before
writing `.dex/cache.json`, but `explore relationships` didn't — since #69
persisted its output to the same cache, this let replica-duplicate edges
written by `relationships` disagree with what a `map` run would produce for
the same warehouse. This was worse for `relationships` since it profiles the
full inventory (vs. map's top-ranked subset), making it more likely to pull a
dev/replica schema into scope.

- Extracted the inline `dev_schemas` frozenset construction from `cmd_map`
  into a shared `_dev_schemas` helper, used by both commands so they can't
  drift apart again.
- `cmd_relationships` now calls `fold_replica_relationships` before merging
  declared joins, so the folded set flows into both the persisted cache and
  the envelope, with the same fold note `map` emits.
- Fixed two latent bugs in `fold_replica_relationships`'s dev-schema matching:
  qualified config values (e.g. BigQuery `project.dataset`) never matched,
  and matching was case-sensitive (breaking on Snowflake/Redshift's typical
  casing). Both sides now compare case-insensitive short (last-segment)
  names.
## Related issues

Closes #70 .

## Tests

- New command-level test: `explore relationships` on a mirrored warehouse
  folds duplicates and persists the folded set.
- New engine tests: qualified dev-dataset config and case-mismatched
  dev-schema config both match correctly.
- Confirmed the existing no-op regression test (non-mirror fixture) still
  passes.

## Verification

- `uv run pytest` — 841 passed, 51 skipped.
- End-to-end: ran `explore relationships` then `explore map` against the same
  mirrored fixture; resulting `.dex/cache.json` relationship sets are now
  identical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`explore relationships` now folds same-lineage/replica duplicate edges
+  before caching, matching `explore map`** ([#70]). `relationships` profiles
+  the full inventory, so it is even more likely than `map` to pull a
+  dev/replica schema into scope alongside its source; without folding, a
+  cache last written by `relationships` could carry replica-duplicate edges
+  that a `map` run would have folded away. The folded set now flows into both
+  the envelope and the persisted cache, and a note reports how many edges were
+  folded and how many objects mirror source lineage. Dev-schema matching is
+  also fixed for two cases the original folding logic missed: a BigQuery-style
+  qualified `dev_dataset` (`project.dataset`) is compared by its bare schema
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
V nice

PR #88, opened 2026-07-16
Issue-84: Periodic stderr progress for long explore runs

### LINKED ISSUE #84: No progress signal on long-running `explore map` runs
## Summary

`explore map` produces no output until it either completes or errors. On a
long profiling run (many objects, or `--verify` adding many probes), there's
no way to distinguish "still working" from "hung" without external tooling.

## Observed

A long `explore map --full --verify` run (compounded, separately, by an
unrelated laptop-sleep incident mid-run) looked indistinguishable from a
genuinely hung process for well over an hour. The only way to tell it was
still alive was to shell out to `ps` and read the process's CPU time as a
proxy signal -- non-zero-but-tiny CPU time over a long wall-clock duration
suggested it was waiting on network I/O rather than stuck, but this
required external tooling dex itself gave no signal for.

## Expected

Even a minimal periodic progress line (e.g. to stderr: "profiled 40/90
objects") during a long `map` run, so a live-but-slow run is visibly
distinguishable from a hung one without needing to shell out and infer from
process metrics.

## Impact

Long profiling runs are common on real warehouses; the lack of any
in-flight signal makes a slow-but-healthy run indistinguishable from a
genuinely stuck one, which is a bad experience regardless of the root cause
of any given slowdown.

## Suggested fix

Emit a periodic progress line (object count profiled so far / total) to
stderr during `explore map`, gated on run duration or object count so it
doesn't add noise to fast runs.

---

**Environment**

spascarella:
# Periodic stderr progress for long explore runs

## Problem

`explore map`, `explore relationships`, and `explore profile` produce no output until they complete or error. On a slow run (many objects, or `--verify` adding a probe per inferred join) there's no way to tell "still working" from "hung."

## Change

Adds a minimal, contract-safe progress line to **stderr** during the slow loops, gated so fast runs stay silent and the stdout envelope is untouched.

- **New `progress.py`**: a stdlib-only `ProgressReporter` that emits `dex: profiled 40/90 objects` (and `dex: verified N/M joins` on `--verify`). Nothing before `first_after` seconds, throttled to one line per `interval`, silent on fast runs. Stream defaults to `sys.stderr`, never stdout. Lines carry only counts, never identifiers or column names.
- Threads an optional `progress` reporter through `profile()` and `verify_relationships()` (defaults to `None`, so existing callers are unaffected).
- Wires reporters at the three call sites in `commands.py`, constructed after the billed-handshake early-return.

## Related issue
This closes #84 

## Notes

- The final item is announced by `done()`, not `advance()`, which keeps `total` of 0 or 1 silent by construction and avoids a duplicate completion line.
- Known limitation (documented): progress proves "N items done," not per-second liveness; a stall inside a single object's profiling still shows no line until it returns. No background timer thread by design.

## Testing

- Unit tests for the reporter (fast-run silence, threshold + throttle, exact line format, `done()` behavior, `total` 0/1 silent).
- Wiring tests: contract guard (fast run gives empty stderr and exactly one stdout envelope) and direct emission-path tests for both loops.
- `ruff check` clean; end-to-end sanity confirmed on a DuckDB fixture.


marcociav-exmergo (approved):
LGTM!

Merging

PR #89, opened 2026-07-16
Incremental cache persistence for explore profiling runs

### LINKED ISSUE #75: `explore map` runs that fail on budget exhaustion persist zero progress
## Summary

When an `explore map --full` run fails partway through because the budget
runs out, none of the profiling work it already paid for gets written to
`.dex/cache.json`. Real spend happens; the cache is unchanged.

## Repro

1. Run `explore map --full --confirm --budget 2000000000 --verify` against
   a ~90-table warehouse. It fails partway through (see the linked verify-cost
   issue) after real BigQuery spend.
2. Inspect `.dex/cache.json`'s `profiled_count` before and after the failed
   run.

## Actual

`profiled_count` is identical before and after -- unchanged from a prior,
smaller successful run (25 of 90-ish objects). The failed run's spend
produced nothing durable.

## Expected

Per-object profiling results should persist incrementally as each object
completes, not only at the very end of a fully successful run -- similar in
spirit to how `map`'s existing `carried_forward_count` mechanism already
handles skip-based incremental updates *across* separate runs. A run that
dies at object 60 of 90 should leave 60 objects' worth of profile in the
cache, not zero.

## Impact

Combined with the verify-cost estimation gap (see linked issue), a single
profiling attempt can burn through a meaningful budget and leave the user
with literally nothing to show for it -- worse than simply refusing
upfront, since real money was spent for zero durable benefit.

## Suggested fix

Write each object's profile to `.dex/cache.json` as it completes, rather
than batching the write to the end of the run. A budget-exhaustion failure
partway through should still leave a cache reflecting everything actually
profiled before the failure.

## Related

Same underlying theme ("paid for a scan, got nothing durable") as #53
(`explore profile` never persisting to cache at all) -- different command
and different specific mechanism, but likely worth considering together.

---

**Environment**
  datasets, ~90 tables total)

spascarella:
### 1. `profile()` gains an optional per-object callback
[packages/dex-core/src/exmergo_dex_core/explore/profile.py](packages/dex-core/src/exmergo_dex_core/explore/profile.py)

- Added `from collections.abc import Callable`.
- New signature:
  `def profile(adapter, identifiers, *, on_complete: Callable[[Dataset], None] | None = None) -> list[Dataset]`.
- Each fully-profiled `Dataset` is bound to `ds`, appended, then handed to
  `on_complete`. The return contract (`list[Dataset]`) is unchanged, so existing
  callers are untouched.
- The callback receives the **raw** dataset. Because the cost gate's `charge()`
  fires *inside* an object's `column_aggregates` — before that object's `Dataset`
  is appended — the callback only ever checkpoints fully-profiled objects, never
  a half-scanned one.

### 2. Shared checkpointer + partial-failure helpers
[packages/dex-core/src/exmergo_dex_core/explore/commands.py](packages/dex-core/src/exmergo_dex_core/explore/commands.py)

- Imports: `Callable`, `CACHE_FILE`, `OverCeilingError`.
- `_profile_checkpointer(store, prior, connector, now) -> (checkpoint, accumulated)`
  persists each dataset as it completes, reusing `_merge_profiles`
  (KEEP_RELATIONSHIPS). A partial write is therefore exactly a
  `cmd_profile`-shaped result: fresh profiles folded over the same-connector
  prior, prior relationships preserved, no fabricated stubs. `accumulated` is
  returned so the failure handler can report "N of M".
- `_over_ceiling_error(store, accumulated, selected)` builds the
  partial-completion error envelope:
  - N > 0: names "N of M" and the cache path.
  - N == 0: says no partial profiles were saved and omits the path.

### 3. Wired into `cmd_map`, `cmd_relationships`, `cmd_profile`

In each command:

- **State capture moved up**: `store`, `now`, and `prior = store.load_cache()`
  are captured once, before the first checkpoint write, so the success-path
  compose still reads the pre-run cache (not a checkpoint the run itself wrote).
- **Billed-connector gate**: the checkpointer is built only when
  `command_args.cost_gate(adapter) is not None`; otherwise `on_complete=None`.
  DuckDB (`FREE_LOCAL`) never checkpoints — it can't exhaust budget and its
  re-runs are free, which also avoids O(n²) full-file writes on large local
  warehouses.
- **`OverCeilingError` handling**: caught specifically (not the `CostGuardError`
  base) around the profile call. On catch, the command skips the rest of the
  success path and, after `adapter.close()`, returns
  `_over_ceiling_error(...)` with spend stamped (`stamp_spend` reads the gate,
  valid after close). Non-budget exceptions still propagate to `cli.py`.
- **Success path unchanged**: `_annotate_grain`, inference, folding, ranking, and
  the final authoritative `save_cache` are untouched. The final save overwrites
  the checkpoints with the composed, ranked, relationship-bearing cache.

### 4. CHANGELOG
[CHANGELOG.md](CHANGELOG.md) — added a `### Fixed` entry under `## [Unreleased]`
(issue #75).

## Design decision: raw checkpoints

Checkpoints persist **un-annotated** profiles:

- `_annotate_grain` appends to `data_quality` and is **not idempotent** —
  annotating on the checkpoint would double notes when the success path
  re-annotates the full list.
- Per-object annotation would pass a single-element `identifiers` to
  `resolve_declared`, changing declared-key/grain ambiguity resolution versus the
  authoritative full-list pass.
- The budget-paid signals (aggregates, PII flags, exact-distinct escalations,
  `composite_keys`) are all on the raw `Dataset`. `candidate_keys`/`grain` are
  cheap local derivations any later run recomputes for free.

Accepted tradeoff: after a mid-run failure the partial cache carries only
`profile()`-level `data_quality` notes (empty-table, adapter notes), not the
grain/candidate-key interpretation, which regenerates on the next successful run.
The partial cache is still fully valid for its main consumer, `explore query`'s
firewall, which needs only columns + PII flags.

## Files modified

- `packages/dex-core/src/exmergo_dex_core/explore/profile.py`
- `packages/dex-core/src/exmergo_dex_core/explore/commands.py`
- `CHANGELOG.md`
- `packages/dex-core/tests/explore/test_incremental_checkpoint.py` (new)

## Verification

| Check | Result |
| --- | --- |
| New checkpoint tests (`test_incremental_checkpoint.py`) | 7 passed |
| Full explore suite (`tests/explore`) | 145 passed |
| Full suite (`uv run pytest`) | 876 passed, 51 skipped |
| `ruff check` / `ruff format` on touched files | clean |

The 51 skips are the pre-existing "connector extra not installed" skips,
unrelated to this change.

### Test coverage (all five plan cases)

1. Mid-run budget exhaustion leaves a partial cache — `map`, `relationships`,
   `profile` (asserts "1 of 2", cache path in message, one full profile in cache).
2. First-object failure saves nothing — `map`, `profile` (asserts "no partial
   profiles were saved" and `.dex/cache.json` absent).
3. Successful billed `map` overwrites checkpoints with the composed cache
   (ranking + relationships present).
4. DuckDB `map --full` never checkpoints (exactly one save; no per-object writes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **`explore map`, `explore relationships`, and `explore profile` now persist
+  each object's profile as it completes on billed connectors** ([#75]).
+  Previously the cache was written exactly once, at the very end of the command,
+  after the whole profiling pass plus inference and ranking finished. When a run
+  against a billed connector (BigQuery, Snowflake, Redshift, Postgres,
+  Databricks) exhausted its budget partway through, the cost gate raised mid-pass
+  and none of the profiling already paid for reached `.dex/cache.json` — real
+  spend, no cache. Each of the three commands now checkpoints every fully
+  profiled object to the cache as it completes, so a run that dies at object 60
+  of 90 leaves 60 objects' worth of raw profile behind, and reports how many of
+  how many objects were saved. A fully successful run still overwrites the
+  checkpoints with the authoritative composed cache (relationships, ranking,
+  carry-forward), and the free DuckDB path is unchanged (its re-runs are free, so
+  it never checkpoints).
 - **`explore relationships` now folds same-lineage/replica duplicate edges
   before caching, matching `explore map`** ([#70]). `relationships` profiles
   the full inventory, so it is even more likely than `map` to pull a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   after the whole profiling pass plus inference and ranking finished. When a run
   against a billed connector (BigQuery, Snowflake, Redshift, Postgres,
   Databricks) exhausted its budget partway through, the cost gate raised mid-pass
-  and none of the profiling already paid for reached `.dex/cache.json` — real
+  and none of the profiling already paid for reached `.dex/cache.json` real
   spend, no cache. Each of the three commands now checkpoints every fully
   profiled object to the cache as it completes, so a run that dies at object 60
   of 90 leaves 60 objects' worth of raw profile behind, and reports how many of
--- a/packages/dex-core/src/exmergo_dex_core/explore/commands.py
+++ b/packages/dex-core/src/exmergo_dex_core/explore/commands.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import argparse
 import json
+from collections.abc import Callable
 from datetime import UTC, datetime
 from pathlib import Path
 
@@ -19,8 +20,16 @@ from .. import command_args, dbt_project
 from .. import envelope as env
 from ..adapters import get_dialect
 from ..adapters.base import Adapter, ObjectMeta, QueryResult
-from ..cache import Dataset, DexCache, DexStore, Relationship, match_identifier
+from ..cache import (
+    CACHE_FILE,
+    Dataset,
+    DexCache,
+    DexStore,
+    Relationship,
+    match_identifier,
+)
 from ..config import DexConfig, QueryLimits, load_config
+from ..guards.cost_guard import OverCeilingError
 from ..guards.query_firewall import (
     InspectedQuery,
     QueryRefusedError,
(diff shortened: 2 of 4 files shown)


marcociav-exmergo (approved):
Yeah this is super useful,
I had a long-ish running `explore map` that took 20 minutes and Claude had no visibility on what was going on in that time window.

This should help, approved!

PR #91, opened 2026-07-16
Scrum 812 dex k means clustering skill

### LINKED ISSUE #87: Dex should be able to perform k-means clustering on table samples

amarkou:
`explore cluster` auto-selected any numeric non-PII column as a feature, excluding only *unique* keys. Foreign keys and ID-named columns slipped through and dominated the clustering.

## Why

Found by running clustering against the ADE-bench DuckDB warehouses:

- 10 of 11 successful runs clustered on key columns
- `f1.lap_times` split old-era vs new-era races by `raceId`/`driverId` — not a segmentation
- `airbnb.RAW_LISTINGS` clustered by `HOST_ID` magnitude and scored silhouette **0.67**, i.e. bad clusters look confident

## Changed

- Auto-selection now also excludes foreign keys (from the joins `explore map` already infers) and key-shaped names (`id`, `uuid`, `guid`, `key`, on word boundaries so `grid`/`paid`/`monkey` stay features)
- `--features` still opts any of them back in
- Notes name every excluded column; the "too few features" error now carries the reason instead of a bare count
- Docs updated (contract, SKILL, CHANGELOG) — they already claimed "non-key", which wasn't true

## Validation

Re-ran across all 9 ADE-bench warehouses, 140 tables, with joins inferred:

- **0** key columns leaked into features (was 10/11)
- 53 tables cluster under both old and new rules — no real measure lost
- 17 flipped to an honest refusal; each had ≤1 actual measure (e.g. `orders` was clustering on `employee_id, customer_id, shipper_id, status_id` + one fee)
- Name rule does the work in practice: inference missed `entity_uuid` despite 756 inferred joins, so both layers are needed
- 11 new tests; 911 pass


marcociav-exmergo (commented):
Awesome findings!

Let's solve the conflicts and then merge this bad boy

marcociav-exmergo (approved):
LGTM

PR #92, opened 2026-07-16
Seed the explore cluster sample and flag degenerate clusters

### LINKED ISSUE #87: Dex should be able to perform k-means clustering on table samples

amarkou:
Two things `explore cluster` got wrong, both found by running it against the ADE-bench warehouses.

**It wasn't reproducible.** The sample was unseeded, so the same command on the same table returned a different answer each run. `f1.lap_times` gave k=2/silhouette 0.95 on one run and k=4/0.37 on the next — not a wobble, a different conclusion about the data. The config docstring already claimed `random_state` made runs reproducible; seeding sklearn is pointless when the sample underneath re-draws.

**A high silhouette can mean the opposite of what it looks like.** On `lap_times`, silhouette 0.95 was k-means peeling off 3 rows in 20,000 (0.01%) — the 28-minute red-flag laps. That's outlier detection wearing a segmentation score, and nothing in the envelope said so. Same failure mode as the FK bug in #91: the number looks confident, the answer misleads.

## Changed

- `cluster.sample_seed` (default `0`; null opts out) seeds the draw. DuckDB emits `USING SAMPLE reservoir(N ROWS) REPEATABLE (seed)` — the bare `USING SAMPLE n ROWS REPEATABLE (...)` is a parser error, so the `reservoir(...)` spelling is pinned by a test
- Only DuckDB seeds, because it's the only dialect verified against a live engine. The other five emit no seed clause, report `sample_repeatable: false`, and say so in the notes rather than implying a reproducibility they don't have (same reasoning as `references/redshift.md` refusing a sampling threshold it can't honor)
- New `sample_repeatable` field in the envelope
- A cluster under 1% of the sample is named an outlier pocket that inflates the silhouette, with the fix (read as outlier detection, or force `-k`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,9 +65,16 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   scanning commands. Only aggregates cross the boundary: per-cluster sizes and
   fractions, centroids (feature means), inertia, and the silhouette score;
   with `-k` omitted the engine sweeps k and reports the silhouette it chose
-  from. scikit-learn rides behind a new `[cluster]` extra, lazy-imported so the
-  light default install stays light and the explore skill wrapper adds it
-  automatically for this subcommand.
+  from. The sample is seeded where the dialect allows it (`cluster.sample_seed`,
+  default 0; DuckDB `REPEATABLE`), because a re-drawn sample is a different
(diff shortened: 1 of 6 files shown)


marcociav-exmergo (approved):
Nice!

PR #95, opened 2026-07-17
Confirmation checkpoint for the verify phase

### LINKED ISSUE #74: `explore map --verify`'s upfront cost estimate doesn't include verify's cost at all
## Summary

`explore map --full --verify`'s dry-run estimate only covers the profiling
pass. Relationship verification's actual cost is not estimated at all, and
can be several times the base estimate -- there's no signal of this before
the budget you agreed to gets exhausted mid-run.

## Repro

1. Run `explore map --full` (no `--verify`) against a warehouse with ~90
   tables. Dry-run estimate: 1.11 GB.
2. Run `explore map --full --confirm --budget 2000000000 --verify` (2 GB,
   comfortably above the 1.11 GB base estimate).

## Actual

Failed partway through:

```
{"status": "error", "data": {}, "cost": {"paradigm": "free_local", "estimate": null, "ceiling": null}, "warnings": [], "diffs": [], "errors": ["the remaining budget (5608448 bytes) is below BigQuery's 10485760-byte minimum billed per query; raise --budget or narrow the work"]}
```

Doubling the budget to 4 GB reproduced the identical failure shape (ran out
right at the very end, again below the 10 MB per-query minimum). Only
dropping `--verify` entirely let the same `--full` profiling pass complete
inside the original 2 GB.

## Expected

The upfront estimate should account for verify's cost, or -- since verify's
real cost genuinely can't be known until profiling has found the candidate
relationships to verify -- present a **second** `needs_confirmation`
checkpoint after profiling completes and candidates are counted, before any
verify probes run (e.g. "found 448 candidate relationships across 87
objects; verifying all of them is estimated at ~Y bytes; proceed?").

## Impact

Turned one profiling pass into three separate user-facing budget
negotiations (2 GB, 4 GB, then dropping `--verify`), with real BigQuery
spend against the failed attempts and nothing durable to show for it (see a
related issue on incremental persistence). For a tool whose central pitch is
cost governance, an unbounded, unestimated cost hiding behind a flag is a
significant gap.

## Suggested fix

After the profiling phase of `explore map --full --verify` completes and
the candidate-relationship count is known, insert a second
`needs_confirmation` step (mirroring the existing preflight shape) scoped
specifically to the verify phase, before issuing any verify probes.

---

**Environment**
  datasets, ~90 tables total)

spascarella:
## Summary

`explore map --full --verify`'s dry-run estimate only covered the profiling pass. Relationship-verify probes ran against the same confirmed budget with no estimate and no checkpoint, so on a large warehouse the run could die mid-verify with a generic error once the budget ran out — with no signal beforehand that verify could cost several times the base estimate.

This adds a second, headroom-gated `needs_confirmation` checkpoint after profiling completes and candidate relationships are known, mirroring the existing preflight shape.

## Changes

- **`guards/cost_guard.py`**: new `CostGate.preflight_phase` — a mid-command gate for costs only knowable after earlier spend. Raises `ConfirmationRequiredError` (recoverable via a bigger budget) only when the phase estimate doesn't fit what remains of the ceiling.
- **`command_args.py`**: new `verify_handshake`, mirroring `billed_handshake`, priced via the existing `probe_statements` + `query_estimate` dry-run hooks (the same pattern `maintain grain` already uses). Payload includes `phase: "verify"`, candidate/object counts, and a hint with the exact `--confirm --budget N` to re-run with. `stamp_spend` now preserves a checkpoint envelope's cost instead of overwriting it.
- **`explore/commands.py`**: new `_verify_estimate` helper; wired into both `cmd_map` and `cmd_relationships`. Verify is now wrapped in `try/except OverCeilingError` for graceful degradation (a warning instead of a crash) when per-probe estimates drift mid-run.

## Related issue
This closes #74 

## Behavior

1. Unconfirmed run → needs_confirmation #1 (profiling estimate), now noting that verify is priced separately and may trigger a second checkpoint.
2. `--confirm --budget B`: profiling and inference run; probes are dry-run priced for free. If they fit the remaining budget, verify runs in the same pass — no change for the common case.
3. If they don't fit → needs_confirmation #2 ("found N candidate relationships across M objects; verifying them all is estimated at ~Y bytes beyond what remains of the budget..."), with the work already paid for **persisted**: cache saved with fresh profiles + unverified relationships, full map summary and spend included, zero probe bytes billed.
4. Re-run with the suggested budget completes in one pass (profiling re-runs at full cost — there's no profile-cache reuse today, which is why the checkpoint is headroom-gated rather than always-fire).
5. Estimate drift mid-verify now degrades to a warning ("verified N of M candidate joins; results saved") instead of an uncaught error.

DuckDB (free) stays fully confirmation-free throughout.

## Test plan

- [x] `pytest tests/guards/test_cost_guard.py` — new `preflight_phase` unit tests (headroom fit, exceeds, session-ceiling binding, zero/no-ceiling edge cases)
- [x] `pytest tests/explore/test_billed_handshake.py` — new coverage: verify within budget (one pass), verify beyond budget (checkpoint shape + persistence), zero candidates (no checkpoint), mid-verify `OverCeilingError` (warning + persisted partial results), credit/seconds `describe_estimate` shape, DuckDB `--verify` regression
- [x] Full suite: `pytest` (969 passed, 68 skipped)
- [x] `ruff check` / `ruff format --check`
- [x] Manual end-to-end: `dex explore map --full --verify --path <duckdb file>` against a two-table FK warehouse — completes in one pass, relationship verified in `.dex/cache.json`

### DIFF
--- a/packages/dex-core/src/exmergo_dex_core/command_args.py
+++ b/packages/dex-core/src/exmergo_dex_core/command_args.py
@@ -11,6 +11,7 @@ instead of re-deriving it.
 from __future__ import annotations
 
 import argparse
+import math
 from pathlib import Path
 
 from . import envelope as env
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
LGTM!

PR #97, opened 2026-07-17
Redshift cold-start race fix

spascarella:
# Fix Redshift cold-start connection race

## Summary

CI's live Redshift integration suite intermittently failed on
`test_connect_test_discovers_connection_and_reports_read_only` with a bare
error envelope (`status: error`, `paradigm: free_local`) and no useful
message. The failing run took 135.85s versus a healthy run's ~47s — the tell
that this was a cold-start issue, not a code defect in `connect test` or
`capabilities()`.

**Root cause:** an idle Redshift Serverless workgroup resumes on first
contact. The startup/auth handshake blocks until compute is ready, and a slow
resume gets torn down mid-handshake by the network path. Whichever command
happens to touch the workgroup first pays for that resume; every command
after it finds the workgroup warm and succeeds. `dex` opened the connection
with `redshift_connector.connect(...)` directly — no timeout, no retry — so a
reset handshake surfaced as a raw, unretried exception.

Verified against the driver source (`redshift_connector`) and CI run history:
70 existing adapter unit tests pass against the fake connection, the same
code path passes warm in CI 10/10, and the only server call unique to the
failing test (`version()`) is the same catalog-lookup class as calls the
passing tests already exercise successfully — ruling out a deterministic bug
in the capabilities probe itself.

## Change

`packages/dex-core/src/exmergo_dex_core/connect.py`

- Added `_connect_redshift()`, wrapping the single `redshift_connector.connect`
  call site used by every Redshift auth path (Serverless IAM, cluster IAM,
  environment, config target, dbt profile).
- Retries only the connection-level errors a resuming workgroup produces
  (`InterfaceError`, `OperationalError`) with backoff (2s / 5s / 10s, 4
  attempts total). A `ProgrammingError` (bad credential, missing database,
  denied grant) is **not** retried and surfaces immediately.
- Applies a 90s per-attempt connect timeout so a stalled handshake fails fast
  instead of hanging, then clears the socket timeout once connected so it
  never caps a later billed query's result read.
- On exhaustion, raises `CredentialDiscoveryError` naming the resume instead
  of leaking a raw driver exception into the generic error envelope.

No billed compute is added by the retries — no statement runs during
connect; the Serverless wake minimum is still billed exactly once, by the
first real statement, as before.

## Tests

- 3 new unit tests in `test_connect_redshift.py` against a fake driver:
  - retry-then-succeed on transient errors
  - give-up-and-raise-with-cause after exhausting the retry budget
  - a credential/permission refusal is **not** retried
- Full non-integration suite: 961 passed, 17 skipped, 0 failures.
- `ruff check`: clean.

## Verification caveat

I don't have access to the CI environment's seeded Serverless workgroup, so
I couldn't force a genuinely cold reproduction end-to-end. The fix targets
the exact exception classes and blocking window traced from the
`redshift_connector` source and the historical CI timing; the unit tests
exercise the retry/backoff/give-up logic directly against a fake.

## Changelog

Added an `[Unreleased] / Fixed` entry in `CHANGELOG.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Redshift connections survive a Serverless cold start.** An idle Serverless
+  workgroup resumes on first contact, and a slow resume can reset the startup
+  handshake, so the first command to touch a cold workgroup failed hard while
+  everything after it ran warm. The connect is now retried with backoff on the
+  transient connection errors a resume produces (a wrong credential or database
+  still fails immediately), across a window wide enough to cover the wake. A
+  per-attempt connect timeout bounds a stalled handshake and is cleared once the
+  connection is up so it never caps a later billed query's result read.
+
 ## [1.2.1] - 2026-07-17
 
 ### Changed
--- a/packages/dex-core/src/exmergo_dex_core/connect.py
+++ b/packages/dex-core/src/exmergo_dex_core/connect.py
@@ -24,8 +24,10 @@ for.
 
 from __future__ import annotations
 
+import contextlib
 import os
 import re
+import time
 import tomllib
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (approved):
Let's add this in and monitor the Redshift integration tests

PR #98, opened 2026-07-17
Skip-if-cached profiling for explore map / explore relationships

### LINKED ISSUE #96: `explore map`/`explore relationships` always re-profile every selected object from scratch, even when a prior `.dex/cache.json` already holds a fresh profile for that exact object from an earlier run
## Summary

`explore map`/`explore relationships` always re-profile every selected object from scratch, even when a prior `.dex/cache.json` already holds a fresh profile for that exact object from an earlier run. There is no skip-if-cached path, so any re-run that re-selects the same objects re-pays the full profiling scan cost.

## Actual

Every invocation of `cmd_map` / `cmd_relationships` (`explore/commands.py:518`/`301`) re-profiles the full selected set every time, regardless of whether `.dex/cache.json` already has an up-to-date profile for those same objects. `profile()` has no notion of "already profiled, still fresh, skip."

## Expected

An object whose cached profile is still fresh (same connector, schema unchanged, profiled within some staleness window) should be reused without a re-scan. Re-profiling should only happen for objects that are new, newly selected, or whose cached profile is stale/invalidated.

## Impact

- Every re-run of `explore map --full` (or `--verify` re-runs triggered by the verify-phase confirmation checkpoint) pays full profiling cost again, even though nothing changed.
- This is the reason the verify-phase checkpoint (see companion fix) had to be **headroom-gated** rather than always-fire: an always-fire checkpoint would force a mandatory extra invocation on every billed `--verify` run, and each of those re-runs would re-scan the entire object set at full price instead of cheaply resuming from cache.
- More generally, iterative workflows ("map, look at results, tweak `--full`/ranking, map again") on a metered warehouse are more expensive than necessary.

## Suggested Fix

Give `profile()` (or its caller in `cmd_map`/`cmd_relationships`) a skip-if-cached path: before profiling an object, check `.dex/cache.json` for an existing `Dataset` with a matching identifier, connector, and a `profiled_at` newer than some freshness threshold (and ideally a schema-version/column-set check to catch drift), and reuse it wholesale instead of re-scanning. This mirrors the reuse `_compose_datasets` already does for unselected objects (`explore/commands.py:1262-1267`) — the same carry-forward logic, applied to selected objects too when the cache is fresh. Needs a way to force a re-profile (e.g. `--refresh` or `--no-cache`) for callers who know the source changed.

### Steps to reproduce

- Run `dex explore map --full --confirm --budget <B> --verify` against a billed connector (BigQuery) where profiling fits the budget but the verify-phase checkpoint fires (see the `--verify` second-checkpoint fix) — or simply run `dex explore map --full` twice in a row with no schema changes in between.
- Inspect `fake_bq_client.query_calls` (or the real job history): the second run issues the same aggregate-scan queries for every object again, identical to the first run.
- `_compose_datasets` (`explore/commands.py:1253`) only reuses a prior profile for an object that was **not** selected/profiled this run (the `ds is None` branch, line 1277-1282); every selected object always goes through `profile_mod.profile(...)` (`explore/profile.py:234`) unconditionally.

### Connector

BigQuery

### dex-core version

1.2.1

### Environment

- Relevant files: `explore/profile.py:234`, `explore/commands.py:1253` (`_compose_datasets`), `explore/commands.py:518`/`301` (`cmd_map`/`cmd_relationships`)
- Discovered while designing the fix for the `explore map --full --verify` missing cost-estimate/checkpoint issue

spascarella:
## Summary

`explore map` and `explore relationships` re-profiled every selected object
from scratch on every run, even when `.dex/cache.json` already held a fresh
profile from an earlier run. Iterative workflows (map, tweak, map again) and
`--verify` re-runs paid full profiling cost every time, even when nothing
changed.

Before profiling a selected object, both commands now check the cache for a
same-connector, schema-matching, recently-profiled entry and reuse it wholesale
instead of re-scanning — mirroring the reuse `_compose_datasets` already did
for unselected (below-rank-cutoff) objects, just gated by freshness instead of
by rank.

## Changes

- **`_split_fresh_stale`** (`explore/commands.py`): splits a selected set into
  `stale` (needs profiling) and `fresh_reused` (served from cache). An object
  is fresh only if: `--refresh` was not passed, the prior cache is from the
  same connector, it has a prior profile with columns and a `profiled_at`
  within `profile_freshness_hours`, and its column signature (name, type,
  nullable) still matches the warehouse's free `table_metadata()` call.
  Fail-closed throughout — any doubt re-profiles.
- **`cmd_map` / `cmd_relationships`**: only `stale` objects are estimated,
  enter the billed handshake, and get scanned. When nothing is stale, the
  handshake and scan are skipped entirely. Reused datasets are not
  re-annotated (they already carry their annotations from the run that wrote
  them) and rejoin the freshly-profiled set for inference/compose/merge.
- New envelope field **`cache_hit_count`**, distinct from
  `carried_forward_count` (rank-cutoff carry-forward is a different axis).
  `skipped_count` is now decoupled from `len(profiled)`. New reuse notes on
  both the success envelope and the billed-handshake estimate.
- New **`--refresh` flag** on `map`/`relationships`: forces a full re-profile
  of every selected object, for callers who know the source changed in a way
  the cheap metadata check can't see.
- New **`profile_freshness_hours`** config knob (`DexConfig`, default `24.0`;
  `0` disables reuse). No cache schema change — `Dataset.profiled_at` and
  stored columns already carry everything the check needs.

## Changes
This closes #96 

## Design interaction worth flagging

The freshness check is **schema + time only, not row-count**. `table_metadata()`
returns an *estimated* row count on Postgres and Redshift (planner stats), so
using it in the freshness gate would either produce false staleness noise on
every run there, or require excluding those connectors inconsistently. Schema
signature was the only cheap, exact, cross-connector signal available.

**Consequence:** a volume-only change (rows added/deleted, no schema change)
within the freshness window is invisible to a default re-map — it will be
served from cache. This surfaced in
`tests/maintain/test_volume.py::test_axis_results_merge_in_drift_json`, whose
"accept new state → re-map → re-snapshot" step relied on `explore map` seeing
a post-DELETE row count; that test now passes `--refresh` explicitly, which is
the intended escape hatch for this exact case.

This is a deliberate tradeoff (matches the plan's schema-drift-only design,
paralleling `maintain/drift.py`'s schema axis), but it's worth calling out
explicitly: **`explore map` is no longer sufficient by itself to refresh
volume/row-count in the cache** within the freshness window. Any workflow that
depends on `map` picking up a pure data change (not a schema change) needs
`--refresh` or `profile_freshness_hours: 0`.

## Testing

- `uv run pytest` — 966 passed, 68 skipped (connector-extra tests requiring
  optional deps), 0 failed.
- New tests cover: reuse across identical re-runs, re-profiling only the
  schema-changed object, `profile_freshness_hours: 0`, `--refresh`, the
  `relationships` equivalents, and two BigQuery billed-handshake tests
  (fresh-cached objects excluded from the preflight estimate; a fully-fresh
  re-run needs no confirmation and issues zero jobs).
- `ruff check` / `ruff format --check` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Changed
+
+- **`explore map` and `explore relationships` skip re-profiling an object whose
+  cached profile is still fresh.** Before scanning a selected object, each
+  command now checks `.dex/cache.json` for a same-connector profile of that
+  exact object that was profiled within a freshness window and whose column
+  signature (name, type, nullability) still matches the warehouse's free
+  metadata; a match is reused wholesale instead of re-scanned, so it never
+  enters the cost preflight or the billed handshake. Iterative workflows on a
+  metered warehouse (map, tweak, map again) and `--verify` re-runs no longer
+  re-pay the full profiling scan when nothing changed. The freshness check is
+  fail-closed: a missing or unparseable `profiled_at`, a schema change, or a
+  different connector re-profiles. The envelope gains a `cache_hit_count` field
+  (distinct from the existing `carried_forward_count`, which covers
+  below-rank-cutoff objects) and a note when reuse happened.
+- **New `--refresh` flag on `explore map` / `explore relationships`** forces a
+  full re-profile of every selected object even when the cache is fresh, for
+  callers who know the source changed in a way the cheap metadata check cannot
+  see.
+- **New `profile_freshness_hours` config knob** (`DexConfig`, default `24.0`)
+  sets how fresh a cached profile must be to be reused; `0` disables reuse
+  (always re-profile). No cache schema change: `Dataset.profiled_at` and the
+  stored column signatures already carry everything the check needs.
+
 ### Fixed
 
 - **Redshift connections survive a Serverless cold start.** An idle Serverless
--- a/packages/dex-core/src/exmergo_dex_core/cli.py
+++ b/packages/dex-core/src/exmergo_dex_core/cli.py
@@ -123,6 +123,14 @@ def _build_parser() -> argparse.ArgumentParser:
                     sp.add_argument(
                         "--verify", action="store_true", default=argparse.SUPPRESS
                     )
+                # Force a full re-profile even when the cache holds a fresh,
+                # schema-matching profile for a selected object (the default is
+                # skip-if-cached; --refresh is the escape hatch when the source
+                # changed in a way the cheap metadata check cannot see).
+                if group == "explore" and name in {"map", "relationships"}:
+                    sp.add_argument(
+                        "--refresh", action="store_true", default=argparse.SUPPRESS
+                    )
                 # Exploration starts bare: warehouse truth, independent of
                 # whatever repo dex runs from. --use-project opts in to folding
                 # the project's declared definitions (joins, grain, metric
--- a/packages/dex-core/src/exmergo_dex_core/config.py
+++ b/packages/dex-core/src/exmergo_dex_core/config.py
@@ -279,6 +279,9 @@ class DexConfig(BaseModel):
     # How many top-ranked objects `explore map` deep-profiles on a large
     # warehouse; the rest stay inventory-only. Selective by default, overridable.
     profile_top_n: int = 25
+    # How fresh a cached profile must be to skip re-scanning it (`explore map` /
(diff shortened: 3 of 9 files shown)


marcociav-exmergo (approved):
LGTM!

Let's monitor whether coding agents pick up this default 24h`profile_freshness_hours` in the config and change it according to the situation and/or user request

PR #99, opened 2026-07-17
SCRUM-934-suppress generic-name relationship collisions before verify

### LINKED ISSUE #77: Relationship inference over generically-named ID columns produces near-total noise on CDC-shaped sources
## Summary

On a warehouse where many unrelated entities share a generically-named id
column (the norm for Firestore/Mongo/DynamoDB-style CDC exports, where every
table has e.g. `document_id`), relationship inference generates a very
large number of candidate relationships, essentially all of them spurious --
and verifying them consumes a large, unbudgeted amount of the verify cost
(see the linked verify-cost-estimation issue).

## Observed

Across ~90 objects sharing a single generic id-column name, inference found
448 candidate matches (after folding 1,959 same-lineage duplicates).
Verifying a subset confirmed effectively all of them were spurious (100%
orphaned -- zero actual row overlap). This is not an edge case: generic
per-collection id/name columns are the norm for this entire class of
CDC-exported source, which is a common real-world input to this connector.

## Expected

A cheap pre-filter before spending verify budget on a candidate --
something like a minimum name-specificity heuristic, or a much cheaper
approximate-overlap sketch computed before committing to a full verify
probe -- so verify isn't spent confirming what a simpler, near-free check
could already have ruled out.

## Impact

On this class of source, relationship verification is close to worthless
signal-to-cost: nearly all of the (real, metered) spend goes toward
confirming non-relationships. The same checkpoint proposed in the
verify-cost-estimation issue is a natural place to also surface something
like "448 candidates found, most look like generic-name collisions, narrow
before verifying?".

## Suggested fix

Add a pre-verify filter stage to relationship inference that scores
candidates by name-specificity (or a cheap approximate-cardinality-overlap
check) before committing verify budget to them, and surface the
score/reasoning in the same checkpoint proposed for verify-cost visibility.

---

**Environment**
  datasets, ~90 tables total, generic per-entity id column naming)

emnarahmouni-exmergo:
## Summary
- On a warehouse where ~90 unrelated tables share a generic per-collection id
  column (the norm for Firestore/Mongo/DynamoDB-style CDC exports, e.g.
  `document_id`), relationship inference's "same-named foreign key" tier
  matched every such pair as a candidate join, purely because they shared a
  column name — no entity-name evidence involved. This generated a
  near-complete cross product (448 candidates in the reported case) that
  `--verify` would then spend real query cost confirming as spurious.
- Added a cheap, metadata-only pre-filter: a same-named-FK match is withheld
  when its column name is held as a single-column key by
  `_GENERIC_NAME_MIN_HOSTS` (3) or more unrelated datasets. Below that, a
  shared name is still a valid same-named-FK signal, so ordinary two-table
  cases are unaffected.
- Withheld matches are recorded and surfaced as a note on `explore
  relationships`/`explore map` (count, shared name(s), host count), so the
  filtering is visible instead of the result silently looking like "nothing
  more to find."
- The other two matching tiers didn't need changes: the entity-name tier
  already ties a match to one specific parent by name, and the dealiased
  tier already refuses a match that collapses to a bare suffix.

tested the issue before : 
Screenshot 2026-07-17 110847

After the fix: 

Screenshot 2026-07-17 111511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   still fails immediately), across a window wide enough to cover the wake. A
   per-attempt connect timeout bounds a stalled handshake and is cleared once the
   connection is up so it never caps a later billed query's result read.
+- **Relationship inference no longer floods `--verify` with generic id-column
+  collisions** ([#77]). On warehouses where many unrelated tables share a
+  generic id-shaped column name (the norm for Firestore/Mongo/DynamoDB-style
+  CDC exports, e.g. every collection has its own `document_id`), name-based
+  inference matched every such pair as a candidate join, spending real verify
+  query cost confirming what was, essentially always, a naming convention
+  rather than a relationship. A same-named-FK match is now withheld when its
+  column name is held as a key by three or more unrelated datasets, and the
+  withheld count and names are surfaced in `explore relationships`/`explore
+  map`'s notes instead of silently inferring less.
 
 ## [1.2.1] - 2026-07-17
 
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
Yeah this is very useful, approved!

PR #104, opened 2026-07-19
SCRUM-942-Fix transform build swallowing dbt's real error on every connector

### LINKED ISSUE #76: BigQuery: `transform build` still surfaces a dbt deprecation warning instead of the actual error (parity gap vs #50/#55)
## Summary

#50 and #55 (both closed) describe and fix `transform build`'s error
channel getting poisoned by a dbt deprecation-warning line instead of the
real failure. Both were verified against the **Snowflake** connector only.
This session reproduced the identical symptom fresh, on **BigQuery**, days
after both were closed.

## Repro

1. Author a model whose compiled SQL is invalid in a way BigQuery rejects
   at execution time (e.g. a JSON-parsing function called with a dynamic,
   non-literal path argument -- BigQuery: "Argument 2 to JSON_VALUE must be
   a constant expression").
2. Run `transform build --target dev --confirm --budget <n>`.

## Actual

```
{"status": "error", "data": {"target": "dev", "success": false, "returncode": 1, "nodes": [ ... , {"name": "<the broken model>", "status": "error", ...} ], "counts": {...}}, "warnings": [ ... "Failure in model <name> (models/.../<name>.sql)", "  Database Error in model <name> (models/.../<name>.sql)" ...], "errors": ["dbt build failed: [WARNING][MissingArgumentsPropertyInGenericTestDeprecation]: Deprecated"]}
```

The top-level `errors[]` is the deprecation-warning line, not the failure.
The `warnings[]` array names the failing model but not *why* it failed --
the actual database error text ("Argument 2 to JSON_VALUE must be a
constant expression") only appears in `logs/dbt.log`, and had to be found by
grepping it directly.

A second instance in the same session was worse: a hard compile-time
failure (a Jinja context error) also surfaced only the same generic
deprecation-warning line as `errors[]`, with **no** mention of the real
compilation error anywhere in the envelope. That invocation's `nodes`/
`counts` payload was also suspiciously identical to a prior *successful*
run (matching per-node execution times almost exactly), while
`logs/dbt.log` showed the actual invocation finished in well under a second
and failed at compile time before touching the warehouse at all -- worth
checking whether the envelope returned stale/cached data rather than
reflecting that specific (near-instant, hard-failed) invocation.

## Expected

`errors[]` (and, for model-level failures, `warnings[]`) should contain
dbt's actual structured error/failure message for the node that failed, not
a deprecation-warning line -- matching what #50 already specifies, just
extended to the BigQuery connector.

## Impact

Every one of the 10 most recently closed issues in this tracker with an
"Environment" section lists **Snowflake** as the connector. This is the
second time in that set a Snowflake-verified fix has needed an explicit
connector-parity follow-up (see #60 -> #48, #61 -> #51) -- suggesting
Snowflake-only dogfooding before closing an issue is a recurring gap in this
project's verification process, not just this one bug.

## Suggested fix

Confirm whether the fix for #50 lives in a connector-agnostic error
extractor or something Snowflake-adapter-specific, and if the latter,
replicate it for the BigQuery adapter. Separately, investigate the
stale-data observation above -- it may be the same root cause (a poisoned
error/result-reporting path) or a distinct issue.

## Note

This is a follow-up to #50 and #55, which fixed/addressed this for the
Snowflake connector only.

---

**Environment**
- dbt `1.11.12`, `dbt-bigquery` (adapter `1.11.3`)
  datasets, ~90 tables total)

emnarahmouni-exmergo:
- #50/#55 fixed transform build's errors[0] getting poisoned by a dbt
  deprecation-warning line instead of the real failure, but were verified
  against Snowflake only. Reproducing the identical symptom on BigQuery
  showed the actual bug is broader than a missing connector-specific port —
  it's connector-agnostic and was never fully fixed for *any* adapter.
- Root cause, confirmed against real `dbt build --log-format json` output
  (not hand-written test fixtures): dbt wraps a failure's real cause behind
  one or more generic, information-free headers, identical on every
  connector since the shape comes from dbt_common:
  - A per-node failure has no `MainEncounteredError` at all. Instead dbt
    fires, in order, a bare progress line (`LogModelResult`), a bare
    `"Failure in model x (path)"` header (`RunResultFailure`), and only
    then `RunResultError`, whose `msg` is the actual exception text. Nothing
    marks the first two as deprecations, so whichever logs first — usually
    the useless one — won `errors[0]`.
  - Even the right message could lose its cause: `"<Type> Error in <node>
    (<path>)"` puts the real cause on the *next* line; a whole-invocation
    fatal wraps again in `"Encountered an error:"`; a nested exception chain
    repeats a bare `"<Type> Error"` once per level. The old code kept only
    `msg.splitlines()[0]`.
- Fix (`transform/build.py`):
  - `RunResultError` (the per-node cause event) is now promoted the same
    way `MainEncounteredError` (the run-level cause event) already was.
  - A new `_GENERIC_HEADER` walks past chained wrapper lines to find the
    first real content line and appends it to the header instead of
    dropping it.
  - ANSI color codes dbt bakes into messages even under `--log-format json`
    are stripped before they reach the envelope.
  - `build()` now clears any stale `target/run_results.json` before
    invoking dbt: a whole-invocation fatal never rewrites that file, so a
    prior successful build's node results could otherwise be misreported as
    belonging to the current (failed, near-instant) invocation — confirmed
    by comparing real file mtimes before/after such a failure.
Before the fix: 
Screenshot 2026-07-19 121651
After fix:
Screenshot 2026-07-19 121301

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,27 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   still fails immediately), across a window wide enough to cover the wake. A
   per-attempt connect timeout bounds a stalled handshake and is cleared once the
   connection is up so it never caps a later billed query's result read.
+- **`transform build` names dbt's real error on every connector, not just
+  Snowflake** ([#76], a connector-parity follow-up to [#50]/[#55]). dbt wraps
+  a failure's actual cause behind one or more generic, information-free
+  headers -- a per-node failure as "<Type> Error in <node> (<path>)", a
+  whole-invocation fatal again in "Encountered an error:", a nested exception
+  chain once more per level -- and, for a per-node failure specifically, logs
+  a bare progress line and a bare "Failure in <node> (<path>)" header before
+  the message that actually names the cause. This shape is identical on every
+  adapter (it comes from dbt_common, not a connector), but keeping only the
+  first captured line let whichever of these uninformative lines happened to
+  log first silently win the envelope's `errors[0]` slot, on Snowflake as much
+  as BigQuery -- #50/#55's own repro just never happened to hit it. The real
+  cause line now rides alongside its header instead of being dropped, and
+  dbt's own per-node/per-run "this is what actually failed" events are
+  promoted ahead of a progress line or bare header the same way the #50 fix
+  already promoted them ahead of a deprecation notice. Also fixed in the same
+  pass: a stale `target/run_results.json` left over from a prior successful
+  build, which a whole-invocation fatal never rewrites, is now cleared before
+  each build so it can never be misreported as this invocation's node
+  results; and ANSI color codes dbt bakes into its messages even under
+  `--log-format json` no longer leak into the envelope.
 
 ## [1.2.1] - 2026-07-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,15 +22,15 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 - **`transform build` names dbt's real error on every connector, not just
   Snowflake** ([#76], a connector-parity follow-up to [#50]/[#55]). dbt wraps
   a failure's actual cause behind one or more generic, information-free
-  headers -- a per-node failure as "<Type> Error in <node> (<path>)", a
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
LGTM

PR #118, opened 2026-07-20
Issue-108: Blob-column pruning

### LINKED ISSUE #108: `explore profile` scans large unprofilable BYTES columns in full; add type/size-aware column pruning
## Summary

Profiling a table scans all columns, including large `BYTES`/`ARRAY<BYTES>`
blob columns whose profile output can only ever be null fraction and a
distinct estimate. On blob-heavy tables those columns dominate the scan
cost while contributing almost nothing to the profile.

## Observed

A Datastore-export snapshot table (~171 MB, 35 columns, of which 7 are
BYTES / ARRAY<BYTES> serialized-state blobs) profiled at an estimated
158,557,788 bytes -- ~93% of the table's total size, dominated by the blob
columns. The 28 informative columns (ids, categoricals, numerics,
timestamps) are a small fraction of that. The profile of the blob columns
returned null fraction and a distinct count -- real but marginal signal
relative to their share of the spend.

## Expected

Type/size-aware pruning at profile time: skip (or sample-bound) columns
whose type can't produce a meaningful profile (BYTES over a size
threshold, deeply nested ARRAY<BYTES>), and name every skipped column in
the envelope's `notes` -- the same "excluded columns are named, check the
notes" convention `explore cluster` already uses for keys. An explicit
opt-in (`--include-blobs` or a per-column config) covers the rare case
where a blob's null/distinct stats matter.

## Impact

Blob-heavy operational tables (serialized agent/session state, ML
artifacts, raw event payloads) are common in exactly the sources worth
exploring. Cutting their profile cost several-fold with zero
decision-relevant signal loss compounds with the billing-floor issue filed
alongside this one: together they made an 8-table profile cost ~4x what
the informative columns alone would have.

## Suggested fix

In the profile planner, classify columns by (type, estimated width) before
building scan queries; route oversized blob columns to a
null-fraction-only pass over a bounded sample (or skip them), and report
the pruning in `notes`.

---

**Environment**

### LINKED ISSUE #122: `profile_estimate` batch count on Postgres/Snowflake/Databricks/Redshift doesn't shrink when blob columns are pruned from the scan
## Summary

#108 excluded blob-type (`BYTES`/`BLOB`/`bytea`/`BINARY`) columns from the
profiling scan itself, and corrected BigQuery's `profile_estimate` (its
dry-run cost estimate) to match. The other four billed adapters
Postgres, Snowflake, Databricks, Redshift each have their own
`profile_estimate`, and none of them were updated: their batch count is
still computed from the full, unpruned column list.

## Observed

Postgres/Snowflake/Databricks/Redshift's `profile_estimate` is a coarse
heuristic, not a literal per-column dry run: `batches = ceil(len(columns) /
50) `, then `estimate = batches * per_batch_time(meta.byte_size)`. It never
priced a column's bytes individually, so it never had BigQuery's "one
column dominates the number" problem. But `len(columns)` is still the
pre-exclusion count. A table whose blob columns push it just over a
50-column batch boundary (e.g. 51 columns, 3 of them blob-typed) reports
one more batch's worth of estimated time than the pruned scan will
actually issue.

## Expected

Each of these four adapters' `profile_estimate` should compute `batches`
from the same post-exclusion column list `explore.profile.profile` will
actually hand to `column_aggregates` i.e. filter with
`adapters.base.is_blob_type` (honoring any `include_blobs` override) before
the `batches = ceil(...)` line, the same shape as the fix already applied
to `BigQueryAdapter.profile_estimate`.

## Impact

Small in practice (at most one batch's worth of estimated time per table,
only when exclusion crosses a batch boundary), but the estimate a user
confirms against should match what the run will actually do the same
principle #108 fixed for BigQuery. Low risk, mechanical, same shape in all
four files.

## Suggested fix

In `postgres.py`, `snowflake.py`, `databricks.py`, `redshift.py`:
`profile_estimate` gains the same `include_blobs: set[str] | None = None`
kwarg as `BigQueryAdapter.profile_estimate`, filters `columns` to
`scan_columns` via `is_blob_type` before computing `batches`, and
`commands.py`'s existing `blob_paths` threading (already updated for
BigQuery in #108) covers these adapters too since `_profile_estimate`
already forwards `include_blobs` generically.

---
Follow-up from #108.

spascarella:
## Summary

- `explore profile` scanned every column of a table, including `BYTES`/`BLOB`-shaped columns (and `ARRAY<BYTES>`/`BLOB[]`) whose profile can only ever be a null fraction and a distinct estimate. On a columnar warehouse, referencing such a column at all bills for its full stored bytes, so on blob-heavy tables these columns dominated scan cost for almost no decision-relevant signal (one reported case: ~93% of estimated scan bytes from 7 of 35 columns).
- Blob-type columns are now excluded from the profiling scan by default. The exclusion is type-based (classified from the already-fetched catalog type string — no adapter exposes a free per-column byte size, so a size threshold isn't possible) and applies uniformly across connectors because the filtering happens once in `explore/profile.py`, upstream of every adapter's SQL builder.
- Excluded columns are named in the dataset's `data_quality` notes, the same "excluded columns are named, check the notes" convention `explore cluster` already uses.
- A durable per-column opt-in (`blob_overrides` in `.dex/config.yml`) restores real stats for a specific column, mirroring the existing `pii_overrides` mechanism exactly.
- Every connector's `profile_estimate` (the pre-execution cost estimate shown at the billed handshake) applies the same exclusion, so the estimate the user confirms against matches what actually runs. This required each of Postgres, Snowflake, Databricks, and Redshift to accept a new `include_blobs` parameter too — without it, `explore profile`/`relationships`/`map` crashed with a `TypeError` against all four, since `explore/commands.py` calls `profile_estimate(identifiers, include_blobs=...)` unconditionally on any adapter that exposes it.

## Related Issues
Closes #108, closes #122

## Changes

- `adapters/base.py`: `is_blob_type()` classifier (BYTES/BLOB/BYTEA/BINARY, substring-matched like `is_numeric_type`).
- `config.py`: `BlobOverride`, `blob_override_paths()`, `DexConfig.blob_overrides`.
- `explore/profile.py`: `profile()` filters blob columns out of the scan (unless overridden) and adds a `data_quality` note naming what was skipped.
- `explore/commands.py`: threads the override set through `cmd_profile`, `cmd_relationships`, and `cmd_map`.
- `adapters/bigquery.py`, `adapters/postgres.py`, `adapters/snowflake.py`, `adapters/databricks.py`, `adapters/redshift.py`: `profile_estimate` on each accepts `include_blobs` and filters the same way, so every connector's cost estimate reflects the pruned scan.
- No changes needed to any adapter's actual scan-building SQL (`column_aggregates`) — the exclusion happens once, upstream, in `explore/profile.py`, before that's ever called.

## Test plan

- [x] `tests/test_config.py` — `BlobOverride`/`blob_override_paths` round-trip, case-insensitivity, clean-when-unset
- [x] `tests/explore/test_profile.py` — default exclusion + `data_quality` note, override restores stats
- [x] `tests/adapters/test_connect_bigquery.py` — `profile_estimate` excludes a `BYTES` column's bytes by default, restores it under `include_blobs`
- [x] `tests/adapters/test_connect_postgres.py`, `test_connect_snowflake.py`, `test_connect_databricks.py`, `test_connect_redshift.py` — regression test reproducing the exact crash (`profile_estimate(identifiers, include_blobs=set())`), plus batch-count accuracy tests
- [x] Full suite: 1121 passed, 78 skipped
- [x] `ruff check` clean on all touched files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,20 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   its only batched filtered-count spelling, and it was the one form the
   firewall still refused. The refusal message's example list and the probe
   playbook now name `COUNTIF` alongside `COUNT` (#105).
+- **`explore profile` no longer scans blob-type columns by default.**
+  `BYTES`/`BLOB`/`bytea`/`BINARY` columns, scalar or repeated, are excluded
+  from the aggregate scan across every connector (DuckDB, BigQuery,
+  Snowflake, Databricks, Postgres, Redshift): their profile can only ever be
+  a null fraction and a distinct estimate, yet a columnar engine bills for a
+  column's full stored bytes once it is referenced at all, so blob-heavy
+  tables had these columns dominating scan cost for negligible signal.
+  Excluded columns are named in the dataset's `data_quality` notes, the same
+  convention `explore cluster` already uses for excluded keys. A new
+  `blob_overrides` list in `.dex/config.yml` (mirroring `pii_overrides`)
+  restores real stats for a specific column when they matter. Every
+  connector's `profile_estimate` reflects the same exclusion, so the
+  pre-execution cost estimate matches what the pruned scan actually runs
+  (#108).
 - **`explore query` now resolves CTE aliases across set operations.** `WITH`
   clause relations attached to `UNION`, `INTERSECT`, or `EXCEPT` roots are
   registered before either branch is inspected, including later CTEs that
--- a/packages/dex-core/src/exmergo_dex_core/adapters/databricks.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/databricks.py
@@ -43,6 +43,7 @@ from .base import (
     QueryResult,
     blame,
     distinct_combination_sql,
+    is_blob_type,
     json_safe,
     name_list,
     shape_stat_expressions,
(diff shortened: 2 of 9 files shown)


marcociav-exmergo (approved):
LGTM! thanks @spascarella
Just some minor changelog merge conflicts

PR #119, opened 2026-07-20
Issue 106

### LINKED ISSUE #106: `pii_overrides` needs a pattern form: one logical column costs O(tables x environments) entries
## Summary

`pii_overrides` entries (shipped via #54) are fully-qualified single
columns. On CDC-shaped sources, the same structurally identical column
exists on every entity's changelog in every environment mirror, so one
logical false positive costs one entry per table per environment, forever.

## Observed

Clearing the `document_name` resource-path false positive (see #54) for
seven changelog tables in one dev dataset took seven near-identical
entries. The same column exists on every changelog in the qa and prod
mirror datasets too: profiling those adds ~14 more identical entries, and
every newly landed entity adds one per environment. The override list
becomes a maintenance surface that grows with the schema while encoding a
single human decision ("this CDC export's `document_name` is a resource
path").

## Expected

An opt-in pattern form alongside the existing exact form, e.g.:

```yaml
pii_overrides:
  - column: "<project>.raw_*_dev|qa|prod.*.document_name"   # glob, or
  - column_name: document_name                              # scoped by
    scope: "<project>.raw_*"                                # dataset glob
    reason: Firestore resource path, not a person name
```

Exact-match should stay the default; the pattern form is for the "same
false positive, N structurally identical tables" case, which is the norm
for Firestore/Mongo/DynamoDB-style exports (every table shares
`document_name`/`document_id`-class columns by construction).

## Impact

The override mechanism works well (this is a follow-up, not a complaint --
it unblocked a whole session's path analysis). But per-column growth means
teams will either fall behind (new tables re-hit the wall) or paste
boilerplate, and a reviewer can no longer see the single decision behind 20
lines of entries.

## Suggested fix

Support a glob/pattern column form (or a `column_name` + dataset-scope
pair) in `pii_overrides`, resolved at query/profile time against the
qualified column, with the same audit fields as exact entries.

---

**Environment**
- dex plugin `1.2.1`, BigQuery connector
  datasets, ~100 objects)

spascarella:
## What this changes

`pii_overrides` entries were exact fully-qualified columns only. On CDC-shaped
sources (Firestore/Mongo/DynamoDB-style exports) the same structurally
identical column, for example `document_name`, exists on every entity's
changelog table in every environment mirror. Clearing one false positive cost
one config entry per table per environment, forever growing with the schema
for a single human decision.

This adds an opt-in pattern form alongside the existing exact form:

```yaml
pii_overrides:
  - column: MY_DB.PUBLIC.REGION.R_NAME       # exact, unchanged
    reason: TPC-H region labels; reviewed
  - column_name: document_name                # new: pattern
    scope: "<project>.raw_*"
    reason: Firestore resource path, not a person name
```

- `PIIOverride` now accepts either `column` (exact) or `column_name` + `scope`
  (pattern, a dataset-identifier glob), enforced by a validator so a config
  entry cannot mix the two shapes or supply half of the pattern pair.
- `pii_override_paths()` returns a `PIIOverrideMatcher` that duck-types the
  exact-match set every call site already used (`path in matcher`), so the
  profiling and drift-reconciliation match sites needed no changes.
- The profile-time typo guard (`_override_mismatches`) now understands
  pattern entries: it warns when a scope matches profiled tables but none
  carries the named column, and stays silent when a scope matches nothing yet
  (new entities landing later under the same scope is the point of the
  pattern form).
- Fixed a latent bug in `maintain/commands.py`, which built its override set
  inline instead of calling `pii_override_paths()`. That inline code would
  have raised on a pattern entry, since it assumed every entry had a `column`.

`blob_overrides` has the identical exact-match-only shape and could hit the
same per-table, per-environment growth problem. It is left untouched here
since #106 only asked about `pii_overrides`; worth its own follow-up issue if
it comes up in practice.

## Related issues

Closes #106.

## Notes

Merge #118 first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,20 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **`pii_overrides` gains an opt-in pattern form** (#106). Alongside the
+  existing exact `column` entry, a `column_name` + `scope` entry clears a
+  named column on every table whose fully-qualified identifier matches the
+  `scope` glob, so one reviewed decision (for example, "this CDC export's
+  `document_name` is a resource path, not a person name") no longer costs one
+  config entry per table per environment on Firestore/Mongo/DynamoDB-style
+  sources, where the same column exists by construction on every entity's
+  table in every environment mirror. `column` and `column_name`/`scope` are
+  mutually exclusive on one entry (enforced at load). The profile-time typo
+  guard now covers pattern entries too: it warns when a `scope` matches
+  profiled tables but none carries the named column, and stays silent when
+  the scope matches no table yet, since new entities landing later under the
+  same scope is the point of the pattern form. `blob_overrides` keeps its
+  exact-only shape for now.
 - **`transform init --layered-schemas`: per-layer schema routing out of the
   box.** The flag additionally scaffolds `models/intermediate/`, a
   `generate_schema_name` macro override, and a `dbt_project.yml` `models:`
--- a/packages/dex-core/src/exmergo_dex_core/adapters/base.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/base.py
@@ -170,6 +170,25 @@ def json_safe(value: object | None) -> object | None:
     return str(value)
 
 
+# Substring hints for arbitrary binary-blob column types across dialects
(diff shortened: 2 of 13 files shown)


marcociav-exmergo (approved):
Nice

PR #121, opened 2026-07-21
Fix multi-CTE resolution across set operations

### LINKED ISSUE #117: `explore query` firewall stops resolving CTE aliases when the WITH clause has more than one CTE
## Summary

`explore query`'s firewall resolves FROM-clause names against the `.dex`
cache, but with more than one CTE in the `WITH` clause it stops recognizing
CTE aliases as CTEs: the query is refused with `'<cte>' is not in the .dex
cache; run explore map (or explore profile <cte>) first` -- advice that
cannot be followed, since the name is a query-local alias, not a table.
Exactly one CTE works fine, including correlated subqueries against it.

## Repro (BigQuery connector, all base tables profiled and PII-cleared)

Works -- single CTE, outer SELECT over it, correlated subquery back into it:

```sql
WITH m AS (SELECT <expr> AS p, <expr> AS db FROM `<project>.<dataset>.<table>`)
SELECT COUNT(DISTINCT IF(db = 'x', p, NULL)) AS a,
       COUNT(DISTINCT CASE WHEN db != 'x'
             AND p IN (SELECT p FROM m WHERE db = 'x') THEN p END) AS b
FROM m
```

Refused -- second CTE references the first; the refusal names the second CTE:

```sql
WITH m AS (SELECT ... FROM `<project>.<dataset>.<table>`),
     imported AS (SELECT DISTINCT p FROM m WHERE db = 'x'),
     live     AS (SELECT p, op FROM m WHERE db != 'x')
SELECT ... FROM live
```

```
query refused: 'imported' is not in the .dex cache; run `explore map` (or `explore profile imported`) first so its columns and PII flags are known
```

Refused -- five *independent* CTEs (each a plain SELECT from a different
profiled table) combined with UNION ALL of per-CTE aggregates; the refusal
names the **first** CTE:

```sql
WITH s AS (SELECT ... FROM `<project>.<dataset>.<table_1>`),
     c AS (SELECT ... FROM `<project>.<dataset>.<table_2>`),
     ...
SELECT 'label1', COUNT(...) FROM s
UNION ALL
SELECT 'label2', COUNT(...) FROM c
...
```

```
query refused: 's' is not in the .dex cache; run `explore map` (or `explore profile s`) first so its columns and PII flags are known
```

So it is not (only) CTE-referencing-CTE: any second CTE appears to knock
out alias registration entirely.

## Expected

CTE aliases are query-local relations; the firewall should resolve them to
their defining subqueries (which it already does for the single-CTE case)
and check PII/projection rules against the underlying base tables,
regardless of CTE count or CTE-to-CTE references.

## Impact

- Multi-table comparison probes (the "same measure across N tables" shape,
  natural for environment-mirrored CDC datasets) must be decomposed into
  one probe per table: N round trips, each billing the 10,485,760-byte
  BigQuery floor (the economics documented in #107). In the motivating
  session a 5-table UNION probe became 5 floor-billed probes.
- Set-overlap probes between two derived row-sets from the *same* table
  must be contorted into one CTE with `IF()`/`CASE` branches -- workable,
  but the rewrite is non-obvious from the refusal.
- The refusal message misdiagnoses: it names a CTE alias as an unprofiled
  table and recommends `explore profile <alias>`, which fails. Per the
  playbook's "rewrite once" contract, a wrong fix-suggestion costs the one
  rewrite the design budgets for.

## Suggested fix

Register all `WITH`-clause aliases (including aliases referencing earlier
aliases) as local relations before resolving FROM-clause names against the
cache -- the same resolution the single-CTE path already performs. Failing
that, keep the limitation but make the refusal accurate: "multi-CTE queries
are not supported; inline the CTEs or split the probe" instead of steering
users toward profiling a query alias. Related prior art on FROM-clause
resolution: #78 (UNNEST over function-derived arrays).

---

**Environment**
- dex plugin `1.2.2`, BigQuery connector, `bytes_scanned` cost paradigm
- Source: a Firestore CDC export landed in BigQuery (environment-mirrored
  datasets)

happykawayigt:
## What this changes

`WITH` clauses are attached by sqlglot to the query root. For queries ending in `UNION`, `INTERSECT`, or `EXCEPT`, that root is the set operation rather than either child `SELECT`.

This moves CTE registration into `_query_outputs`, so all query shapes share the same local relations. CTEs are processed in declaration order, which also lets later CTEs reference earlier ones.

Regression coverage includes both independent CTEs joined by `UNION ALL` and a later CTE derived from an earlier one.

## Related issues

Closes #117.

## Validation

- `pytest tests/guards/test_query_firewall.py tests/test_safety_spine.py -q -k query_firewall` (118 passed, 81 deselected)
- `ruff check src/exmergo_dex_core/guards/query_firewall.py tests/guards/test_query_firewall.py`
- `ruff format --check src/exmergo_dex_core/guards/query_firewall.py tests/guards/test_query_firewall.py`
- The full non-integration engine suite was attempted twice but exceeded the local 5-minute limit without reporting a failure.

## Checklist

- [x] Engine tests pass (`uv run pytest` in `packages/dex-core`) - focused suite passes; full suite timed out locally
- [x] Eval scoring-core tests pass (`uvx pytest evals`) - not run; evaluation code is unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,14 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   schemas are checked. Backing this, every adapter gains a
   `list_namespace_objects` metadata method alongside `missing_dev_namespaces`.
 
+### Fixed
+
+- **`explore query` now resolves CTE aliases across set operations.** `WITH`
+  clause relations attached to `UNION`, `INTERSECT`, or `EXCEPT` roots are
+  registered before either branch is inspected, including later CTEs that
+  reference earlier ones. Multi-CTE probes no longer misdiagnose query-local
+  aliases as tables missing from the `.dex` cache (#117).
+
 ## [1.2.2] - 2026-07-18
 
 ### Changed
--- a/packages/dex-core/src/exmergo_dex_core/guards/query_firewall.py
+++ b/packages/dex-core/src/exmergo_dex_core/guards/query_firewall.py
@@ -341,11 +341,22 @@ def _query_outputs(
 ) -> _Outputs:
     """Output taints of a whole query node (SELECT or a set operation)."""
 
+    # sqlglot attaches a WITH clause to the query root. For a plain query that
+    # root is a Select, but for UNION/INTERSECT/EXCEPT it is the set operation.
+    # Register CTEs here so both sides of any query shape resolve the same local
+    # relations. Process them in order because later CTEs may reference earlier
+    # ones.
+    local_env = dict(env)
+    for cte in node.ctes:
+        local_env[cte.alias_or_name.lower()] = _query_outputs(
+            cte.this, local_env, known, tables, dialect
+        )
+
     if isinstance(node, exp.Select):
-        return _select_outputs(node, env, known, tables, dialect)
+        return _select_outputs(node, local_env, known, tables, dialect)
     if isinstance(node, _QUERY_ROOTS):  # set operation: UNION/INTERSECT/EXCEPT
-        left = _query_outputs(node.left, env, known, tables, dialect)
-        right = _query_outputs(node.right, env, known, tables, dialect)
+        left = _query_outputs(node.left, local_env, known, tables, dialect)
+        right = _query_outputs(node.right, local_env, known, tables, dialect)
         # Column names come from the left side; taints merge positionally, and a
         # length mismatch (invalid SQL anyway) merges everything conservatively.
         left_items = list(left.items())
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (commented):
Hi @happykawayigt this looks great!

Let's get this out of Draft mode and into "Ready for Review" so we can let the CI tests run.

If there are no errors with the CI I will approve this PR, merge it into main, and insert these changes in the next release notes.

Thanks!

marcociav-exmergo (approved):
LGTM! Thanks for this @happykawayigt

PR #123, opened 2026-07-21
Issue 122

spascarella:
## Summary

- `explore/commands.py::_profile_estimate` (added in #108) unconditionally calls `adapter.profile_estimate(identifiers, include_blobs=...)` for any adapter that exposes `profile_estimate`. Postgres, Snowflake, Databricks, and Redshift each implement their own `profile_estimate(self, identifiers: list[str])` with no `include_blobs` parameter, so `explore profile`/`relationships`/`map` crashed with `TypeError: profile_estimate() got an unexpected keyword argument 'include_blobs'` against any of these four connectors.
- Each of the four now accepts `include_blobs: set[str] | None = None` and filters blob-type columns out of its batch-count calculation before estimating, mirroring the fix already applied to `BigQueryAdapter.profile_estimate` in #108.
- This also fixes a secondary, lower-severity gap: these adapters' `batches = ceil(len(columns) / 50)` heuristic wasn't reflecting the pruned column count either, so a table whose blob columns crossed a 50-column batch boundary reported one extra batch's worth of estimated time.

## Changes

- `adapters/postgres.py`, `adapters/snowflake.py`, `adapters/databricks.py`, `adapters/redshift.py`: `profile_estimate` gains `include_blobs`, filters `columns` to `scan_columns` via `adapters.base.is_blob_type` before computing `batches`. No other logic touched.

## Test plan

- [x] Regression test per adapter: `profile_estimate(identifiers, include_blobs=set())` -- the exact call shape that crashed -- doesn't raise
- [x] Batch-count test per adapter: a blob column crossing the 50-column batch boundary drops the estimate by one batch; `include_blobs` restores it
- [x] Full suite: 1121 passed, 78 skipped
- [x] `ruff check` clean on all touched files

---
Base: `issue-108`. This branch depends on #108's `is_blob_type`/`include_blobs` foundation; retarget to `main` once #108 merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,14 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   schemas are checked. Backing this, every adapter gains a
   `list_namespace_objects` metadata method alongside `missing_dev_namespaces`.
 
+### Fixed
+
+- **`explore query` now resolves CTE aliases across set operations.** `WITH`
+  clause relations attached to `UNION`, `INTERSECT`, or `EXCEPT` roots are
+  registered before either branch is inspected, including later CTEs that
+  reference earlier ones. Multi-CTE probes no longer misdiagnose query-local
+  aliases as tables missing from the `.dex` cache (#117).
+
 ## [1.2.2] - 2026-07-18
 
 ### Changed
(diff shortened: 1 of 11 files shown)


marcociav-exmergo (approved):
LGTM! @spascarella let's keep it pushing ‼️

PR #124, opened 2026-07-21
Issue-105: Allow COUNTIF over a PII-flagged column in the query firewall

### LINKED ISSUE #105: BigQuery: `COUNTIF` over a PII-flagged column is refused with no batched alternative -- revisit the conservatism for dialects without `FILTER (WHERE ...)`
## Summary

The query firewall refuses `COUNTIF(<condition involving a flagged
column>)` as value-carrying. This is understood to be deliberate
conservatism (the safe-aggregate allowlist is syntactic), and this issue is
explicitly a request to *revisit* that call for BigQuery, with two pieces
of evidence the original decision may not have priced in: BigQuery has no
`COUNT(*) FILTER (WHERE ...)`, so `COUNTIF` is the dialect's only batched
filtered-count spelling; and the refusal message actively misleads in this
case.

## Repro (isolated A/B/C on one flagged column, `{category: name, confidence: 0.6}`)

Passes (bare measuring aggregate):

```sql
select count(distinct name_col) as distinct_names from `<project>.<dataset>.<table>`
```

Passes (measuring aggregate over an expression):

```sql
select count(distinct upper(name_col)) as distinct_names_upper from `<project>.<dataset>.<table>`
```

Refused:

```sql
select countif(name_col like 'A%') as names_starting_a from `<project>.<dataset>.<table>`
```

```
query refused: the projection would carry values from PII-flagged column(s): <table>.name_col (name). Use a measuring aggregate over them (COUNT, APPROX_COUNT_DISTINCT, AVG(LENGTH(...))), or drop them from the output. PII values never cross the envelope. A column reviewed as not PII can be cleared durably with a pii_overrides entry in .dex/config.yml.
```

The same information is legal one condition at a time:

```sql
select count(*) as names_starting_a from `<project>.<dataset>.<table>` where name_col like 'A%'
```

## Why revisit

1. **Information class**: `COUNTIF(cond)` releases exactly what
   `COUNT(*) ... WHERE cond` releases -- a row count. The operands of the
   condition never cross the envelope. The firewall already accepts the
   WHERE form, so the conservatism buys no additional confidentiality; it
   only forbids *batching*.
2. **Dialect asymmetry**: on engines with `COUNT(*) FILTER (WHERE ...)`
   the playbook's own recipes (4 and 6) express batched filtered counts;
   BigQuery's only equivalent spelling is `COUNTIF`. If `FILTER (WHERE ...)`
   passes the firewall on those engines (worth checking -- if it's also
   refused, this issue generalizes), BigQuery users are strictly worse off
   for a pure dialect reason.
3. **Cost multiplication**: the WHERE decomposition turns one batched probe
   into one probe per condition, and small probes each bill BigQuery's
   10,485,760-byte minimum. In the motivating session, four path-structure
   probes (~5 conditions each) would decompose into ~20 floor-billed
   probes: ~5x the spend for identical information.
4. **Message accuracy** (worth fixing even if the verdict stands): the
   refusal recommends "a measuring aggregate over them (COUNT, ...)" while
   refusing a counting aggregate, and does not mention the rewrite that
   would pass (move the condition into `WHERE`). In the motivating session
   the practical unblock ended up being a `pii_overrides` adjudication --
   which exists for false positives, not as a toll on legal counting.

## Suggested fix

Either allowlist `COUNTIF(<condition>)` (and, if currently refused,
`COUNT(*) FILTER (WHERE <condition>)`) as the batched spelling of the
already-legal WHERE-filtered count -- refusing only forms whose *output
expression* carries column values -- or keep the verdict and (a) have the
refusal message name the WHERE-decomposition workaround explicitly, (b)
note the restriction in the probe playbook so recipes 4/6 aren't presented
as universally available.

---

**Environment**
  datasets, ~100 objects)

spascarella:
## Summary

- Allow `COUNTIF(cond)` over a PII-flagged column in the query firewall. `COUNTIF`/`COUNT_IF` (BigQuery, Snowflake, DuckDB) releases exactly what `COUNT(*) FILTER (WHERE cond)` already releases — a row count, with the condition never crossing the envelope — so it's now treated as a measuring aggregate instead of refused as value-carrying.
- Closes a BigQuery-specific gap: BigQuery has no `FILTER (WHERE ...)` clause, so `COUNTIF` was its only batched filtered-count spelling, and the one form the firewall still blocked.
- Updated the refusal message's example list and the probe playbook to name `COUNTIF`, and added a playbook callout that BigQuery rejects `FILTER (WHERE ...)` at the engine (recipes 4/6/8 need the `COUNTIF` spelling there).

Fixes #105.

## Test plan

- [x] `pytest tests/guards/test_query_firewall.py` — new `COUNTIF` cases (default dialect + BigQuery) pass, no existing refusals flip.
- [x] Full `dex-core` suite (`pytest -q`) — 1105 passed, 78 skipped, no regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,8 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 - **`explore query` now allows `COUNTIF(cond)` over a PII-flagged column.**
   `COUNTIF`/`COUNT_IF` (BigQuery, Snowflake, DuckDB) releases exactly what
-  `COUNT(*) FILTER (WHERE cond)` already released -- a row count, with the
-  condition never crossing the envelope -- so the firewall now treats it as a
+  `COUNT(*) FILTER (WHERE cond)` already released a row count, with the
+  condition never crossing the envelope so the firewall now treats it as a
   measuring aggregate instead of refusing it as value-carrying. This closes a
   dialect gap: BigQuery has no `FILTER (WHERE ...)` clause, so `COUNTIF` was
   its only batched filtered-count spelling, and it was the one form the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,15 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **`explore query` now allows `COUNTIF(cond)` over a PII-flagged column.**
+  `COUNTIF`/`COUNT_IF` (BigQuery, Snowflake, DuckDB) releases exactly what
+  `COUNT(*) FILTER (WHERE cond)` already released -- a row count, with the
+  condition never crossing the envelope -- so the firewall now treats it as a
+  measuring aggregate instead of refusing it as value-carrying. This closes a
+  dialect gap: BigQuery has no `FILTER (WHERE ...)` clause, so `COUNTIF` was
+  its only batched filtered-count spelling, and it was the one form the
+  firewall still refused. The refusal message's example list and the probe
+  playbook now name `COUNTIF` alongside `COUNT` (#105).
 - **`explore query` now resolves CTE aliases across set operations.** `WITH`
   clause relations attached to `UNION`, `INTERSECT`, or `EXCEPT` roots are
   registered before either branch is inspected, including later CTEs that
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
LGTM, just fix CI check and we're good!
Thx @spascarella

PR #126, opened 2026-07-21
Issue-112: Type-aware PII de-rating: stop flagging non-string columns as EMAIL/N…

### LINKED ISSUE #112: PII detection: de-rate name-based email/phone flags on non-string columns (INT64 `*_email_count` flagged email@0.9)

## Summary

The PII name heuristic flagged a column named `<x>_email_count` as `email` at **0.9 confidence** in every mart it appears in. The column is an **INT64** — and it is literally the PII-safe derived replacement for a staging-only array of email addresses (the mart exposes only the count, by design). The 0.9 flag blocks value-carrying aggregates (SUM/MAX/etc.) on an integer count.

## Why this is fixable in the existing design

Profiling already runs value-shape corroboration that adjusts name-based `name` flags in both directions (person-shaped values corroborate, reference vocabularies de-rate). Two cheap signals would kill this class of false positive:

1. **Type evidence:** an INT64/FLOAT/BOOL/DATE column cannot hold an email address or phone number. Category flags like `email`/`phone` on non-string columns should be de-rated below the blocking threshold categorically (or not emitted at all).
2. **Suffix evidence:** `_count`, `_num`, `_total`, `_sum` are strong negative evidence for value-carrying PII even on string columns.

## Workaround and its cost

`pii_overrides` works (2 entries, reviewable, effective immediately) — but note the interaction with #106: this project is now at **9 override entries for 2 logical columns** (7 of them the same resource-path column repeated per changelog table, plus 2 for this count column across marts). Over-flagging and per-column overrides multiply each other; type-aware de-rating attacks the left factor.

spascarella:
Closes #112

## Problem

The name heuristic flagged an integer `<x>_email_count` as `EMAIL` at 0.9 in every
mart it appeared in. That column is the PII-safe derived replacement for a
staging-only array of addresses, yet at 0.9 it sat above `PII_BLOCK_CONFIDENCE`, so
the query firewall refused value-carrying aggregates (MAX, MIN, ANY_VALUE,
percentiles) on a plain integer count and `is_min_max_safe` stripped its min/max.
The `pii_overrides` workaround worked but did not scale (9 override entries for 2
logical columns).

All PII logic is connector-agnostic and lives in `explore/profile.py`, so the column
flagged identically on every connector. The design already gated the weak
generic-name and free-text patterns to string columns; the exact-token table never
received the same treatment.

## Changes

All in `packages/dex-core/src/exmergo_dex_core/explore/profile.py`:

1. **Per-category type gate** (`_type_permits`). Type evidence is structural and
   known before any scan, so an impossible pairing suppresses the flag at
   classification time rather than being re-scored. Reuses the existing
   `_is_string_type` / `_BOOLEAN_HINTS` / `_TEMPORAL_HINTS` predicates.
   - `EMAIL` / `NAME` / `FREE_TEXT`: string only.
   - `PHONE`: anything but boolean or temporal (phone-as-INT still flags).
   - `ADDRESS` / `GOVERNMENT_ID` / `FINANCIAL` / `LOCATION` / `DOB`: unchanged, so
     `zip`, `ssn`, `salary` as INT, `lat`/`lng` as FLOAT, `dob` as DATE keep flagging.

2. **Aggregate-suffix gate** (`_is_derived_statistic`). A non-string column whose
   name ends in `_count`, `_cnt`, `_sum`, `_avg`, `_pct`, or `_ratio` is a derived
   statistic, not a value, so it is suppressed even for categories the type gate
   still permits on numbers (`ssn_count`, `zip_count`). `_num` and `_total` are
   deliberately excluded.

   Both gates `continue` past a match (never `return None`), so a later pattern
   still gets a fair look.

3. **Snowflake `FIXED`**. Snowflake's `SHOW COLUMNS` surfaces every integer/NUMBER
   as `FIXED`, which matched no numeric hint, leaving the gates inert there.
   `FIXED` is now in `_NUMERIC_HINTS`.

## Side effect worth noting

`is_numeric_type` is also read by `is_min_max_safe` and by `explore cluster` feature
selection. Fixing `FIXED` means Snowflake numeric columns now surface min/max and
become eligible cluster features. That is the correct pre-existing behavior (every
other connector already did this) but it is a visible change beyond the bug, and it
gets its own CHANGELOG bullet.

## Testing

- `tests/explore/test_profile.py`: extended the parametrized blocks with the de-rate
  cases and the must-not-regress cases (`zip`/`ssn`/`salary`/`lat`/`dob`/`phone`),
  plus assertions that min/max is restored on `user_email_count` and that
  `is_numeric_type("FIXED")` is `True`. 87 passed.
- `tests/test_safety_spine.py`: 89 passed. Genuinely sensitive numeric columns still
  block.
- Full suite (all connector extras + cluster): 1167 passed, 60 skipped.
- ruff: clean.

## Scope

No adapter changes. No config-schema changes. `pii_overrides` is untouched and
existing entries keep working; they simply stop being necessary for this class of
column.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,32 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **`explore profile` no longer flags non-string columns as EMAIL/NAME/PHONE,
+  nor aggregate-count columns as PII.** PII classification is now type-aware: a
+  category that cannot structurally live on a column's type is suppressed at
+  classification time rather than flagged and then worked around. An integer
+  `<x>_email_count` (the PII-safe derived replacement for a staging-only array
+  of addresses) is no longer flagged `EMAIL` at 0.9, so the query firewall stops
+  refusing value-carrying aggregates on it and its min/max are surfaced again.
+  The gate is per-category impossibility, not blanket: `EMAIL`/`NAME`/`FREE_TEXT`
+  are string-only, `PHONE` excludes only boolean and temporal (a phone-as-INT
+  still flags), and `ADDRESS`/`GOVERNMENT_ID`/`FINANCIAL`/`LOCATION`/`DOB` keep
+  flagging on numeric and temporal types where they legitimately belong (`zip`,
+  `ssn`, `salary` as `INT`, `lat`/`lng` as `FLOAT`, `dob` as `DATE`). Separately,
+  a non-string column whose name ends in an aggregate suffix (`_count`, `_cnt`,
+  `_sum`, `_avg`, `_pct`, `_ratio`) is treated as a derived statistic and
+  suppressed even for those categories, so `ssn_count` and `zip_count` no longer
+  flag. `pii_overrides` still works and existing entries are untouched; they
+  simply stop being necessary for this class of column (#112).
+- **Snowflake integer and NUMBER columns now read as numeric everywhere the
+  engine reasons about type.** Snowflake's `SHOW COLUMNS` surfaces every
+  integer/NUMBER as the token `FIXED`, which matched none of the numeric type
+  hints, so on Snowflake no integer column was recognized as numeric. `FIXED` is
+  now a numeric hint. Beyond making the type-aware PII gate above effective on
+  Snowflake, this is a visible pre-existing correction: Snowflake numeric columns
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
Yeah I have been noticing this issue quite heavily.

Looks good! Thanks @spascarella

PR #130, opened 2026-07-23
SCRUM-950-Fix BigQuery profile estimate omitting the per-query billing floor

### LINKED ISSUE #107: Multi-table `explore profile` cost estimate omits the per-query billing floor (actual spend 1.7x the estimate on small tables)
## Summary

`explore profile`'s two-step handshake estimated 231,958,108 bytes for an
8-table batch; actual billed spend was 400,556,032 bytes (+73%). The gap is
BigQuery's 10,485,760-byte minimum billed per query: profiling issues many
small per-column/per-check queries per table, and on small tables each
bills the floor rather than its dry-run bytes. The dry-run-sum estimate is
accurate for large tables and systematically low for many-small-tables
batches -- the common case in early exploration.

## Repro

1. Run `explore profile` over several small tables (each well under 10 MB
   of relevant column data). Note `cost.estimate` in the
   `needs_confirmation` envelope.
2. Confirm with an ample budget; compare `data.spend.bytes_billed`.

Observed: estimate 231,958,108; billed 400,556,032 across 8 tables (7 of
them 0.1--3.7 MB, one ~171 MB). Individually-billed ad-hoc probes in the
same session showed the same floor: every small probe bills exactly
10,485,760 regardless of dry-run size.

## Expected

The estimator knows (or can know) how many queries a profile will issue
per table, so the estimate can be
`sum(max(dry_run_bytes_i, minimum_billed))` instead of
`sum(dry_run_bytes_i)` -- near-exact rather than systematically low, on
the connector's own documented billing rule.

## Impact

Enforcement was never at risk (the budget capped correctly); this is about
estimate trustworthiness. For a tool whose central pitch is cost
governance, "actual = 1.7x estimate" on the flagship scanning command
teaches users to mentally pad the numbers, which erodes the value of
surfacing them at all. The same floor-awareness would also make ad-hoc
`explore query` estimates exact for small probes.

## Suggested fix

On connectors with a per-query billing minimum (BigQuery on-demand), apply
`max(estimate, floor)` per planned query when building the preflight
estimate, and mention the floor in the envelope's hint (e.g. "estimate
includes N queries at the 10 MB minimum").

---

**Environment**
- Source: a Firestore CDC export landed in BigQuery (~100 objects, mostly
  small tables)

emnarahmouni-exmergo:
https://github.com/exmergo/dex/issues/107

- `explore profile`'s BigQuery cost estimate floored each aggregate batch to
  the 10 MB per-query minimum, but never accounted for the two escalation
  queries a profile can still issue per table (exact-distinct-count for a
  near-unique column, composite-key probe) -- each of which also bills at
  least that floor. Invisible to the upfront estimate, this made actual
  spend come in up to 1.7x over the quoted number on batches of small tables.
- `profile_estimate()` now reserves one floor per table for each possible
  escalation query, skipped only when a table is provably empty at estimate
  time (an unknown row count, e.g. a view, still reserves, since escalation
  isn't ruled out). The estimate is now a ceiling actual spend won't exceed.
- The escalation queries' own internal budget charging is now floored the
  same way (previously unfloored, inconsistent with the rest of the gate).
- Added `describe_estimate()` so BigQuery's confirmation hint names the
  floor directly.
*Before the fix* : 
image
*After the fix* : 
image

Screenshot 2026-07-23 121925


marcociav-exmergo (approved):
LGTM

PR #131, opened 2026-07-23
SCRUM-953-Gate BigQuery dev_dataset warning on whether any node targets it

### LINKED ISSUE #110: `transform build`'s "dev_dataset does not exist" warning fires unconditionally, even when no node targets the fallback dataset
## Summary

Every `transform build` against BigQuery prints the "dev_dataset
`<project>.<name>` does not exist; dbt will create it on the first build
..." warning, even in a project where every model resolves to an explicit
per-layer schema via a `generate_schema_name` convention and no node in
the build graph ever touches the fallback dataset the warning is about.

## Observed

A project with per-layer `+schema:` config (staging/intermediate/marts
datasets per target) has built successfully many times; the named
`dev_dataset` has never been created and never will be, yet the warning
prints on every single run.

First noted as an aside in a comment on #61 (2026-07-16); filing it
separately now since it survived a release cycle and comment-asides are
easy to lose.

## Expected

Emit the warning only when at least one node in the resolved build graph
would materialize into the fallback dataset. dbt's manifest knows every
node's resolved schema before execution, so the condition is checkable for
free.

## Impact

A warning that fires on every run regardless of relevance trains users to
skim past that line -- which matters, because in a real
missing-and-unwritable scenario this same warning is the one that would
have explained the failure.

## Suggested fix

Gate the warning on `any(node.schema == dev_dataset for node in
manifest)` (or the adapter equivalent) instead of emitting it
unconditionally from config.

---

**Environment**
- dex plugin `1.2.1`, BigQuery connector
- dbt `1.11.12`, `dbt-bigquery`, per-layer `generate_schema_name` override

emnarahmouni-exmergo:
https://github.com/exmergo/dex/issues/110
- `transform build`'s BigQuery "dev_dataset does not exist" warning fired
  unconditionally on every build, regardless of whether the project's
  compiled graph would ever actually write into that fallback dataset. A
  project using per-layer `+schema:` config (or a `generate_schema_name`
  convention) resolves every model into its own schema and never touches
  `dev_dataset` at all -- the warning was pure noise there, training users
  to skim past a line that, in a genuine missing-and-unwritable scenario,
  is the one that would explain the failure.
- `_bigquery_namespace()` now consults the compiled manifest
  (`target/manifest.json`) via a new `_manifest_schemas()` helper: it
  collects every schema a materializing node (`model`/`seed`/`snapshot`,
  excluding `ephemeral` models, which never issue their own `CREATE`)
  resolves into, and the warning is suppressed when the missing dataset
  isn't among them.
- With no manifest on disk yet (a project's first build), there's nothing
  to rule the dataset out with, so the warning stays unconditional, same
  as before -- this only removes noise once there's compiled evidence to
  act on.
- Reading is defensive by design: a missing or corrupt manifest degrades
  to `None` (never raises), since this is a best-effort preflight signal,
  not a hard requirement.
- Scoped to BigQuery only, matching the reported issue. The same gap likely
  exists in the other connectors' analogous namespace checks (Snowflake/
  Databricks database-catalog refusals, Postgres/Redshift schema-privilege
  checks), but that's unreported and left for a follow-up.
*Before the fix* : 
image

*After the fix* : 
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`transform build`'s BigQuery "dev_dataset does not exist" warning no
+  longer fires when nothing in the project would ever write there**
+  ([#110]). A project with per-layer `+schema:` config (or an equivalent
+  `generate_schema_name` convention) resolves every model into its own
+  schema and never touches the connector-level `dev_dataset` fallback at
+  all, yet the warning printed unconditionally on every build regardless,
+  training users to skim past a line that, in a real missing-and-unwritable
+  scenario, is the one that would explain the failure. A compiled manifest
+  from a prior build already answers "does anything resolve into this
+  dataset" for free; the warning now consults it and stays silent when
+  nothing does, falling back to the previous unconditional warning only
+  when no manifest exists yet (a project's first build).
+
 ## [1.3.0] - 2026-07-21
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,18 +11,24 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
-- **`transform build`'s BigQuery "dev_dataset does not exist" warning no
-  longer fires when nothing in the project would ever write there**
-  ([#110]). A project with per-layer `+schema:` config (or an equivalent
-  `generate_schema_name` convention) resolves every model into its own
-  schema and never touches the connector-level `dev_dataset` fallback at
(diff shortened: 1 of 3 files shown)


emnarahmouni-exmergo:
About the last update: 
Extended this from BigQuery to every connector with a dev-namespace preflight (Snowflake, Databricks, Postgres, Redshift) same root cause, same fix.

None of the other connectors consulted the compiled manifest either, and it's worse there: Snowflake's and Databricks's missing-database/catalog check **refuses the build outright**, not just warns.

Same mechanism everywhere: a compiled manifest (when one exists) proves whether any node resolves into the namespace in question; no manifest yet falls back to the original unconditional behavior.

6 new tests, one per connector.

marcociav-exmergo (approved):
LGTM

PR #132, opened 2026-07-23
SCRUM-957-Carry forward out-of-scope datasets and relationships in explore map

### LINKED ISSUE #111: Scoped `explore map` replaces the cache, dropping out-of-scope dataset profiles instead of carrying them forward

## Summary

`explore map --scope <dataset>` rewrites `.dex/cache.json` with **only the scoped datasets** when the scope set differs from what produced the existing cache. Out-of-scope dataset profiles are silently dropped (`carried_forward_count: 0`, no warning), even though the skill text promises "objects skipped this run keep their prior profiles (`carried_forward_count`)".

## Repro

1. Start from a cache holding profiles across several datasets (ours: ~88 objects across 7 datasets — three environments' raw data plus a marts dataset; the cache is committed to git like a lockfile).
2. Run `explore map --scope <marts dataset>` → envelope returns OK with `carried_forward_count: 0`.
3. `jq '[.datasets[].identifier]' .dex/cache.json` → only the 20 in-scope objects remain. 87 previously profiled objects from the other six datasets are gone.
4. A subsequent `explore query` against one of the dropped (raw) tables is refused with "not in the .dex cache; run `explore map` ... first" — for a table profiled and committed days earlier.

## What does work

Re-running map with the **same** scope set carries forward correctly (later in the same session: `cache_hit_count: 5, carried_forward_count: 20`). The failure is specifically across *different* scope sets.

## Impact

- Profiles cost real money to produce; on a metered connector the loss is a re-profiling bill (we recovered the working-set datasets via `git checkout -- .dex/cache.json` + one multi-scope re-map, and simply ate the loss of the other environments' profiles).
- Mid-session probe refusals for tables the user knows are profiled.
- Since #96 (closed) landed cache-reuse/carry-forward machinery, this looks like an edge of that work: merge/carry-forward keyed to the current selection rather than the whole cache.

## Expected

Scoped map merges its results into the existing cache; out-of-scope datasets are carried forward with their per-dataset `profiled_at` stamps (the documented staleness mechanism). At minimum: a loud warning plus a `dropped_count` in the envelope before discarding anything.

emnarahmouni-exmergo:
https://github.com/exmergo/dex/issues/111

- `explore map --scope <dataset>` rebuilt `.dex/cache.json` purely fromthis run's inventory (`metas`), which `--scope`/`--dataset` narrows at the adapter level. `_compose_datasets()` only ever iterated `metas`, so a prior-cache dataset absent from it entirely not stale, just never seen this run -- had no code path to reach the output and silently vanished, with `carried_forward_count: 0` and no warning. A subsequent `explore query` against the dropped table then refused with "not in the .dex cache", for a table profiled and committed days earlier.
- `explore relationships` had the identical defect for `relationships` specifically: its own docstring assumed "inventories and profiles the full set, so its relationship set is authoritative" -- an assumption `--scope` breaks. Both commands replaced `relationships` wholesale from this run's inference alone.
- `_compose_datasets()` now also unions in any prior dataset absent from this run's inventory, unchanged (original `rank_score`, `profiled_at`, columns intact) -- returns a new `out_of_scope` count alongside the existing `carried` count.
- New `_carry_forward_relationships()` helper (shared by both `cmd_map` and `cmd_relationships`) unions back any prior relationship with an endpoint outside what this run actually examined (profiled or fresh-cache-reused), deduped against the freshly-built set by the same edge key `_merge_relationships` already uses.
- New envelope fields (`out_of_scope_carried_count`, `carried_relationship_count`) and explanatory notes make the carry-forward visible rather than silent.
- Re-running with the *same* scope as before was already correct; only a scope *change* between runs triggered the original loss.

*Before fix* : 

Screenshot 2026-07-23 140432

*After fix* : 
Screenshot 2026-07-23 140453
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,25 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A scoped `explore map` no longer drops out-of-scope datasets from the
+  cache** ([#111]). `explore map --scope <dataset>` (or `--dataset`) rebuilt
+  `.dex/cache.json` from only this run's inventory, so a scope narrower than
+  what built the existing cache silently discarded every other dataset's
+  profile, with `carried_forward_count: 0` and no warning: a subsequent
+  `explore query` against a dropped-but-still-real table then refused with
+  "not in the .dex cache". A prior dataset entirely absent from this run's
+  inventory is now carried forward untouched, and its own count
+  (`out_of_scope_carried_count`) and a note say so explicitly. `explore
+  relationships` and `explore map` also both used to replace the cache's
+  relationships wholesale from this run's inference alone; a relationship
+  with an endpoint outside what this run examined is now carried forward the
+  same way (`carried_relationship_count`), since neither declared-resolution
+  nor inference had visibility into that endpoint to regenerate or supersede
+  it. Re-running with the same scope as before was already unaffected by
+  this; only a scope *change* between runs triggered the loss.
+
 ## [1.3.0] - 2026-07-21
 
 ### Added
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
LGTM!

PR #146, opened 2026-07-26
SCRUM-954-Classify orphaned warehouse relations in maintain check

### LINKED ISSUE #113: maintain: classify warehouse relations with no backing dbt model as `orphan_relation` findings (advisory DROP proposals; down-rank in explore)
**Version:** dex 1.2.2, BigQuery connector. Detection-side companion to #94.

## Summary

Warehouse relations that no longer have a backing dbt model (residue of model renames/removals — dbt never drops) are invisible **as a class** to dex. In one session, two stale tables (dims that had been reclassified to facts and renamed weeks earlier) were touched by three dex surfaces, and none said the one actionable thing:

- `explore map` ranked one of the stale tables the **4th most interesting object** in the dataset — actively misleading for anyone browsing the map.
- `maintain check` reported both only as generic low-severity `table_added` findings against the (stale) snapshot. The telltale was there — `impacted_models: []` — but no classification, so the signal drowned among 47 other `table_added` findings.
- `transform` correctly refuses to touch the warehouse, but there is no advisory surface proposing the cleanup for human execution either.

## Proposal

1. **`maintain check`: an `orphan_relation` finding class** — "relation exists in the warehouse dataset a dbt layer builds into, but no model/source resolves to it" — with an advisory-kind reconcile proposal containing the `DROP TABLE` statement *for the human to run* (consistent with the read-only guarantee: propose, don't impose).
2. **`explore` ranking: down-rank or badge orphan relations** when a dbt project is present (this pairs naturally with the existing `--use-project` fold-in, which the map notes already nag about).

## Relation to #94

#94 covers the authoring side: plan/apply can't express a delete/rename, so renames escape the reviewable-diff guarantee. This issue is what the warehouse looks like weeks *after* that gap — even with #94 fixed, drift can still produce orphans (manual DDL, abandoned experiments), so the detection class is worth having independently.

emnarahmouni-exmergo:
Closes: https://github.com/exmergo/dex/issues/113

Adds orphan_relation detection for warehouse relations left behind by a model rename/removal dbt never drops them, so today they're invisible or misreported.

maintain check: new orphan_relation finding fires when a relation was model/source-backed at the last snapshot but isn't anymore, while still present in the warehouse
maintain reconcile: advisory proposal with a DROP TABLE statement for a human to run (dex never executes it)
explore map --use-project: down-ranks and badges orphan relations so they don't surface as top objects
Fixes a pre-existing Windows path-separator bug in dbt_project.py that silently broke this (and prior impacted_models tracing)
Scope: detection + advisory only, two-state classification (no snapshot history) per the issue's own suggested split; execution (macro + dbt run-operation) deferred to a follow-up issue

the issue : 
image
After fix : 
image

### DIFF
--- a/packages/dex-core/src/exmergo_dex_core/dbt_project.py
+++ b/packages/dex-core/src/exmergo_dex_core/dbt_project.py
@@ -216,7 +216,10 @@ def load(project_dir: Path | str = ".") -> DbtProjectView:
         for path in sorted(base.rglob("*")):
             if path.suffix not in {".sql", ".yml", ".yaml"} or not path.is_file():
                 continue
-            rel = str(path.relative_to(root))
+            # Posix-separated regardless of OS: every consumer of `files` keys
+            # (transform_layer's model-name parsing, scaffolded model paths,
+            # this module's own backed_relation_names) assumes "/".
+            rel = path.relative_to(root).as_posix()
             content = path.read_text(encoding="utf-8")
             files[rel] = SourceFile(
                 path=rel, content=content, sha256=content_hash(content)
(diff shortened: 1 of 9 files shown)


marcociav-exmergo (approved):
LGTM!

PR #153, opened 2026-07-26
docs are clarify explore query cells are row-major not columnar

### LINKED ISSUE #152: `explore query` returns `cells` row-major while the skill documents results as "columnar"
**Version:** dex 1.4.0, BigQuery connector.

## Summary

The `explore` skill states that `query` results "come back columnar and capped". The envelope key is `cells`, and it is a **list of rows**. Consuming it as documented transposes the result.

## Repro

Any `explore query` with more than one column:

```
explore query "select a, b, count(*) as n from <t> group by 1, 2"
```

`data.cells[0]` is the first **row** (`[a0, b0, n0]`), not the first column. The documented contract implies `data.cells[0]` is every value of `a`.

## Why this is worth correcting despite being cosmetic

The failure is not loud. With a query whose leading column is low-cardinality, a transposed render still produces a plausible-looking table — aligned header, sensible-looking values — and it surfaced here only as an `IndexError` on a ragged final row. A consumer with a square-ish result set would have silently misread the data and never known.

For a tool whose value proposition is not putting wrong numbers in front of people, a documented-contract mismatch on the one command that returns arbitrary values seems worth tightening.

## Expected

Either the skill text says row-major, or the key is renamed `rows`. `rows` is unambiguous and matches the actual shape.

## Non-issue, noted so it is not inflated

The cost of the mistake was two repeat probes, both of which **billed zero bytes** — BigQuery's own query cache absorbed the identical SQL. No dex cost defect involved.

Felix-sof:
## What this changes

The explore query skill and command contract said results come back "columnar," but the cells envelope key is actually a list of rows. Consuming it as documented transposes the result, which fails silently on a square-ish result set (only surfaces as an IndexError on a ragged final row, as reported in #152).

Fixed the wording in four docs (skills/explore/SKILL.md, references/command-contract.md, references/semantic-layer.md, references/evaluation.md) to say "row-major" instead, matching what the code actually does. Also fixed a matching docstring in commands.py. No behavior change, no key rename — went with the docs fix over renaming cells to rows since the key is used across many tests and connector adapters and a rename felt like a bigger, riskier change for what's fundamentally a wording bug.

## Related issues

Closes #152

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`explore query` docs now say row-major, matching what `cells` actually
+  returns** ([#152]). The `explore` skill and command contract described
+  `query` results as columnar, but `cells` is a list of rows; a square-ish
+  result set could be silently misread with no error. Docs updated to say
+  row-major across `explore query` and `explore semantic query`; no behavior
+  change.
+
 ## [1.4.1] - 2026-07-26
 
 ### Added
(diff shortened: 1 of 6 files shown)


marcociav-exmergo (approved):
LGTM! Thanks @Felix-sof!
I'm going to merge this

Issue #139, opened 2026-07-24
[core] Opt-in SQLite cache backend


marcociav-exmergo:
**Depends on #137.**

## Problem
The filesystem store writes a directory of loose JSON/JSONL files under `.dex/`. That is the right default (the files are git-committed and PR-reviewable, which is the "persistence is git" guarantee, see `cache.py:11`), but it does not suit every consumer: a library user or a long-running local session wants durable state without scattering JSON files, and the hosted backend (#140) needs real database semantics.

## Goal
Add a `SqliteStore` implementing the storage `Protocol` from #137, writing a single `.dex/dex.db`. **Opt-in only; filesystem-JSON remains the default.**

## Scope
- `SqliteStore` covering the full `Protocol` (cache, snapshot, drift, plans, and the two ledgers).
- Backend selection via a `cache.backend: filesystem | sqlite` field on `DexConfig` (`config.py:370`), default `filesystem`, mirroring how `connector:` is selected. Optionally a `--cache-backend` CLI override.
- The append-only ledgers become real tables; `spend_since` (`cache.py:285`), currently a full-file scan and filtered sum, becomes an indexed query. This also de-risks the Postgres work in #140.
- Its own schema-migration story (the JSON path uses `CACHE_SCHEMA_VERSION`, `cache.py:29`; SQLite needs an equivalent).

## CLI impact
One new selector (`cache.backend` in config, optional `--cache-backend` flag). Default is unchanged, so existing repos keep loose JSON. Document that `.dex/dex.db` is a binary blob and is **not** PR-reviewable the way the JSON cache is, and add it to `.gitignore` guidance (unlike the JSON cache, a `.db` should generally not be committed).

## Acceptance
- `cache.backend: sqlite` routes all state to `.dex/dex.db`; the default still writes JSON.
- Behavioral parity: the CLI test suite passes against both backends.
- `spend_since` and the session-budget guard work against the SQLite ledger.

marcociav-exmergo:
## Decision from #144: backend selection is an open registry, not a closed enum

#144 made the storage seam a public extension point, so the question of how a
backend gets *selected* had to be settled there even though this issue is where
the first alternate backend actually arrives. Recording the answer here because
this issue owns the field.

**The scope above proposes `cache.backend: filesystem | sqlite` on `DexConfig`.
Do not land it as a closed set.** A backend published as its own package would be
library-only forever, reachable through `DexEngine(store=...)` and never from the
[...]

PR #155, opened 2026-07-27
Make the storage seam a public extension point, and give every refusal a name

### LINKED ISSUE #144: [core] Make the storage seam usable by a host that ships its own backend
**Depends on #137. Sibling of #139 and #140.**

## Problem

#137 introduced the storage `Protocol` with `FilesystemStore` and `MemoryStore`, #138 made the config injectable, and #142 made the credential injectable. Together those are most of what an embedding host needs in order to explore and profile a warehouse with nothing on the filesystem, per end user.

What is missing is not another backend dex ships. It is that the seam is not usable by a third party without reading the engine's source and guessing at several contracts that exist only implicitly. The driver is a host that federates `.dex/` state per user in its own datastore and profiles each user's warehouse under that user's own credential. The credential and the configuration are covered. The store is where it gets rough.

Worth being precise about why the store is a hard requirement in that shape and not a performance choice: the exploration cache is not incidental state. The query firewall resolves every table and column reference against it, and cache membership decides what a query may name and under whose PII policy. So a `map()` in one request must still be visible to a query in the next. `MemoryStore` gives "nothing on disk" but not "durable across requests", and those are different requirements.

## Scope

- **Separate the explore-only surface from the transform-only one.** Five of the sixteen protocol members (`save_plan`, `load_plan`, `list_plans`, `latest_plan`, `plan_locator`) are transform-only, and nothing in explore or maintain calls them. A host that only explores must either implement five dead methods or ship a partial implementation that satisfies no declared contract. It works today only because `Store` is never `isinstance`-checked at runtime, which is accident rather than design.

- **Document the concurrency contract.** `spend_since` followed by `append_spend_log` is a read-then-write with no stated atomicity requirement, and `FilesystemStore` races it. #140 already names the requirement for the backend it will ship, but a third-party implementer has nothing written telling them whether the cumulative session ceiling depends on their making it atomic. State it on the protocol, where an implementer will read it.

- **Decide what `save_cache` owes a backend with a document size limit.** Both shipped backends write the whole cache as one blob. A document store typically caps a single document, and a profiled cache over a large warehouse can exceed that, so such a backend has to shard with no seam to shard along. Either add a per-dataset or chunked write, or state plainly that whole-document write is the contract and sharding is the backend's problem. This is a question #139 and #140 both have to answer anyway.

- **Type the refusals a programmatic host can hit.** #138 gave the cost refusals real exception classes, but the connection, project, and config refusals still surface as bare `ValueError`, `DbtProjectError`, and `CredentialDiscoveryError` from varying depths. The CLI hides this behind a catch-all; a library consumer cannot. An embedding host needs a documented set it can map onto its own responses, rather than catching `Exception` at its API boundary.

- **State which explore calls are filesystem-free.** `repo_root` is still needed by `--use-project` and by everything in transform, and the boundary is currently discoverable only by reading the code. A host needs to know which calls are safe with no project on disk.

- Consider whether `FilesystemStore.spend_since` should read backwards to the cutoff instead of parsing the whole ledger on every gate construction. Carried over from the #138 review, cheap to fix here, and it is the read a long-lived host performs most often.

## CLI impact

None. This is protocol shape, documented contracts, and typed refusals. The shipped filesystem backend, the command surface, and every envelope field stay as they are.

## Non-goals

- dex does not ship a document-store backend. A host's datastore is the host's to implement, and the protocol being public is the point. #139 and #140 own the backends dex ships.
- Per-tenant dbt projects stay out of scope, as #142 established. The dbt project is the source of truth and remains a git-reviewable filesystem artifact, so a host in this shape gets the read-only explore surface, not transform.
- No change to what the guards do. This issue makes the seam legible; it does not relax anything.

## Acceptance

- A third-party-shaped `Store` implementation in the test tree, neither filesystem nor memory, drives a multi-step explore flow (map, then a firewall-checked query) across **separate engine instances**. That is the shape a request-per-engine host actually has, and no current test exercises it: today's in-memory coverage runs both steps through one engine.
- Two such stores keyed differently share nothing, asserted for the cache, both ledgers, and the firewall's view of what a query may name.
- An explore-only implementation, with the transform members absent, satisfies the declared contract.
- The documented exception set covers every refusal reachable from the public API without a `repo_root`.

## Why now

#142 closed the identity axis, so identity and configuration are both per-request today while state is the axis still shaped for one repo on one disk. Writing these contracts down before `PostgresStore` (#140) is cheaper than discovering them twice, and #140 needs the same answers about atomicity and document size that a third-party backend does.

marcociav-exmergo:
Closes #144.

## Summary

`.dex/` state has lived behind a `Store` protocol since the storage seam landed,
and configuration and credentials became injectable after it. On paper that was
already most of what an embedding host needs. In practice the seam was only usable
by someone willing to read the engine's source and infer several contracts that
existed nowhere but its tests, which is a different thing from a public extension
point.

This makes it implementable from what is published. A backend can now live in
someone else's package, prove itself against a suite it installs, and know the
rules from written contracts rather than from reading our code.

Four changes carry that, plus a refusal taxonomy that turned out to be the same
problem one layer up.

**The protocol is tiered.** `ExploreStore` is six members and covers a host that
explores, profiles, and queries. `MaintainStore` adds the reconcile baseline and
the drift report. `Store` adds the five transform plan members and is unchanged in
shape, so existing annotations and backends keep working. The split is not a
judgment call; it falls out of the call sites exactly. A host implementing the
narrow tier is now writing a complete implementation of a declared contract rather
than a partial one, and the transform surface, the single place the widest tier is
required, refuses an explore-only store by naming the tier and the missing members
instead of failing on a missing attribute several frames down.

**The conformance contract ships.** `exmergo_dex_core.storage.conformance` holds
what `tests/storage/test_parity.py` used to hold alone. A backend installs
`[storage-conformance]`, subclasses the class matching its tier, supplies a
`make_store(key)` factory, and runs every assertion we hold our own backends to,
isolation across keys included. The in-repo file is now the same three lines a
third party writes, so exactly one copy of the contract exists.

**Refusals have a taxonomy a host can branch on.** Everything the engine refuses
on purpose descends from `DexError`. Below it, `ConfigurationError` and
`RequestError` (both also `ValueError`, so nothing breaks for existing callers),
`PrerequisiteError` for state that a named command creates, and `ConnectorError`
for a warehouse that could not be reached. Eleven refusals that were typed but
not importable are now exported, alongside the new bases.

**`py.typed` ships.** The seam is enforced entirely by structural typing, so
without the marker a downstream type checker treated the package as untyped and
verified nothing about a candidate backend.

Plus `references/storage.md`, the seven behavioral contracts written onto the
protocol members they govern, and pointers from `AGENTS.md` and `CONTRIBUTING.md`.

43 files changed, 2,242 insertions, 402 deletions.

## Decisions worth a reviewer's attention

**`save_cache` is a whole-document, atomic write, and there is deliberately no
per-dataset seam.** The query firewall resolves every table and column reference
against the cache, so cache membership decides what a query may name and under
whose PII policy. A half-written cache is a security-relevant state, not merely an
inconsistent one. A backend with a document size cap chunks internally and
presents one logical document; the seam that would let it shard is the same seam
that would let a reader observe half a cache.

**The ledger read-then-write is not atomic at the protocol level, and the bound is
stated rather than fixed.** The session ceiling binds exactly under serialized
commands, which is the CLI's one-command-per-process shape. Under genuinely
concurrent commands the overshoot is bounded by the sum of the concurrent
estimates. A backend that can make the pair atomic for one key should. The
protocol does not grow a member for it, because an optional member no implementer
can rely on is worse than a stated bound.

**The ledger is scoped to the store instance**, so store granularity is ceiling
granularity. One principal spanning two stores has two independent ceilings and
nothing bounds their sum. It errs permissive and nothing warns, so this is stated
on the protocol and in the reference page: key stores exactly as you key
principals.

**`ConfigurationError` and `RequestError` also inherit `ValueError`.** Those
refusals raised bare `ValueError` before they were typed, and a consumer written
against that API keeps catching what it caught. The retained base has a docstring
saying so, otherwise a later reader deletes it as redundant.

**`CredentialDiscoveryError` is deliberately not a `ConnectorError`.** A host backs
off and retries on connector failures. A credential that was never configured will
not appear on a retry, and its message names the config that fixes it, so filing
it under the retry family would send hosts into a loop that cannot terminate.
There is a test pinning this.

**The two missing-extra errors were not re-parented under `ConfigurationError`,**
even though it would have made a tidier tree. Neither is a `ValueError` today, and
`ConfigurationError` carries `ValueError` for compatibility, so re-parenting would
newly make `except ValueError` swallow a missing-extra error somewhere it does not
today. Not worth a silent behavior change for tidiness.

**Backend selection is decided but not built.** An open registry, not a closed
enum: `cache.backend` should accept the shipped names plus a dotted
`module:ClassName` path or an entry-point group. Recorded in `references/storage.md`
and on #139, which is where the first alternate backend arrives and where the
field belongs. Shipping a resolver with no caller here would have been dead code.

## Verification

1472 tests pass, ruff and format clean, em-dash check clean.

The end-to-end proof is a packaging test, because nothing else answers the actual
question. It builds the wheel, installs it into an environment with no access to
this source tree, writes a backend there against the published protocol alone, and
runs the shipped conformance suite at it. If that passes, an outside contributor
can do the same. It also caught a real bug while being written: the conformance
module imported the plan model at top level, which reaches the SQL guard and so
the dialect engine, meaning an explore-only implementer on a light install could
not import the contract at all. Those imports are lazy now, and only the tier that
stores plans pays for them.

Live, because the exception tree moved under the connectors and the storage seam
is what the cost gate writes through:

- BigQuery 11 passed, Snowflake 14 passed, Databricks 9 passed 3 skipped
- Hosted dbt Cloud Semantic Layer: catalog read and metric query through an
  injected credential with nothing ambient left to discover, PII dimension refused
  before any query was submitted, a bogus token rejected without leaking it
- A six-method explore-only backend, written against the published protocol,
  driving a live billed flow across **two separate engines**: the first profiled a
  real table into that backend, the cost gate wrote its ledger entries through it,
  the second engine ran a firewalled query resolving against the first engine's
  cache, and a second tenant key saw none of it

One note on that last run: `billed_bytes` came back zero, which looks like broken
plumbing until you read the entries. BigQuery served the identical statements from
its own result cache. The ledger entries are real and the bytes are legitimately
zero, so the assertion checks entries rather than totals; asserting on spend would
be flaky by design.

## On the field report

@catincloudlabs, thank you for this. It was specific enough to act on directly,
and all three points are in this PR rather than waiting as follow-ups. Taking them
in order.

**§1, the two unexported refusals.** You were right, and the split was not
deliberate. `CacheRequiredError` and `NoBaselineError` are exported. We went wider
than the two, because the same wall was waiting for you at
`CredentialDiscoveryError` and the next person at `ScopeError`: eleven refusals
that were typed but unreachable are now importable from the package root.

Your 409-versus-502 case is the reason the taxonomy has the shape it does rather
than being a flat list of exports. `PrerequisiteError` is exactly the class of
refusal your worker can resolve on its own, so the branch is now a type check:

```python
def status_for(exc):
    if isinstance(exc, PrerequisiteError): return 409   # run the command, retry
    if isinstance(exc, ConnectorError):    return 502   # upstream, back off
    if isinstance(exc, DexError):          return 422   # refused on purpose
    return 500
```

A test now asserts every refusal reachable from the public API is importable, with
an explicit internal set carrying a reason per entry. That omission is precisely
how this survived, so adding a refusal now forces a decision rather than allowing
a silent one.

Worth saying plainly: your diagnosis was better than the framing in the issue. The
scope item talked about *typing* refusals, and you pointed out these were already
typed and merely unreachable, which is a different defect with the same symptom.
The rewording in 1.4.0 was not the bug; prose being the only machine-readable part
of the error was.

**§2, the ledger's scope.** Also right, and your reading of why the docstring
misleads is the part we would not have found on our own. It was specific about
`field` and `connector` keeping paradigms apart, which reads as the complete set of
scoping rules, and the store axis was the one it omitted. It now says so on the
protocol and in the reference page, close to your wording, including that it errs
permissive and nothing reports it. Your two-ledger numbers made the case better
than an abstract argument would have.

**§3, the assertion messages.** All 59 now carry one (the suite grew from the 46
you measured). You were right that the docstring pass and the message pass are one
piece of work with two outputs, so each message restates the rule its protocol
member documents. And rather than assert that this helps, there is now a test that
runs a deliberately aliasing backend against the shipped contract and checks the
failure names the rule and the fix:

```
a stored document must not alias the caller's object: mutating what you saved
reached into the store, so a mutation after a save silently rewrites history.
Serialize on write, or deep-copy.
```

**One you did not name.** Cross-checking your report against our own work found a
fourth. `Store.load_plan` is documented as raising `PlanNotFoundError`, and that
name was not exported either, so a third-party full-tier backend could not satisfy
the contract we were about to publish without importing from a private module.
That is the same defect as §1, one layer over, and we would have shipped it. It is
fixed and there is a test asserting the protocol docstring and the public name
agree.

**On your offer.** All three items you proposed are consumed here, so there is
nothing left in them to hand back, and we would rather not have you duplicate
work. The useful thing, if you still want it, is the layer above: your §3 says the
conformance suite is the piece you would use first and pin against, and a `Store`
implementation built against it once this lands is worth more to us than any of
the three. It would be the first backend written by someone who did not write the
protocol, which is the only real test of whether this PR did its job. Fixture names
and assertions are public API from this point, as you asked.

If you do build one, the isolation assertions are the ones to read first if
anything fails. Two tenants leaking into each other is the failure the whole seam
exists to prevent.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dex",
   "description": "Analytics engineering for Claude Code and any agent: data warehouse exploration, dbt transformation and semantic modeling, and schema-drift maintenance on dbt.",
-  "version": "1.4.1",
+  "version": "1.4.2",
   "author": {
     "name": "Exmergo, Inc.",
     "email": "support@exmergo.com",
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,3 +175,4 @@ them.
 - Connector and methodology notes: `references/`.
 - The contract in full: `references/command-contract.md`.
 - The source of truth (dbt) and `.dex/` cache: `references/canonical-model.md`.
+- Where `.dex/` state lives, and writing a storage backend: `references/storage.md`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+## [1.4.2] - 2026-07-28
+
 ### Changed
 
 - **Every refusal dex raises deliberately is catchable as one type** ([#144]).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,112 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **The storage seam is a public extension point: a shipped conformance suite, a
+  tiered protocol, and a typed package** ([#144]). `.dex/` state has lived behind
+  a `Store` protocol since the storage seam landed, but the seam was only usable
+  by someone willing to read the engine's source and infer several contracts that
+  existed nowhere but its tests. It is now implementable from what is published.
+
+  The protocol is three nested tiers, so a backend implements what its host
+  actually uses instead of stubbing what it does not. `ExploreStore` is six
+  members (the exploration cache, the two ledgers, locators) and covers a host
+  that explores, profiles, and queries; `MaintainStore` adds the reconcile
+  baseline and the drift report; `Store` adds the five transform plan members and
+  is unchanged in shape, so existing annotations and backends keep working. The
+  transform surface is the one place the widest tier is required, and it now
+  refuses an explore-only store by naming the tier and the missing members rather
+  than failing on a missing attribute several frames down.
+
+  `exmergo_dex_core.storage.conformance` ships the contract as an executable
+  suite. A backend outside this distribution installs `[storage-conformance]`, subclasses
+  the class matching its tier, supplies a `make_store(key)` factory, and runs
+  every assertion dex holds its own backends to, isolation across keys included.
+  A packaging test proves this end to end: it builds the wheel, installs it in an
+  environment with no access to this source tree, writes a backend there against
+  the published protocol alone, and runs the shipped suite at it.
+
+  The package now ships a `py.typed` marker. The seam is enforced entirely by
+  structural typing, so without it a downstream type checker treated the whole
+  package as untyped and verified nothing about a candidate backend.
+
+  Six behavioral contracts that previously existed only inside the test tree are
+  written onto the protocol members they govern: documents do not alias the
+  caller's object in either direction, `save_cache` stamps the caller's object as
+  well as the stored copy, documents round-trip as pydantic models, a corrupt
+  document raises while a corrupt ledger line is skipped, a stored
+  `schema_version` is not the store's to police, and the ledger read-then-write is
+  not atomic at the protocol level (the session ceiling binds exactly under
+  serialized commands, and the overshoot under concurrency is bounded by the sum
+  of the concurrent estimates). `save_cache` is now explicitly a whole-document
+  atomic write: cache membership decides what a query may name, so a backend with
+  a document size cap chunks internally rather than getting a per-dataset seam
+  that would let a reader observe half a cache.
+
+  A seventh contract, reported from the field: **the spend ledger is scoped to the
+  store instance**, so store granularity is ceiling granularity. One principal
+  spanning two stores has two independent session ceilings and nothing bounds
+  their sum. The existing docstring was specific about `field` and `connector`
+  keeping paradigms apart, which read as the complete set of scoping rules, and
+  the store axis was the one it omitted. It errs permissive and nothing warns, so
+  a host federating state per user should key stores exactly as it keys
+  principals.
+
+  Every assertion in the shipped contract now carries a message naming the rule it
+  enforces. In-repo a bare comparison was fine because a failure sent you to the
+  source; as the artifact an outside implementer debugs against, the assertion
+  text is the documentation. The no-aliasing and clock-stamping rules matter most
+  there, since both fail silently and late and present as data corruption rather
+  than as a protocol violation.
+
+  New reference page `references/storage.md` covers the tiers, those contracts,
+  which calls need nothing on the filesystem, and how a backend is selected.
+
+### Changed
+
+- **Every refusal dex raises deliberately is catchable as one type** ([#144]).
+  The CLI renders refusals into an error envelope behind a catch-all, so it never
+  needed a hierarchy; a library consumer cannot do that, and catching `Exception`
+  at an API boundary swallows real bugs alongside deliberate refusals. All of
+  them now descend from `DexError`, with `ConfigurationError` for an engine built
+  without something it needs and `RequestError` for a call that named something
+  unusable. `NoConnectorSelectedError`, `RepoRootRequiredError`, and
+  `StoreRequiredError` replace the bare `ValueError`s that the public API raised
+  for the three refusals a host hits most.
+
+  **Nothing breaks for an existing consumer.** Both families also inherit
+  `ValueError`, which is what those refusals raised before they were typed, so
+  code written against the previous API keeps catching what it caught.
+
+  Two families name the distinctions a host actually branches on.
+  **`PrerequisiteError`** covers refusals where the engine is fine, the call is
+  fine, and some state does not exist yet that a named command creates:
+  `CacheRequiredError` and `NoBaselineError`. It is the one family a caller can
+  resolve automatically, by running the command the message names and retrying.
+  **`ConnectorError`** covers a warehouse connection that could not be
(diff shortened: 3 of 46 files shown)


catincloudlabs:
@marcociav-exmergo Glad the field report was useful, and thanks for taking all
three points into the PR directly.

I'll start working on the backend and will come back with findings once I have
them.

PR #158, opened 2026-07-28
SCRUM-956-Scope maintain check to its objects, and fix semantic cardinality pla…

### LINKED ISSUE #115: `maintain check` cannot be scoped: paid grain/cardinality estimate always covers every configured dataset

## Summary

`maintain check`'s paid axes (grain + dimension cardinality) produce **one combined estimate over every configured dataset**, with no way to narrow it. In a session focused on one environment's marts, the estimate (~315 MB) was dominated by ~30 raw tables across *other environments* that were irrelevant to the work — so the whole paid sweep was declined, and the layer under active change got no grain coverage from `maintain` at all (dbt uniqueness tests had to carry it).

`explore` already has the answer: repeatable `--scope`, free to resolve, only narrowing the configured allowlist. `maintain check` (and the focused `maintain grain`/`maintain semantic` detectors) should accept the same flag.

## Two adjacent nits from the same session (fold in or split, maintainer's choice)

1. **Batch unprofiled-table refusals in `explore query`.** A probe joining N tables is refused on the first unprofiled table only; each fix costs a round trip (+ profile spend). One refusal listing *all* unprofiled referenced tables would let a single profile call clear the probe.
2. **`transform apply` envelope should prominently report the written-file count and paths.** After applying a 30-file plan the OK envelope didn't surface an obvious count (git status was the verification path). Self-verifying apply output would remove one manual check from every apply.

emnarahmouni-exmergo:
Closes: https://github.com/exmergo/dex/issues/115

Adds object scoping to maintain check, and fixes a related bug in maintain semantic's paid scan found while implementing it.

maintain check <objects> now accepts the same scope as schema/volume/grain/semantic, narrows every axis, including the two paid ones (grain, cardinality), not just the free ones
Fixes cardinality_plan() estimating/billing for every semantic model regardless of scope; only the reported findings were filtered before, after the money was already spent
Scope vocabulary for the semantic axis matches identifier, column, dimension, or semantic model name, same as existing finding-level scoping

CHANGELOG.md updated under [Unreleased]

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **`maintain check` accepts the same object scope as its focused detectors**
+  ([#115]). The paid grain and cardinality axes always estimated and billed for
+  every configured dataset, with no way to narrow them; a session focused on one
+  environment's marts saw its estimate dominated by dozens of irrelevant raw
+  tables, so the whole paid sweep was declined and the layer under active change
+  got no grain coverage at all. `maintain check <objects>` now resolves the same
+  scope `maintain schema`/`volume`/`grain`/`semantic` already accept, narrowing
+  every axis, including the two paid ones, to what was actually asked for.
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


marcociav-exmergo (approved):
LGTM

PR #173, opened 2026-07-29
Select a storage backend from configuration, including one `dex` does not ship

### LINKED ISSUE #156: [core] Decide the construction contract for a store named in configuration
**Depends on #144. Blocks the storage-backend selector, and with it #139, #140, and #154.**

## Problem

#144 made the storage seam public: the `Store` protocol is structural and tiered,
the conformance suite ships in the wheel, the package carries `py.typed`, and
`references/storage.md` records that backend selection should be an open registry
rather than a closed enum, so a backend published as its own package is selectable
without a change to dex.

What #144 deliberately did not decide is what happens between "configuration names
a class" and "the engine holds a store". Resolving `mypkg.stores:MyStore` from a
string is fifteen lines and not the hard part. The hard part is what the resolver
calls it with, and the shapes dex already has disagree:

| Backend | Constructed with |
|---|---|
| `FilesystemStore` | a `repo_root` path |
| `MemoryStore` | nothing |
| `PostgresStore` (#140) | a `(tenant, project)` identifier, explicitly **not** a `repo_root` |
| a hosted document store (#154) | a tenant identifier plus service coordinates, with no repo at all |

There is no single call that satisfies those four, so the contract has to be
chosen rather than discovered.

## Why this cannot wait for the implementation

Because the obvious implementation gets it wrong in a way that is expensive to
reverse. The two backends that exist today are both path-shaped, and the next one
scheduled (#139) is a file as well. Design the contract with only those in view
and `resolve(name)(repo_root)` looks correct, at which point `repo_root` is baked
into the config schema. The first tenant-keyed backend then arrives and cannot be
constructed, and widening it afterwards is a schema change with a deprecation.

Deciding it now, against `FilesystemStore` on one side and a tenant-keyed store on
the other, forces the awkward shape to be answered up front. That is the whole
reason this is separate from the selector that consumes it.

## The question, concretely

Three candidate contracts, none of them obviously right:

1. **Always pass `repo_root`.** Simplest, matches both shipped backends, and rules
   out every backend that has no repo. That is the class of backend the public
   seam exists for, so this is almost certainly wrong, but it is worth stating so
   the rejection is on the record.
2. **Require a `from_config` classmethod on the store.** Flexible and explicit.
   The cost is that it adds an obligation the structural protocol deliberately
   does not have, and one the conformance suite does not check, so a backend can
   pass the whole suite and still be unconstructable. `references/storage.md`
   currently advertises "no base class to inherit and no registration step" as a
   feature, and this erodes it.
3. **Name a separate factory contract.** `Store` stays purely structural and
   construction lives in its own small protocol, so a dotted path may point at
   either a conforming class or a factory function. Keeps the two concerns apart
   at the cost of one more named thing to document.

My inclination is (3), but the argument belongs here rather than in a code review.

## Two constraints the answer has to satisfy

**Secrets never enter the configuration.** `.dex/config.yml` is committed, and the
config module is explicit that passwords, keys, and tokens are read at runtime
from their own stores. So a `cache.options` block, if the answer has one, carries
non-secret coordinates only, and the credential arrives the way `ConnectionSource`
and `SemanticSource` already deliver one: injected per request, or discovered from
the environment. This is easy to get wrong, because putting a connection string in
config is the obvious-looking design.

**Whatever is decided must not weaken the conformance suite's promise.** Today
"the suite is green" is the complete definition of a correct backend. If
construction becomes a second, unchecked obligation, that promise quietly stops
being true. Either the contract is checkable by the suite, or the documentation
says plainly that conformance covers behavior and not construction.

## Acceptance

- The contract is written in `references/storage.md`, next to the tiers, in enough
  detail that a third party can construct their backend from configuration
  without reading engine source.
- A backend with no `repo_root` is constructable under it. This is the case that
  fails if the decision defaults to the path-shaped backends.
- If the answer introduces a factory seam, it is named and documented alongside
  `Store`, and the relationship between the two is stated.
- #139, #140, #154, and the selector issue all reference this rather than
  re-answering it.

## Non-goals

- No selector, no `cache.backend` field, no CLI flag. Those belong to the selector
  issue, which consumes this decision.
- No new backend.
- No change to what any guard does. Where a store comes from does not change what
  it owes its callers.

### LINKED ISSUE #157: [core] Select a storage backend from configuration, including one dex does not ship
**Depends on #156. Unblocks #139, #140, and #154 being usable from the CLI.**

## Problem

#144 shipped everything a third party needs to *write* a storage backend: the
tiered `Store` protocol, the conformance suite in the wheel, `py.typed`, and the
written contracts in `references/storage.md`. It shipped nothing that lets anyone
*select* one.

A backend from outside is reachable through exactly one route today,
`DexEngine(store=...)` (`engine.py:156`). There is no path from the command line
at all: `DexEngine.from_repo` hardcodes `FilesystemStore(root)` at `engine.py:190`,
and `DexConfig` has no cache section to override it. So a contributor can
implement a backend, watch the shipped suite go green, and then discover the only
way to use it is to stop using the CLI.

That is the half-finished part of the seam, and it is what makes the difference
between a protocol that is publicly documented and an extension point that is
publicly usable.

## Direction

The decision is already recorded in `references/storage.md` and on #139: an
**open registry, not a closed enum**. A closed set of shipped names would make
out-of-tree backends library-only permanently, and opening it later is a
config-schema change with a deprecation attached.

- A `cache.backend` field on `DexConfig` (`SemanticConfig` is the precedent for a
  sub-block) accepting the names dex ships, plus either a dotted
  `module:ClassName` path or a name registered through an
  `exmergo_dex_core.stores` entry-point group.
- One resolver in the storage package, so the CLI, `from_repo`, and any embedding
  host select identically. Selection logic in two places is how they drift.
- `from_repo` consults it instead of hardcoding the filesystem backend. That is
  the single wiring change (`engine.py:190`); an explicitly passed `store=` still
  wins, as it does today.
- Optionally a `--cache-backend` override for an ad-hoc run, matching how
  `--connector` overrides the configured connector.
- Construction follows whatever #156 decides. Do not re-answer it here.

## Refusals

Every failure mode gets a typed refusal naming the fix, using the exception
taxonomy #144 introduced rather than a bare `ValueError`:

- an unknown shipped name, listing the ones that exist
- a dotted path that does not import, or imports to something that is not a store
- a class that does not satisfy the protocol, named as the tier it falls short of
  (the tiers are `isinstance`-checkable, so this can be precise)
- a backend selected for a command whose tier it does not implement, which already
  refuses through `require_full_store` and should keep the same message shape

## Worth deciding inside this issue

**Whether `memory` is selectable from the CLI at all.** It probably should not be.
CLI subcommands are separate processes, so a `MemoryStore` would silently drop the
cache between `explore map` and `explore query`, and the second command would
refuse with "run `explore map` first" having just run it. That reads as a broken
tool rather than a chosen backend. Either leave it library-only with a documented
reason, or refuse the selection with a message that explains the process boundary.

## Acceptance

- `cache.backend: filesystem` routes to the shipped backend and the default is
  unchanged, so existing repos behave exactly as they do now.
- A dotted path to a store defined outside this package resolves and drives a
  multi-step explore flow from the CLI. `tests/storage/test_conformance.py`
  already has a tenant-keyed store that is neither shipped backend and can serve
  as the target.
- An entry-point registration resolves the same way, tested against an installed
  distribution rather than a source tree. `tests/test_packaging.py` already builds
  a wheel and runs it in an isolated environment.
- Every refusal above is typed and names its fix.
- `references/storage.md` gains the concrete syntax, replacing the paragraph that
  currently records the direction.

## CLI impact

One new config field and an optional flag. The default is unchanged, so a repo
that selects nothing keeps writing loose JSON under `.dex/` exactly as before.

## Non-goals

- No new backend. #139, #140, and #154 own those.
- No change to the construction contract, which #156 owns.
- No relaxation of any guard. Where the cache lives does not change what the query
  firewall resolves against it, so a selected backend carries the same correctness
  bar as the filesystem one.

marcociav-exmergo:
The storage seam has been public since the last release: the `Store` protocol is
tiered and structural, a conformance suite ships inside the wheel, and the package
carries `py.typed`. A third party could write a backend and prove it correct. They
could not select it.

This closes that. `.dex/config.yml` gains a `cache` block, backend selection is an
open registry rather than a closed enum, and what a named backend is constructed
with is now a documented contract instead of an open question.

```yaml
cache:
  backend: mypkg.stores:my_store
  options:
    tenant: acme
```

### A host can keep dex's state wherever it already keeps state

The exploration cache, the reconcile baseline, the drift report, the two ledgers,
and the transform plans no longer have to be loose JSON files in someone's repo. A
process serving several end users can federate all of it per user, in its own
datastore, without forking the engine and without every call site changing.

That was already possible from Python, `DexEngine(store=...)`, and it stays the
shortest path for a host that builds its own store. What is new is that the choice
can live in configuration, so the CLI reaches it too, and a deployment can change
where state lands without changing code.

The source of truth does not move. The dbt project stays a git-reviewable
filesystem artifact, as it always has. Only the non-canonical scratch state is in
scope here.

### A backend can be published as its own package and selected by name

Three kinds of name resolve, in this order:

| Name | Example | For |
|---|---|---|
| shipped | `filesystem` | dex's own backends, never shadowable by anything installed |
| dotted path | `mypkg.stores:my_store` | a factory reachable by import, with no packaging work at all |
| entry point | `acme` | a name an installed distribution registered |

The entry point is what makes a published backend feel like a shipped one:

```toml
# in your own pyproject.toml, not in dex
[project.entry-points."exmergo_dex_core.stores"]
acme = "dex_acme_store:acme_store"
```

Install it beside dex and `backend: acme` resolves. No fork, no pull request here,
no place in dex where your name has to be listed. That is the point of an open
registry: a closed set of shipped names would have made out-of-tree backends
library-only permanently, and opening it later would have been a schema change with
a deprecation attached.

Writing one is unchanged and still small. Implement the tier you actually use
(`ExploreStore` is six methods), add a factory, and prove both with the shipped
suite:

```python
from exmergo_dex_core.storage import Store, StoreContext
from exmergo_dex_core.storage.conformance import StoreContract, StoreFactoryContract


class TestMyStore(StoreFactoryContract, StoreContract):
    tier = Store

    def build(self, context):
        return my_store(context)

    def context_for(self, key):
        return StoreContext(options={"tenant": key})
```

Composing `StoreFactoryContract` in front routes every behavioral and isolation
assertion through your own factory, so "the conformance suite is green" still means
your backend is correct **and** constructable, rather than quietly meaning only the
first.

### For the skills plugin

**Nothing in `explore`, `transform`, or `maintain` changes, and nothing needed to.**
The command surface is identical, the envelope shape is identical, and the default
backend is identical. The bootstrap in each skill's `run.py` reads only an
unindented `connector:` line out of `.dex/config.yml`, so the new `cache:` block is
invisible to it.

Two things do improve for an agent driving the engine:

**A failed engine build now comes back as an envelope.** Every wrapper reads exactly
one JSON object from stdout. The CLI used to construct the engine outside its error
handler, so anything raised while reading config or building the store escaped as a
traceback that no wrapper can parse. That was already reachable through a malformed
config file. It now renders as a normal error envelope with an actionable message,
and a test pins it.

**Refusals name the fix rather than the symptom.** An unknown backend lists what
exists and both open forms; a dotted path that will not import says so and points at
the environment; a factory that builds something which is not a store names the
members it lacks. These are the messages an agent surfaces to a human, so they are
written for that reader.

Worth stating plainly, because it is the obvious next question: a skill cannot yet
*use* a third-party backend. Each skill runs the engine through
`uv run --with exmergo-dex-core[<connector>]` in an isolated environment, so a
backend published as a separate distribution is not on that environment's path.
Making that reachable is a small additive change to the skill bootstrap, and it is
deliberately not in this pull request.

## Nothing changes if you do not opt in

`backend` defaults to `filesystem`. A repo that configures nothing writes the same
loose JSON under `.dex/` it always did, reviewable in a pull request exactly as
before, and a test asserts that default rather than assuming it. `DexEngine()` still
defaults to the in-process store, so importing the package still cannot leave a
`.dex/` directory in a consumer's repo. An explicitly passed `store=` still wins over
anything configured, because a caller holding an instance has already made the
decision configuration exists to make for the callers who have not.

Two deliberate refusals rather than silent behavior:

**`memory` is not selectable from the CLI.** Each command runs as its own process, so
an in-process store would drop the cache between `explore map` and `explore query`,
and the second command would refuse with "run `explore map` first" having just run
it. That reads as a broken tool rather than a chosen backend, so it refuses by
explaining the process boundary. It remains the default for a library caller, where
one process holds the engine.

**Credentials never belong in `cache.options`.** `.dex/config.yml` is committed, so a
password, key, token, or connection string there would be a secret in version
control. A backend needing one reads it at runtime the way the connection and
semantic-layer seams already do, or the host builds the store itself and passes it
in. This is the design that looks obvious and is wrong, so it is stated on the
config model, on the context, and in the reference doc.

Closes #156. Closes #157.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -175,4 +175,5 @@ them.
 - Connector and methodology notes: `references/`.
 - The contract in full: `references/command-contract.md`.
 - The source of truth (dbt) and `.dex/` cache: `references/canonical-model.md`.
-- Where `.dex/` state lives, and writing a storage backend: `references/storage.md`.
+- Where `.dex/` state lives, selecting a backend, and writing one:
+  `references/storage.md`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,108 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **A storage backend dex does not ship can be selected from configuration**
+  ([#157]). `.dex/config.yml` gains a `cache` block, and it is the schema change
+  worth reading first:
+
+  ```yaml
+  cache:
+    backend: mypkg.stores:my_store
+    options:
+      tenant: acme
+  ```
+
+  `backend` defaults to `filesystem`, so a repo that configures nothing behaves
+  exactly as it did before the setting existed, and `options` reaches the selected
+  backend's factory verbatim. `--cache-backend` overrides the name for one run,
+  the way `--connector` overrides the configured connector. **Credentials never
+  belong in `options`**: this file is committed, so a backend needing one reads it
+  at runtime the way `connect.py` does.
+
+  It is an **open registry, not a closed enum**, which is the part that would have
+  been expensive to get wrong: a closed set of shipped names would make out-of-tree
+  backends library-only permanently, and opening it later would be a config-schema
+  change with a deprecation attached. Three kinds of name resolve, in order: a
+  shipped name (`filesystem`), a dotted `mypkg.stores:my_store` path, or a name an
+  installed distribution registered under the `exmergo_dex_core.stores`
+  entry-point group. A shipped name always wins over a registration, so installing
+  a package can never silently move where an existing repo's state lands.
+
+  Before this, a contributor could implement a backend, watch the shipped
+  conformance suite go green, and then discover the only way to use it was to stop
+  using the CLI. `DexEngine.from_repo` now builds whatever the configuration names
+  instead of hardcoding the filesystem backend; an explicitly passed `store=` still
+  wins over both, since a caller holding an instance has already made the decision.
+
+  **`memory` is deliberately not selectable.** Each CLI command runs as its own
+  process, so a `MemoryStore` would drop the cache between `explore map` and
+  `explore query`, and the second command would refuse with "run `explore map`
+  first" having just run it. That reads as a broken tool rather than a chosen
+  backend, so it refuses by explaining the process boundary. It remains the default
+  for a library caller, where one process holds the engine.
+
+  Every failure refuses with a `ConfigurationError` naming the fix: an unknown name
+  lists what exists and both open forms, a dotted path that will not import points
+  at the environment, a name resolving to something uncallable says so, and a
+  factory that builds something which is not a store names the members it lacks.
+  The tier check is on the constructed store rather than on the factory, because a
+  callable protocol can only verify that `__call__` exists.
+
+- **A storage backend can now say how it is constructed, not just how it behaves**
+  ([#156]). #144 made the seam implementable: the tiered `Store` protocol, the
+  shipped conformance suite, `py.typed`. What it left undecided was what happens
+  between "something names a backend" and "the engine holds a store", and the
+  shapes disagree: `FilesystemStore` is built from a repo root, `MemoryStore` from
+  nothing, and a backend serving several end users from a tenant id with no
+  repository anywhere in the picture. There is no single call that satisfies all
+  three, so the contract had to be chosen.
+
+  Construction is now its own small contract, deliberately separate from `Store`.
+  A `StoreContext` carries a `repo_root` (`None` when there is no repository) and
+  an `options` mapping of the backend's own non-secret coordinates, passed through
+  verbatim; a `StoreFactory` is anything callable that turns one into a store. A
+  plain function, a class whose `__init__` takes the context, and a classmethod
+  all qualify, so a backend is constructable in whatever shape it already has.
+  `FilesystemStore.from_context` and `MemoryStore.from_context` are the shipped
+  reference implementations, one path-shaped and one that needs nothing.
+
+  **`Store` stays purely structural.** Putting a construction obligation on the
+  protocol would have cost the property that makes this seam cheap to implement,
+  that a class with the right methods is a store, with no base class to inherit
+  and no registration step. A host that builds its own store and passes it to the
+  engine is untouched by any of this.
+
+  **Construction is checkable, so the conformance suite's promise stays whole.**
+  `StoreFactoryContract` composes with the contract for your tier and routes
+  `make_store` through your own factory, so every behavioral and isolation
+  assertion then runs against stores built the way dex builds them. "The suite is
+  green" therefore still means a backend is correct *and* constructable, rather
+  than quietly meaning only the first. The packaging test that installs the wheel
+  in an environment with no access to this source tree now builds its backend
+  through a factory and a context carrying no repo root, because a construction
+  contract that only works from inside this repo has not been tested.
+
+  **No secret ever reaches a `StoreContext`.** `.dex/config.yml` is committed, so
+  a password, key, token, or connection string among the options would be a
+  credential in version control. A backend reads its credential from the
+  environment at construction, exactly as the connection and semantic-layer seams
+  already do, or a host with per-request credentials skips the contract and hands
+  the engine a store it built itself.
+
+  Both shipped backends refuse an option they would otherwise have ignored, and
+  `FilesystemStore.from_context` refuses a context with no repo root rather than
+  falling back to the working directory, which would write one project's
+  exploration cache into wherever the process happened to start. Accepted-and-
+  ignored is worse than rejected: the caller believes a setting took effect and
+  nothing in the output says otherwise.
+
+  `references/storage.md` documents the contract, both shapes a factory can take,
+  and the two rejected alternatives: always passing a `repo_root`, which cannot
+  construct a tenant-keyed backend at all, and requiring a `from_config`
+  classmethod, which puts the obligation back on the protocol.
+
 ## [1.4.2] - 2026-07-28
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,17 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   construct a tenant-keyed backend at all, and requiring a `from_config`
   classmethod, which puts the obligation back on the protocol.
 
+### Fixed
+
+- **`maintain snapshot` told every host to commit a file it may not have**
+  ([#157]). The hint was a fixed string: "commit `.dex/snapshot.json` like a
+  lockfile". A backend that keeps the baseline as a row or a document has no such
+  file, so the advice named something that does not exist, while the
+  `snapshot_path` beside it correctly reported that backend's own locator. The
+  half that holds everywhere, re-pinning after each known-good build, is now what
+  a backend dex does not ship gets; the git half is added only when the baseline
+  really is a file in the repo. Nothing changes for the filesystem backend.
+
 ## [1.4.2] - 2026-07-28
 
 ### Changed
--- a/packages/dex-core/README.md
+++ b/packages/dex-core/README.md
@@ -60,9 +60,12 @@ envelope never crosses this boundary.
 Nothing above touches disk. The default store keeps state in the process, so
 importing this package cannot leave a `.dex/` directory in a consumer's repo;
 pass `store=` for anything durable, or use `DexEngine.from_repo(repo_root)` to get
-the CLI's behavior (filesystem store, config read from `.dex/config.yml`). The
-`Store` protocol is public, so a host can back state with its own session store
-or database instead.
+the CLI's behavior (config read from `.dex/config.yml`, and the backend that
+config selects, which defaults to plain files under `.dex/`). The `Store` protocol
+is public, so a host can back state with its own session store or database
+instead, and a backend published as its own package is selectable by name from
+`cache.backend` without a change to dex. See
+[`references/storage.md`](../../references/storage.md).
 
 [`examples/quickstart.py`](examples/quickstart.py) is the whole flow in one
 runnable file: map a warehouse, read the inferred joins, see PII flagged, ask a
--- a/packages/dex-core/src/exmergo_dex_core/__init__.py
+++ b/packages/dex-core/src/exmergo_dex_core/__init__.py
@@ -95,6 +95,8 @@ _EXPORTS = {
     "SemanticSource": "connect",
     "Snapshot": "maintain.snapshot",
     "Store": "storage",
+    "StoreContext": "storage",
+    "StoreFactory": "storage",
     "StoreRequiredError": "errors",
     "to_envelope": "results",
 }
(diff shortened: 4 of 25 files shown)


catincloudlabs:
@marcociav-exmergo Findings from running our `Store` implementation against this
branch. It is `PrincipalStore`: the full `Store` tier, keyed per principal instead
of per repo root, composing one `FilesystemStore` per principal. It does not make
the read-then-write atomic. Instead `spend_since` adds other in-flight requests'
published reservations to the ledger read, so the ceiling binds for one key by
over-refusing rather than by a transaction or a conditional write.

## Results against `pull/173/head`

`StoreFactoryContract` mixed in front of `StoreContract`, our own factory, clean
Linux container:

* **36 collected, 36 passed.** No change to the backend was needed.
* `build_store` resolves our factory by dotted path from outside the package and
  returns a store. Its exit check (`isinstance(store, ExploreStore)`) passes, and
  the object returned also satisfies `MaintainStore` and `Store`. The factory
  lives outside our repo for now, since committing it would pin us to an unmerged
  API. We have not exercised the entry-point form.
* `StoreContext(repo_root=None, options={...})` builds a working store whose state
  lands under a root named in `options`. Ours ignores `repo_root` entirely, which
  is the shape #156 asks about.
* A `ConfigurationError` raised inside our factory reaches the caller intact
  rather than as `build_store`'s generic build failure.

At released 1.4.2 the same backend passes `StoreContract` 34 of 34, so these
numbers are not the PR changing a working backend.

## `[storage-conformance]` cannot run `StoreContract`

The extra is `["pytest>=8"]`, on this branch and at 1.4.2 alike. The 10 plan
assertions reach a module-level `import sqlglot`:

```
conformance.py:480            (1.4.2: :473)
  -> conformance.py:111       (1.4.2: :104)
  -> transform/__init__.py:26
  -> transform/plans.py:41
  -> transform/validate.py:17
  -> guards/sql_guard.py:14   ModuleNotFoundError: No module named 'sqlglot'
```

Released 1.4.2, `pip install "exmergo-dex-core[storage-conformance]==1.4.2"` and
nothing else: **10 failed, 42 passed.** The conformance module itself imports
cleanly, so this surfaces at assertion time rather than at import, as ten failures
reported against the backend under test. `dev` pulls `[sql,storage-conformance]`,
so an in-repo run does not reproduce it.

Predates this PR. `conformance.py`'s own docstring says to install the
`[storage-conformance]` extra "to get pytest alongside it", which is the line that
is incomplete for the plan tier.

## `options` is static, so per-user keying does not come from config

Our key is a per-request value. `options` reaches a factory from the config file
only: `--cache-backend` overrides the name for one run and, as `CacheConfig`
documents, leaves `options` behind. So a `cache:` block names one tenant, and
varying the key per request needs either a store instance passed to the engine or
a factory reading ambient process state.

`StoreContext`'s docstring already says `DexEngine(store=...)` is the right shape
for a host that already holds per-request credentials. The same holds for a host
whose per-request value is a key rather than a credential, and saying so next to
the `cache:` block in `references/storage.md` would settle where a multi-tenant
host is meant to look.

One instance of the boundary: our factory requires `max_request_spend`, the most
one in-flight command may reserve. For us that is the ceiling the host enforces on
a caller-supplied `--budget`, refusing anything above it rather than clamping. It
is not a secret, so `cache.options` accepts it, and the cap is then stated in two
places that can disagree silently. We refuse a context without it, and the reason
is measurable: a `PrincipalStore` built to reserve nothing passes all 36
assertions while bounding nothing.

## `make_store` builds a fresh instance per call

`make_store` is `build(context_for(key))`, so every call is a fresh instance.
Across `StoreContract`'s 34 assertions, 30 build one store, 4 build three, and
none builds the same key twice.

Where `spend_since` is not a pure read, two instances for one key behave as two
in-flight requests: on ours the first reads `0.0` and the second reads the
outstanding reservation. Nothing in the suite reaches that, and the contract
states the sharing requirement across different keys only, so whether
`make_store(k)` twice must return independent stores is unstated either way.

## Not filing

The `spend_since` into `append_spend_log` race, and store granularity being
ceiling granularity, are documented in `storage/base.py`, including that the
latter errs permissive. This backend is where we handle the second one, by the
reservation described above rather than by making the pair atomic.

marcociav-exmergo:
@catincloudlabs thanks for the report!
I filed the issue here: https://github.com/exmergo/dex/issues/174
We'll address it and add the fix in 1.4.3

catincloudlabs:
@marcociav-exmergo The `StoreContext` notes I owed you.

We run `PrincipalStore` in production: the full `Store` tier, composing one
`FilesystemStore` per principal so your serialization rules stay in one place and
ours is only the keying. Named by dotted path in `.dex/config.yml`, built through
`from_context`.

**What the construction contract had to carry for us.**

Our service shells out to the `dex` CLI per request, so `DexEngine(store=...)` is
not available to us: there is no engine object of ours to hand a store to.
`from_context` is the only door.

`options` is static because it lives in a committed file, so the value that varies
per request cannot travel in it. Our principal key arrives as `DEX_PRINCIPAL` in
the spawned process's environment, and the per-request cap as
`DEX_MAX_REQUEST_BUDGET`. Your docstring already gives that advice, scoped to
secrets: *"read the credential the way the rest of the engine does, from the
environment at construction time."* For a subprocess host it holds for any
per-request value, not only secret ones. We went looking for a way to make
`options` carry the key before we found that paragraph, which is framed as being
about credentials.

Two smaller confirmations: we ignore `repo_root`, which your docstring permits and
which is load bearing for us because explore and transform run against different
roots and keying on the root is the split this backend exists to collapse; and we
refuse an unhonored option rather than dropping it.

**`#174` is fixed.** `[storage-conformance]` resolves `exmergo-dex-core[sql]`, so
`StoreContract` runs from a clean install with no dev extra: 34 tier assertions
plus `StoreFactoryContract`'s 2, green first run, no changes to our backend. Every
file under `storage/` is byte identical from `v1.4.3` to `v1.5.0`.

**Two things we would want to know if we maintained it.**

1. `_namespaced_keys` closed the reused-key bug and closed the reuse case with it.
   Every assertion gets `{run}-{seq}` and every multi-store assertion uses two
   different labels, so nothing ever rebuilds a store from a key it already used.
   The suite proves isolation between keys and says nothing about persistence for
   one key, which for us is the property that matters most: the principal has to
   find its own state tomorrow. A backend returning a fresh empty store on every
   construction would pass all 34. We cover that ourselves.

2. `maintain/commands.py:173` picks the snapshot hint with
   `isinstance(engine.store, FilesystemStore)`. `PrincipalStore` composes one
   rather than subclassing it, so that reads `False` for us while
   `locator(Document.SNAPSHOT)` still returns a real reviewable path. We write the
   file the long hint describes and get the short hint. It errs safe. It is also
   the one `isinstance` on a concrete class inside a seam whose premise is that a
   class with the right methods is a store.

Neither is blocking for us.

PR #175, opened 2026-07-30
SCRUM-1009-Fix explore cluster reporting dropped_null_rows: 0 despite dropping rows

### LINKED ISSUE #160: `explore cluster` reports `dropped_null_rows: 0` while the sample SQL drops the rows
**Version:** observed live at 1.3.1; mechanism re-verified at 1.4.2. BigQuery connector.

## Summary

`build_sample_sql` puts an `IS NOT NULL` predicate for every feature column into the sample SQL itself, so rows with any null feature are removed **by the warehouse** and never arrive in Python:

```python
# explore/cluster.py:154
f"{render(exp.column(name))} IS NOT NULL" for name in feature_names
```

`_clean_rows` (`explore/cluster.py:227`), which produces the reported `dropped_null_rows`, only sees rows that did arrive. Its docstring is accurate about itself. The consequence is that **`dropped_null_rows` is structurally near-always `0`**: it can only ever count non-numeric coercion failures, never nulls.

Both behaviours are individually correct. Filtering in SQL is right, and it is why the scan stays narrow. Counting what you dropped is right. The defect is that the two live in different layers and the envelope reports the inner one under a name that reads like the outer one.

## Evidence

First real run of the route against production data, one API call:

```
total_rows           267     (cache-derived)
n_samples            248     (live)
dropped_null_rows      0     <-- 21 rows were dropped for nullity
live row count       269
```

269 − 248 = **21 rows, 7.8% of the table**, every one of them missing the same single feature column. No note and no warning mentioned the exclusion. Verified numerically with a separate free query rather than inferred: exactly one feature column had any nulls, and its null count was 21.

## A second wrinkle, same envelope

`total_rows` is read from the `.dex` cache (last written by `explore map` / `profile`) while `n_samples` is computed from the live sample. So the only two numbers a careful reader could subtract to *notice* the gap come from different moments. Above, that subtraction reads as 19 against a true 21, wrong in the reassuring direction, and it is the sole available signal that anything was excluded.

## Why it matters analytically

Nulls are rarely random. A single column being null across a subpopulation removes exactly that subpopulation from the clustering, and that is frequently the segment the exercise exists to find. The result is a segmentation of a biased sample, presented with a completeness field that reads as reassurance.

A caller cannot currently tell "clustered the table" from "clustered the 92% of it that had no nulls", and the one field named for that distinction says zero.

## Expected, cheapest first

1. **Report the drop where it happens.** `build_sample_sql` knows the predicate it emitted, so the count is one `COUNT(*)` on the same filter, or a difference against the live row count the sample clause already needs. Populate `dropped_null_rows` from that, **or** rename the current field to `dropped_non_numeric_rows` and add the SQL-side count alongside it.
2. **Attribute it.** Per-feature null counts turn "8% dropped" into "8% dropped, all from one column", which is the difference between an alarming number and an actionable one. A `notes` entry is enough, and there is precedent in the same envelope (`excluded 1 unique-key column(s)`).
3. **Make `total_rows` and `n_samples` commensurable**, or label `total_rows` as cache-derived. Two numbers a reader is invited to subtract should come from the same instant.

(1) alone removes the misleading zero. (2) makes the result trustworthy rather than merely honest.

## Related

emnarahmouni-exmergo:
Closes https://github.com/exmergo/dex/issues/160

Fixes explore cluster reporting dropped_null_rows: 0 while its own sample SQL was silently dropping rows with any null feature.

build_sample_sql's IS NOT NULL filter excludes null rows in the warehouse before Python ever sees them, so the old count only ever measured post-arrival non-numeric coercion failures never nulls, structurally
Added a companion count query (build_null_count_sql), same table and sample scope as the fetch, that measures exactly what the filter excludes and attributes it per feature column
Renamed the clustering engine's own count to dropped_non_numeric_rows (what it actually measures); dropped_null_rows is now a real, SQL-measured top-level field
Nonzero drops get a notes entry naming the responsible column(s) (e.g. visits: 20) and flagging that total_rows is cache-derived, a different moment than the live count
Cost note: billed connectors see ~2x the estimate for explore cluster (two queries instead of one), same order of magnitude as the sample fetch, not a full-table scan
Tests: 8 new (per-dialect SQL builder tests + an end-to-end DuckDB test proving 20 known-null rows are correctly reported and attributed, not swallowed as 0), 1 existing test updated for the new combined cost estimate

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,29 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **`explore cluster` reported `dropped_null_rows: 0` while its own sample SQL
+  was silently dropping rows** ([#160]). `build_sample_sql` puts an
+  `IS NOT NULL` predicate for every feature column into the sample query
+  itself, so rows with any null feature are filtered by the warehouse and
+  never reach Python; the reported count only ever saw rows that arrived, so it
+  could count a non-numeric coercion failure but never a null, structurally.
+  One production run silently clustered 92% of a table (the missing 8% shared
+  a single null feature column) with no note or warning. `explore cluster` now
+  runs a companion count query, over the same table and the same sample scope
+  as the fetch, that measures exactly what the null filter excludes; a nonzero
+  count gets a notes entry attributing it to the responsible feature column(s)
+  (`"visits: 20"`, not a bare number) and flags that `total_rows` is
+  cache-derived, a different moment than this live count. The clustering
+  engine's own count (rows that arrived but failed float coercion, which is
+  what the old field actually ever measured) is renamed
+  `dropped_non_numeric_rows` so the two are never conflated again.
+
+  **Cost note**: on a billed connector, `explore cluster`'s estimate roughly
+  doubles, since two queries now price into it instead of one; each still
+  scans only the feature columns over the same sample scope, so the added cost
+  is the same order of magnitude as the sample fetch itself, not a full-table
+  scan.
+
 - **`maintain snapshot` told every host to commit a file it may not have**
   ([#157]). The hint was a fixed string: "commit `.dex/snapshot.json` like a
   lockfile". A backend that keeps the baseline as a row or a document has no such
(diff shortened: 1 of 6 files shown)


marcociav-exmergo (approved):
LGTM!

PR #176, opened 2026-07-30
SCRUM-1015-Derate low-cardinality name-column PII flags via value-shape cardinality

### LINKED ISSUE #167: PII detection: low-cardinality enumerations keep a `name` flag through value-shape profiling
**Version:** observed live at 1.3.x and 1.4.x, BigQuery connector. Related to but distinct from #112 and #106, both closed.

## Summary

A `name`-category PII flag at 0.6 confidence, above `PII_BLOCK_CONFIDENCE = 0.5`, survives schema-3 value-shape profiling on a column with **7 distinct values**.

Concretely, on a conventional date dimension: the string column holding weekday names (7 distinct values) and the one holding month names (12 distinct values) are flagged as `category: name` on every fresh re-profile, and the projection is then refused with a 422. That blocks ordinary queries against what is typically the most-joined dimension in a warehouse.

## Why this is not #112

#112 de-rated `email` and `phone` on **non-string** columns, on the reasoning that an `INT64` cannot hold an email address. These columns are strings, so type evidence does not reach them.

The signal that should reach them is **cardinality**. A column holding 7 distinct values across thousands of rows is an enumeration, not a person. Value-shape corroboration is exactly what schema 3 advertises, and it does adjust name-based flags in both directions elsewhere (person-shaped values corroborate, reference vocabularies de-rate). A closed set of 7 repeated short tokens is about as clear a negative signal as the value shape can offer, and it is not currently being used.

## Why this is not #106 either

The pattern form for `pii_overrides` is the workaround's ergonomics, not the detector.

One integrator deliberately did **not** use the pattern form here, with reasoning worth repeating: each cleared column is a per-column human decision about a specific table, and the pattern form would silently clear any same-named column added to a matching table later. They used the exact, fully qualified form so that every entry is a claim a reviewer can check in a pull request rather than assume.

## Status downstream

Worked around with four exact-form `pii_overrides` entries, each carrying its justification in a comment. **The detector is unchanged**, so each new `%_name` column arrives blocked until a human clears it by hand, and the override list grows with the warehouse.

## Framing, because the direction of the error matters

**This is not a request to loosen the threshold.** Blocking is the right direction to be wrong in, and over-flagging is preferable to under-flagging for PII. A detector tuned to stop annoying its operator is one that will eventually let something through.

The ask is narrower: **value-shape evidence should be able to down-weight a name-only signal when the values are visibly an enumeration.** That is the mechanism schema 3 already exists to provide, applied to a case it currently misses.

## Expected

Low-cardinality string columns whose values are a small closed set of short repeated tokens should have a name-only category flag de-rated below the blocking threshold, or the flag should be retained with a note explaining what corroborated it, so an operator can tell a real hit from a suffix match.

A reasonable guard on the guard: only de-rate when distinct count is small in both absolute terms and as a fraction of rows, so a genuinely small table of people is not cleared by the same rule.

emnarahmouni-exmergo:
Closes https://github.com/exmergo/dex/issues/167
PR description:

Fixes low-cardinality enumerations (weekday names, month names, status codes) keeping a blocking name PII flag through value-shape profiling.

A string column matching the generic *_name pattern starts at 0.6 confidence, above the query firewall's 0.5 blocking threshold
Existing value-shape rules only de-rate all-caps reference vocabularies or long multi-token labels; single-token Title Case enumerations (Monday, January) matched neither, so a date dimension's weekday/month columns stayed blocked on every re-profile
Added a third corroborating rule: distinct count small both in absolute terms (≤32) and as a fraction of non-null rows (≤1%) de-rates the same way the existing rules do
Guard on the guard, per the report: a genuinely small table of distinct people has a low absolute count but a high fraction, so it's correctly excluded; person-shape corroboration still takes priority over cardinality
The flag is never removed only where it lands relative to the blocking threshold
Tests: 7 new parametrized unit tests on _refine_confidence (weekday/month cases, absolute cap, fraction guard both directions, missing row-count, person-shape priority) + 1 end-to-end DuckDB test with a real date-dimension fixture

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,25 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **Low-cardinality enumerations (weekday names, month names, status codes) no
+  longer keep a blocking name-only PII flag** ([#167]). A string column matching
+  the generic `*_name` pattern (`day_name`, `month_name`) starts at 0.6
+  confidence, above the query firewall's 0.5 blocking threshold; value-shape
+  profiling can already de-rate a name-only flag when the values are visibly an
+  all-caps reference vocabulary or long multi-token labels, but a closed set of
+  single-token Title Case values (`Monday`, `January`) matched neither rule, so
+  a conventional date dimension's weekday/month columns stayed blocked on every
+  fresh re-profile. Cardinality is now its own corroborating signal: a column
+  whose distinct count is small both in absolute terms and as a fraction of
+  non-null rows de-rates the same way the existing shape rules do. The fraction
+  half is the guard on the guard, verbatim from the report: a genuinely small
+  table of distinct people has a low absolute distinct count but a *high*
+  fraction (most rows are their own distinct value), so it is not cleared by
+  this rule, and a person-shaped distribution still corroborates as a real name
+  before cardinality is ever considered. The flag itself is never removed,
+  consistent with every other shape rule; only where it lands relative to the
+  blocking threshold.
+
 - **`maintain snapshot` told every host to commit a file it may not have**
   ([#157]). The hint was a fixed string: "commit `.dex/snapshot.json` like a
   lockfile". A backend that keeps the baseline as a row or a document has no such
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
LGTM

PR #177, opened 2026-07-30
SCRUM-1023-Declare a composite grain via unique_combination_of_columns

### LINKED ISSUE #169: `DeclaredKey.column` is a single `str`, so a composite grain cannot be declared
**Version:** measured against 1.2.1, re-read at 1.3.1, re-verified at 1.4.2. Evidence from a production integration driving dex against a composite-grained warehouse.

## Summary

`DeclaredKey.column` is a single `str`, and `_primary_entity_column` returns exactly one column:

```python
# dbt_project.py:709-717
class DeclaredKey(BaseModel):
    """A column carrying ``unique`` and/or ``not_null`` tests on one model."""
    model: str
    relation: str | None = None
    column: str          # <-- single
```

A composite grain therefore cannot be **declared**. This is structural: a column-level `unique` is a dbt column test, and a semantic model's primary `entities:` is one column, so the ceiling comes from the source format as much as from dex's model of it.

## Measured, not argued

An integrator declared 13 keys and 11 joins across a reporting layer and diffed elected grains before and after: **0 of 12 elected grains moved.**

The channel's reach is precisely the complement of where it is needed. The tables whose grain a column-level `unique` *can* express already had that grain elected correctly by measurement, so declaration is redundant exactly where it is legal. The tables that need it cannot use it.

## The forced-silence half is the sharp bit

Declaring one column of a composite key is not merely insufficient, it is **actively harmful**:

- dex holds it as a permanent contradiction and fires false high-severity `key_lost_uniqueness`
- for a semantic model's `entities:` specifically, it **overrides** a correctly inferred composite grain with a wrong single-column one

So on a composite-grained table the only safe move is to **declare nothing**. And "declared nothing because a composite cannot be expressed" is indistinguishable, from dex's side and from a reviewer's, from "has not got round to declaring it yet". Silence is the correct declaration, and it looks like neglect.

## The other half of the channel works well and should be left alone

The same project's 11 declared joins went from 0 to 11 once declared. Inference finds none there (no id-shaped column matched a parent table by name), so **every real dimensional join dex knows about in that project exists only because a human stated it.** The declared-relationship channel is earning its keep; it is only the key half that cannot reach.

## Expected

Widen `DeclaredKey.column` (or add an equivalent) to express a composite key, so a project can state a grain that measurement cannot reach.

Two things worth deciding alongside:

- **Where the declaration comes from.** dbt has no native composite `unique` at column level, so this probably wants a model-level construct (a `unique_combination_of_columns`-shaped test, or a dex-specific key declaration) rather than a widening of the column-test parse.
- **Interaction with election.** A declared composite should presumably win over a measured single-column election rather than contradicting it, which is the opposite of the current single-column behaviour described above.

## Related

Pairs with #168: measurement cannot reach a two-column grain that ranks below three junk pairs, and declaration cannot express it either. Together they leave a real grain with no route in.

Both are instances of the shape collected in #172.

emnarahmouni-exmergo:
Closes https://github.com/exmergo/dex/issues/169
Fixes DeclaredKey.column being a single string, so a project could never declare that a table's real key is the COMBINATION of several columns and worse, declaring one member column of a composite key made dex hold it as a permanent contradiction, while a semantic model's declared primary entity could silently override a correctly-measured composite grain with a wrong single column.

New DeclaredCompositeKey reads dbt's own unique_combination_of_columns test both the dbt-core 1.9+ built-in and the dbt_utils macro compile to the same stripped-namespace test name from the compiled manifest and from schema YAML's model-level tests:/data_tests: block (a blind spot that never existed before). No new dex-specific schema.
A resolved, column-existence-checked declared composite now overrides a measured/heuristic grain unless the current grain is already a measurement-proven single column, in which case the proven single wins and a note records the composite was also declared
Because the declared composite lands in Dataset.grain as a multi-column list, it automatically routes through maintain grain's existing combination-check path (never the single-column path), so it structurally cannot reproduce the false key_lost_uniqueness bug the old single-column declaration caused
Tests: 3 parsing tests (manifest, bare YAML, namespaced dbt_utils. YAML), 5 direct unit tests on the election logic, 1 integration test, 2 end-to-end maintain grain tests including a named regression pinning that declaring one column of a composite as unique (the issue's opening complaint) still never fires a false key_lost_uniqueness
CHANGELOG.md updated under [Unreleased] / Added

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,27 +13,27 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 - **A project can declare a composite grain via `unique_combination_of_columns`**
   ([#169]). `DeclaredKey.column` was a single string, so a project could only
-  ever declare that ONE column is unique -- never that a table's real key is
-  the COMBINATION of several. This wasn't merely insufficient: on a genuinely
+  ever declare that ONE column is unique, never that a table's real key is the
+  COMBINATION of several. This wasn't merely insufficient: on a genuinely
   composite-grained table, declaring one member column (the only thing the old
   format allowed) made dex hold it as a permanent contradiction, and a
   semantic model's declared primary entity could unconditionally override a
   correctly-measured composite grain with a wrong single column. Measured
   evidence from the report: 13 declared keys across a reporting layer moved 0
-  of 12 elected grains -- the channel only ever reached tables that already had
+  of 12 elected grains; the channel only ever reached tables that already had
   a correctly-measured single-column grain.
 
   A new `DeclaredCompositeKey` reads dbt's own `unique_combination_of_columns`
   test (both the dbt-core 1.9+ built-in and the `dbt_utils` macro compile to
   the same stripped-namespace test name) from the compiled manifest and from
-  raw schema YAML's model-level `tests:`/`data_tests:` block -- no new
+  raw schema YAML's model-level `tests:`/`data_tests:` block, with no new
   dex-specific schema. A resolved, column-existence-checked declared composite
   now wins over a measured or heuristic grain, *unless* the current grain is
   already a measurement-proven single column, in which case the proven single
   stays and a note records that a composite was also declared. Because the
   declared composite lands in `Dataset.grain` as a multi-column list, it flows
-  through `maintain grain`'s existing combination-check path automatically --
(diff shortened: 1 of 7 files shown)


marcociav-exmergo (approved):
LGTM!

PR #183, opened 2026-07-31
SCRUM-1006-Fix explore map carry-forward resurrecting relations deleted from the…

### LINKED ISSUE #149: `explore map` carry-forward resurrects relations deleted from the warehouse; `maintain snapshot` then pins them as false `table_dropped`
**Version:** dex 1.4.0, BigQuery connector. Edge of the #111 fix (PR #132).

## Summary

`explore map` unions prior-cache profiles back in for any object the run did not examine, but never intersects that set against the **live inventory**. A relation dropped from the warehouse between two maps is therefore re-added to `.dex/cache.json` from the prior cache, carrying its original `profiled_at`. `maintain snapshot` then pins the ghost, and the next `maintain check` reports it as `table_dropped` at **severity high** — the top of a blast-radius-ranked report — for a relation the operator deliberately deleted.

## Repro

1. Cache and snapshot in a normal state; warehouse has >50 objects, so `map` applies its rank cutoff and most objects are "not examined".
2. Drop three relations that no longer have backing models (here: one intermediate view and two marts tables, residue of a layer move and a `dim` → `fct` rename). Dropped via `dbt run-operation`, i.e. outside dex, which is the only route available pre-#146.
3. `explore map --confirm --budget <n>` — succeeds.
4. Inspect the cache: **all three dropped relations are present**, with `profiled_at` stamps 3–5 days old, carried from the prior cache.
5. `maintain snapshot` — succeeds, no warning.
6. `maintain check` → **3 findings, `code: table_dropped`, `severity: high`**, naming exactly the three relations from step 2: `"<relation> is in the snapshot but gone from the warehouse"`.

## Diagnosis

`_carry_forward_relationships` documents the intent precisely — union back anything "this run could not possibly have regenerated or superseded", keyed on endpoints outside the `examined` set — and `_compose_datasets` guards datasets the same way per #111.

The membership test is *"was not examined this run"*. It needs to be *"was not examined this run **and still exists**"*. Absent from the live inventory is not the same as out of scope, but carry-forward currently cannot distinguish them.

## Why this matters more now that #146 has landed

With `orphan_relation` shipping in 1.4.1, the intended lifecycle becomes:

> dex classifies an orphan → operator drops it → operator re-maps

Step three resurrects what step two removed. The remediation loop dex now recommends does not converge, and the resurrected object is a strong candidate to be re-classified as an orphan on the next pass.

## Expected

`explore map` intersects carry-forward candidates against the current inventory and drops any object no longer present, reporting the count — a `dropped_count` counterpart to the `carried_forward_count` #111 asked for. At minimum, an envelope warning naming the objects being resurrected.

## Workaround

`rm .dex/cache.json` before re-mapping. That costs a full re-profile (2.62 GiB here) and is only discoverable by re-running `check` after a snapshot and disbelieving the result.

## Note

Filed alongside a second, independent defect observed in the same run (partially-profiled cache pinned by `maintain snapshot`). They compound but have separate fixes.

emnarahmouni-exmergo:
Closes https://github.com/exmergo/dex/issues/149

Fixes explore map resurrecting a relation deleted from the warehouse via its carry-forward mechanism, which maintain snapshot then pins and maintain check reports as a false high-severity table_dropped.

Carry-forward's "not examined this run" test couldn't distinguish "outside --scope/--dataset" (issue #111, correct to carry forward) from "gone from the warehouse" (should be dropped, not resurrected)
Fix derives the schema/dataset namespaces this run's inventory actually observed; an unexamined prior identifier is only carried forward when its own namespace was never observed, and dropped when the namespace was observed but the object itself is missing
explore map's envelope now reports dropped_count and a warning naming what was dropped
explore relationships shares the same carry-forward function and gets the fix for free; explore profile is untouched (no full inventory scan, no reliable "observed" signal)
Now actively supports the orphan_relation (#146) remediation loop instead of undoing it: classify an orphan, drop it, re-map, and the re-map no longer resurrects what was just cleaned up
Known accepted gap: an entire schema emptied out (not just a few relations) still looks identical to "never in scope" and resurrects narrower than the reported repro, left for a future fix
Tests: 4 new unit tests (drop case + preserved-#111 control, at dataset and relationship level), 1 end-to-end explore map test, 1 full-loop regression proving no false table_dropped survives drop → remap → snapshot → check
Before the fix : 
image
After the fix : 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,37 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`explore map` no longer resurrects a relation deleted from the warehouse**
+  ([#149]). Carry-forward unions back any prior cached profile this run "did
+  not examine," correct when that means outside a narrower `--scope`/
+  `--dataset` (issue #111), but the same test also fired when an object was
+  simply dropped between two maps: the prior profile came back with its
+  original `profiled_at`, `maintain snapshot` pinned the ghost as a real
+  dataset, and the next `maintain check` reported it as a high-severity
+  `table_dropped` finding for something the operator deliberately removed.
+  With `orphan_relation` (#146) shipped, this actively worked against dex's
+  own remediation loop: classify an orphan, drop it, re-map, and the re-map
+  resurrected exactly what was just cleaned up.
+
+  Neither the cache nor any adapter records what scope built a prior run, so
+  "not examined" could not tell "out of scope" from "gone" apart. The fix
+  derives the schema/dataset namespaces this run's inventory actually
+  observed; an unexamined prior identifier is carried forward only when its
+  own namespace was never observed (genuinely out of scope, so #111 still
+  holds), and dropped, not carried, when the namespace was observed but the
+  object itself is missing from it. `explore map`'s envelope now reports a
+  `dropped_count` and a warning naming what was dropped. `explore
+  relationships` shares the same carry-forward function and gets the fix for
+  free; `explore profile` is unaffected, since neither does a full inventory
+  scan wide enough to know what "observed" means.
+
+  Known gap: if an entire schema is emptied out (every object in it dropped,
+  not just a few), it still looks identical to "never in scope" and
+  resurrects; narrower than the reported repro (a handful of relations gone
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
LGTM

PR #186, opened 2026-08-01
SCRUM-1024-A machine-readable `reason` on every error envelope, alongside `status`

### LINKED ISSUE #170: No machine-readable refusal reason on the envelope (CLI-side follow-up to #155)
**Version:** 1.4.2, after #155. Narrow follow-up, not a re-ask.

## Two precision notes first, because this is easy to state wrongly

1. **`Status` is already a 4-value enum**, so the *class* of outcome already crosses the boundary. What does not cross is the **why within an `error`**, because `errors` is `list[str]` of free prose.
2. **`reason` already exists** at three sites in the codebase. The `Envelope` is not one of them.

So this is not "there is no reason field". It is that the envelope, which is the CLI's entire machine-readable surface, cannot distinguish one refusal from another.

## Summary

#155 gave every refusal a typed exception a **library** consumer can catch, rooted in `DexError`. That is the right fix for a host embedding the engine, and it closes the axis for the programmatic API.

It does not reach a host driving the **CLI**, which still receives an envelope carrying prose and must infer intent from it. `OverCeilingError` and a BigQuery connection fault arrive as the same shape: `status: error`, `errors: ["..."]`.

## What that costs, concretely

A host mapping the envelope onto HTTP statuses its automation branches on has to pattern-match error text. Those statuses are **instructions**: 409 means "run the setup command and retry", 422 means "stop, this is a policy violation". Getting the mapping wrong does not produce a wrong answer, it produces a wrong *behaviour*.

Before that host anchored its patterns, `connection refused: could not connect to bigquery` scored as 422. **A transient network fault became a permanent refusal the worker was forbidden to retry.** A blip became an outage by classification.

They now anchor on the exact shapes dex emits rather than searching for substrings, and carry a canary test asserting their regex still matches dex's own `_NO_SNAPSHOT_ERROR` constant, so a reword here fails loudly in their CI instead of silently downgrading a retry into a stop.

**A test in a consumer's repo defending a string constant in ours is the wrong place for it.** It also means our error prose is now a de facto API surface we cannot reword freely.

## Interaction with the paradigm default

The one structured field a caller would naturally use to disambiguate a spend refusal from a connection fault is `cost.paradigm`, and it currently reports `free_local` on every refusal envelope (#162). So the structured escape hatch that exists today actively points the wrong way. That one is a small fix and worth doing first regardless of what happens here.

## Expected

A typed `reason` or code on the envelope, **alongside** `Status` rather than replacing it. Enough to answer:

- was this refused by a guard (cost, PII, firewall), or did something fail?
- is the caller expected to retry, to run a setup step first, or to stop?

The second question is the one a host actually branches on, and it is currently answerable only by reading English.

## Note

The prose messages are good and should stay. This asks for a code beside them, not instead of them.

emnarahmouni-exmergo:
Closes https://github.com/exmergo/dex/issues/170
Adds a closed Reason enum (guard, prerequisite, connection, configuration, request, execution_failure, internal) and a reason field on Envelope, populated only when status == "error".
reason_for(exc) classifies via isinstance against #155's existing exception hierarchy (most-specific-first); error_for(exc, message=None, **kwargs) is the new envelope constructor wired into all 17 existing env.error(...) call sites across cli.py, explore/commands.py, explore/semantic/commands.py, maintain/commands.py, and transform/commands.py.
Classification is two-tiered: a small always-safe set (from errors.py and guards/dialect.py, both import-light) resolves first, then a broader set that pulls in sqlglot/scikit-learn/dbt is attempted behind a try/except ImportError — so classifying a missing-extra refusal can't itself crash the process on that same missing extra. Caught by the project's own isolated-wheel packaging test.
Prose error messages are unchanged; reason is additive, for hosts that want to branch on retry/setup/stop semantics without parsing strings.
BEFORE the fix : 
image
AFTER fixing : 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,45 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **A machine-readable `reason` on every error envelope, alongside `status`**
+  ([#170]). #155 gave every deliberate refusal a typed exception a library
+  consumer can catch, but a host driving the CLI still only ever saw
+  `status: error` and a prose string: `OverCeilingError` and a BigQuery
+  connection fault arrived in the identical shape. A host mapping the
+  envelope onto its own automation had to pattern-match error text to tell a
+  policy refusal from a transient fault, and got it wrong in the field: a
+  `connection refused` blip scored the same as a real policy violation and
+  was treated as a permanent stop rather than something to retry.
+
+  `reason` is one of seven values (`guard`, `prerequisite`, `connection`,
+  `configuration`, `request`, `execution_failure`, `internal`), derived
+  automatically from the caught exception's class, never invented per call
+  site: `guard` covers a cost/PII/firewall/safety policy declining
+  (`OverCeilingError`, `QueryRefusedError`, a production-target refusal);
+  `prerequisite` covers a named setup command that needs to run first, then
+  retry (`CacheRequiredError`, `NoBaselineError`, a missing optional extra);
+  `connection` covers an unreachable warehouse; `configuration` covers the
+  engine or repo being wired without something it needs; `request` covers
+  bad input to this specific call; `execution_failure` covers an operation
+  that ran (a `dbt build`) and failed partway; `internal` is reserved for
+  whatever is not a deliberate dex refusal at all. `reason` is `null` outside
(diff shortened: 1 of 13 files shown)


marcociav-exmergo (approved):
LGTM

Issue #171, opened 2026-07-29
[core] Make the dormant `ProjectAdapter` seam load-bearing; a second non-dbt implementation is on offer


marcociav-exmergo:
**Version:** measured at 1.2.1 and 1.3.1; re-verified at 1.4.2.

**This needs a decision more than a fix.** A "the seam is retired" answer is genuinely useful and closes it.

## The seam is dormant, not merely bypassed

`adapters/project.py` defines the `ProjectAdapter` Protocol (`name`, `load()`, `write_edits()`) with a docstring framing it as the extension point for non-dbt project formats.

**At HEAD it has exactly one reference in the entire distribution: its own `class` statement.** `DbtProject` is never constructed anywhere.

Meanwhile **19 modules reference `dbt_project` directly**, counting neither the seam itself nor `dbt_project.py`. That was 18 at 1.3.1.

## The two consumers added since that count are the interesting part

They are `engine.py` and `storage/base.py`. The new public programmatic API from #138 and #144 also reads projects through `dbt_project` rather than through the seam.

So every new project-reading consumer across three minor releases has routed around it, **including the ones written specifically to make the engine embeddable.** That is not a complaint. It is the strongest available evidence that the seam will not become load-bearing as a side effect of ordinary feature work, and that a first non-dbt implementation is what would force the question.

## The offer

A production host runs an orchestrated Python transformation graph on BigQuery with no dbt project, and generates a dbt-shaped tree from its asset graph, regenerated daily and swapped whole, live for months. That generator already performs the exact source-graph to `TransformLayer` / `SemanticLayer` reduction a real adapter's `fingerprint()` would do, just serialized through the filesystem instead of returned as objects.

They are volunteering to build and maintain the second, non-dbt reference implementation, carrying the integration cost, and reporting back what the contract actually needs. If a non-dbt project format is on the near-term path, that de-risks it. If the seam is set aside, they would rather hear that than keep building toward it.

## Two pieces of evidence surfaced by doing it

[...]

PR #192, opened 2026-08-04
Tier the project seam, and ship a conformance contract for it

catincloudlabs:
Closes nothing on its own; this is the prototype #171 asks for, and the decision that issue wants is still yours to make.

#171 asked which of two contract shapes to prototype. We said on the issue we would bring a third that contains both, tiered like `ExploreStore` / `MaintainStore` / `Store`. This is that, built against 1.5.1.

It is deliberately scoped so that **none of the three questions we asked on the issue has to be answered for it to land**, and none of the three answers you offered there is foreclosed by it. In particular it does not touch construction: dex builds `DbtProject` itself, at the one place explore reads a project, exactly as it called `dbt_project` before.

## What it does

`adapters/project.py` becomes three nested protocols:

| tier | protocol | adds |
|---|---|---|
| 1 | `ExploreProject` | `definitions() -> ProjectDefinitions` |
| 2 | `MaintainProject` | `transform_layer()`, `semantic_layer()` |
| 3 | `EditableProject` | `write_edits()`, unchanged from today |

`tier_of(project)` reports the highest tier an object satisfies, by `isinstance` rather than declaration, so a format cannot claim a tier it has not implemented. `DbtProject` implements all three, every member delegating to a function that already existed.

**One line makes it load-bearing.** `explore/commands.py:1465`:

```diff
-    return dbt_project.definitions(repo_root, pin)
+    return DbtProject(repo_root, pin).definitions()
```

That is the whole rewiring. `DbtProject` is now constructed for the first time, and tier 1 sits on the real explore read path. `maintain` still calls `snapshot.transform_layer(view)` directly at its six call sites: tier 2 is implemented and contract-tested, but not rewired, because rewiring it is a bigger change than this PR should carry and the methods being present is what makes the tier real.

Two corrections to what we said on the issue, both found by building it:

**Tier 2 is two methods, not one `fingerprint()` returning a union.** Our comment on #171 proposed `fingerprint() -> TransformLayer | SemanticLayer`. That was wrong for your tree: `maintain.snapshot` already exposes exactly `transform_layer(view)` and `semantic_layer(view)`, and `maintain.commands` already calls both. A union return would have to be unpacked at every one of those call sites for no gain.

**The tiers are not named `Project` at the top.** `ExploreStore`/`MaintainStore`/`Store` was the model, but dbt itself ships two classes called `Project` (`dbt.config.project`, `dbt.contracts.project`), so the unqualified noun would collide for anyone reading both. Hence `EditableProject` for the write tier, named for the capability.

## The contract found a real bug on the first run

`adapters/conformance.py` ships the seam's rules as assertions, the way `storage/conformance.py` does. Its most important assertion is behavioural, and it is the one we flagged on the issue as the property a second implementation is most likely to get wrong: **`definitions()` must not raise.**

It went red against the shipped dbt format.

`load` parses `dbt_project.yml` with `yaml.safe_load`, so a project file that is not valid YAML arrives as `yaml.YAMLError`, and `definitions()` caught only `DbtProjectError`. The docstring two lines above the `try` promises that an unreadable project yields the empty view and never an exception; `explore --use-project` raised out of it instead.

The fix is to pair the two exceptions, which is what the three profile readers in that same module already do (`dbt_project.py:321`, `:356`, `:386`). It is a one-line change plus a regression test, and it is in this PR because shipping a contract whose reference implementation fails it is not an option.

**One adjacent instance we did not touch:** `_duckdb_dev_path`'s `try` block at `dbt_project.py:415` is byte-identical to the one at `:317` but also catches only `DbtProjectError`. It is not on the project-seam read path, so it is out of scope here rather than forgotten.

## Re-measured at 1.5.1, with the rule stated

The counts in #171 hold, and one moved:

- `ProjectAdapter`: still **one** reference before this PR, its own `class` statement. Nothing referenced it, tests included.
- `DbtProject`: **never constructed** anywhere, tests included.
- Modules referencing `dbt_project` directly: **19 → 20**.

That last one needs its rule, because the two rules disagree. Under *"module references `dbt_project`"* it is 20 at 1.5.1. The newcomer is `envelope.py:263`, which imports `DbtProjectError` to classify a refusal reason: under *"module reads a project through it"* it is still 19. It arrived with the `reason` work in 1.5.1, so it is not a project read routing around the seam, and we are not counting it as one.

`definitions()` itself is unmoved at `dbt_project.py:773`, still one caller (`_project_definitions`, `explore/commands.py:1433`, ending `:1465`), reached from `:279`/`:416`/`:759`.

## Deliberately not here

- **The construction contract.** The currency question, the first of the three we asked on the issue and the one we most wanted the call on, is still open and this PR does not pre-empt it. `references/project.md` states it as open and says why picking now would mean picking the directory-shaped answer on the strength of the only format that exists.
- **Tier 3's shape.** `write_edits(edits) -> Any` is moved, not redesigned.
- **`definitions()` taking a view.** Its four disk touches (`discover_projects`, `load`, `_read_semantic_manifest`, and `_flag_stale_manifest`'s `rglob` + `stat` walk) are a separate and larger argument.
- **Anything in `maintain/`.** No call site there changes.

## Proving it out of tree

`ProjectContract` is the piece we said we would want landed even if everything else came out smaller, because it is what makes #144's criterion checkable by you without running our code. We are wiring our own Dagster format up against it next, from a separate distribution depending only on the published wheel.

The extra needs **only pytest**: nothing in the contract reaches the dialect engine, and `test_packaging.py` holds that true. That is a difference from `[storage-conformance]`, and worth keeping rather than a coincidence.

## Checks

- Full suite on this branch: same result as `v1.5.1` before it, plus 13 new tests.
- `ruff check` and `ruff format --check` clean across the repo (162 files).
- `scripts/check_no_em_dashes.py` clean on every prose file touched.
- The behavioural assertion and the regression test were both confirmed red without the `dbt_project.py` fix and green with it.
- One pre-existing failure is unrelated and present on `v1.5.1` too: `test_dev_target.py::test_retargeted_dev_database_is_refused_naming_both_values` asserts `analytics/profiles.yml` while the message is built with a native separator, so it fails on Windows only. Happy to send that separately if it is not already known.

If the answer to #171 is that the seam is set aside or retired, this closes without hard feelings and we document the gap on our side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,49 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **The project seam is tiered, and load-bearing** ([#171]). `ProjectAdapter` had
+  exactly one reference in the distribution, its own `class` statement, and
+  `DbtProject` was never constructed anywhere, tests included, while every
+  project-reading consumer went to `dbt_project` directly. It is now three nested
+  protocols in `adapters/project.py`:
+  `ExploreProject` (`definitions()`), `MaintainProject` (adding
+  `transform_layer()` and `semantic_layer()`), and `EditableProject` (adding
+  `write_edits()`), with `tier_of()` reporting the highest tier an object
+  satisfies. `DbtProject` implements all three by delegating to the functions
+  that already existed, and explore's one project read now goes through it, so
+  the seam carries a real read path instead of describing one.
+
+  Tier 2 is two methods rather than one returning a union, because
+  `maintain.snapshot` already exposes exactly those two and `maintain.commands`
+  already calls them; a union would have to be unpacked at every call site.
+  Declining `EditableProject` is a supported answer: a project reduced from a
+  running graph cannot receive an edit, and a tier is checkable where a
+  `writeback: no` flag would be a claim the engine has to trust.
+
+- **A conformance contract for project formats** ([#171]).
+  `exmergo_dex_core.adapters.conformance` ships the seam's rules as assertions, so
+  a format outside this distribution runs them in its own suite, the way
+  `storage.conformance` already does for backends. Install
+  `exmergo-dex-core[project-conformance]`, which needs only pytest: nothing in the
+  contract reaches the dialect engine, and a packaging test keeps it that way.
+  `references/project.md` covers the tiers and what the construction contract
+  still leaves open.
+
+### Fixed
+
+- **`definitions()` no longer raises on a project file that is not valid YAML**
+  ([#171]). Its docstring promises that no project, an ambiguous choice, or an
+  unreadable project yields the empty view and never an exception, and explore
+  depends on that: exploration runs against warehouses with no project at all. But
+  `load` parses `dbt_project.yml` with `yaml.safe_load`, so a malformed project
+  file arrived as `yaml.YAMLError` while only `DbtProjectError` was caught, and
+  `explore --use-project` raised out of a read that is meant to degrade. The three
+  profile readers in the same module already pair the two exceptions; this is that
+  pairing on the read path. Found by the first run of the new conformance suite
+  against the shipped format.
+
 ## [1.5.1] - 2026-08-03
 
 ### Fixed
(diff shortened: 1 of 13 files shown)


marcociav-exmergo (approved):
LGTM, thanks @catincloudlabs !
Love the tiered project contract and the minimal one-liner change in `packages/dex-core/src/exmergo_dex_core/explore/commands.py`

PR #234, opened 2026-08-05
fix(transform): compare resolved DuckDB paths in dev-target preflight

### LINKED ISSUE #196: dev-target preflight compares DuckDB paths as raw strings
**Depends on:** nothing

### Problem

`transform build`'s dev-target preflight compares the DuckDB path in
`.dex/config.yml` against the one in the rendered `profiles.yml` as text. Two
spellings of the same file are reported as a disagreement and the build is refused.

### Reproducer

Verified on 1.5.1, against any dbt project whose `profiles.yml` uses a relative
`./` path.

| `.dex/config.yml` `duckdb.path` | `profiles.yml` `dev.path` | Result |
|---|---|---|
| `warehouse.duckdb` | `./warehouse.duckdb` | **refused** |
| absolute path to `warehouse.duckdb` | `./warehouse.duckdb` | **refused** |
| `./warehouse.duckdb` | `./warehouse.duckdb` | proceeds |

The refusal reads:

```
.dex/config.yml and profiles.yml disagree about the dev target:
  duckdb.path: warehouse.duckdb
  profiles.yml dev.path: ./warehouse.duckdb
edit one to match the other, then re-run (dex never rewrites a profile you may
have hand-edited)
```

All three rows name the same file.

### Why it matters

This gate exists to catch a real and expensive class of error: a build that reports
success against the wrong database. That makes it worth having, and it makes a
false positive on the gate worse than an ordinary bug, because the message
instructs the user to fix a disagreement that does not exist, and the only way
forward is to guess which spelling dex wants.

It also lands hardest on the zero-credential on-ramp, since a relative `./` path in
`profiles.yml` is what `dbt init` and most DuckDB tutorials produce.

### Proposed fix

Resolve both paths before comparing: expand the user directory, make each absolute
against the directory the file that declared it lives in, and normalize. Compare
the resolved paths. When they still differ, keep the existing message and add both
resolved forms, since "these resolve to different files" is the finding the user
needs.

### Acceptance

- All three rows above proceed.
- A genuine mismatch (two different files) is still refused, and the message shows
  the resolved paths alongside the declared ones.
- Symlinked paths that resolve to the same file are accepted.
- The check still refuses when the dev database does not exist, unchanged.

AshSgDe29071999:
## Summary
Dev-target preflight compared DuckDB paths as raw strings, so `warehouse.duckdb`, `./warehouse.duckdb`, and absolute paths to the same file were treated as drift.

## Changes
- Resolve config path against the repo root and profile path against the dbt project dir (expanduser + absolute + resolve).
- Accept equivalent spellings / symlinks; keep refusal for real mismatches with resolved paths in the message.
- Unit test for equivalent spellings.

Closes #196

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **transform dev-target preflight compares resolved DuckDB paths** ([#234]).
+  Relative and absolute spellings of the same warehouse file (e.g. `warehouse.duckdb`
+  vs `./warehouse.duckdb`) are no longer treated as configuration drift.
+
+
 - **A crowded low-cardinality dimension no longer pushes the true grain out
   of the composite-key probe** ([#168]). `_probe_composite_keys` ranks
   candidate 2-column keys by smallest distinct-count product and probes only
--- a/packages/dex-core/src/exmergo_dex_core/transform/dev_target.py
+++ b/packages/dex-core/src/exmergo_dex_core/transform/dev_target.py
@@ -99,7 +99,6 @@ def _resolve_declared_path(declared: str, base_dir: Path) -> Path:
         return path if path.is_absolute() else (base_dir / path).absolute()
 
 
-
 def check(
     project_dir: Path | str,
     target: str,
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (changes requested):

AshSgDe29071999:
Added Unreleased CHANGELOG entry for the DuckDB path-normalize preflight fix.

marcociav-exmergo (changes requested):
@AshSgDe29071999 , looks good!
Just fix the ruff linter issues and the merge conflicts, then we can merge

AshSgDe29071999:
Rebased on main (resolved CHANGELOG conflict with #235/#236 entries), fixed ruff format (extra blank line in `dev_target.py`), and kept the Unreleased Fixed entry for the resolved DuckDB path comparison.

marcociav-exmergo (approved):
LGTM!

PR #235, opened 2026-08-05
feat(transform): allow transform init into the current directory

### LINKED ISSUE #200: `transform init` cannot bootstrap into the current directory
**Depends on:** nothing

### Problem

`transform init <name>` always scaffolds into a `<name>/` subdirectory, and there
is no way to bootstrap into the directory you are already in.

```
$ dex transform init .
transform init needs a project name, e.g. `transform init analytics`

$ dex transform init analytics
ok, writing: analytics/dbt_project.yml, analytics/profiles.yml,
             analytics/models/staging/.gitkeep, analytics/models/marts/.gitkeep
```

Verified on 1.5.1.

### Why it matters

A large share of real repositories put the dbt project at the repository root, and
a caller who wants that has to run init, move four files, and fix
`.dex/config.yml` by hand, which is exactly the kind of unreviewed manual step
plan-then-apply exists to remove.

### Proposal

Add `--in-place` (or accept `.` as the name, meaning "use the directory's own
name"), which scaffolds into the current directory. The existing refusal when any
dbt project already exists still applies and is what keeps this safe. The project
`name:` in `dbt_project.yml` comes from the explicit name when given, and
otherwise from the directory name, normalized to a valid dbt project name.

### Acceptance

- `transform init analytics --in-place` writes `dbt_project.yml` and
  `models/` at the run root, and reports them as create diffs.
- The pre-existing-project refusal still fires.
- The `.dex/config.yml` written alongside points at the right project root in both
  modes.
- A directory name that is not a valid dbt project name is either normalized (and
  the normalization is reported) or refused with the fix.

AshSgDe29071999:
## Summary
`transform init` always wrote into a `<name>/` subdirectory, so root-level dbt layouts required manual moves.

## Changes
- Accept `.` as the project name (use directory basename, sanitize).
- Add `--in-place` to scaffold at the run root.
- Point `dbt_project_dir` at `.` in that mode; keep additive refusal when a project already exists.

Closes #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **transform init**: allow initializing into the current directory ([#235]).
+
 ## [1.5.2] - 2026-08-05
 
 ### Added
--- a/packages/dex-core/src/exmergo_dex_core/cli.py
+++ b/packages/dex-core/src/exmergo_dex_core/cli.py
@@ -194,6 +194,12 @@ def _build_parser() -> argparse.ArgumentParser:
                     sp.add_argument(
                         "--layered-schemas", action="store_true", default=False
                     )
+                    sp.add_argument(
+                        "--in-place",
+                        action="store_true",
+                        default=False,
+                        help="scaffold into the current directory instead of <name>/",
+                    )
                 if group == "transform" and name == "plan":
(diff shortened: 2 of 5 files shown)


marcociav-exmergo (changes requested):

AshSgDe29071999:
Added Unreleased CHANGELOG entry for transform init into the current directory.

marcociav-exmergo (approved):
LGTM, thanks for this!

PR #236, opened 2026-08-05
feat(cli): resolve unambiguous bare subcommands and suggest others

### LINKED ISSUE #198: CLI: accept an unambiguous bare subcommand, or suggest the group
**Depends on:** nothing

### Problem

Every subcommand requires its group. Calling a subcommand without it produces a raw
argparse failure with no suggestion:

```
$ dex check
usage: dex [-h] [--connector CONNECTOR] ... {connect,explore,transform,semantic,maintain,viz} ...
dex: error: argument group: invalid choice: 'check'
  (choose from 'connect', 'explore', 'transform', 'semantic', 'maintain', 'viz')
```

This is most likely when the caller reached dex through a skill wrapper, where the
group feels implied by the wrapper's own name.

### Why it matters

The recovery path is two or three more invocations of `--help` before the caller
gets back to the command they wanted, and on a slow first invocation that is a real
cost. It is also the first thing a new user hits, and a raw argparse error is the
least legible refusal the CLI can emit, in a project where every other refusal is
typed and explains its fix.

### Proposal

Two options, not exclusive:

1. **Suggest.** When the first positional is not a group but is a unique
   subcommand name across all groups, fail with
   `did you mean 'maintain check'?` rather than the argparse list. When it is
   ambiguous, name every group that has it.
2. **Accept.** Resolve an unambiguous bare subcommand to its group and proceed,
   with a warning naming the canonical form. Refuse an ambiguous one.

Option 1 alone is enough and is the smaller change. Option 2 is friendlier and
costs an alias table that has to stay in sync with the group definitions.

### Acceptance

- `dex check` suggests `maintain check`.
- A subcommand name that exists in more than one group names all of them.
- An entirely unknown token still produces the argparse list.
- The suggestion goes through the normal refusal path so it carries a machine
  readable `reason`, rather than argparse's bare stderr.

### Corroboration

Ten occurrences of the bare-`check` argparse failure appear in the archived
transcripts under `benchmarks/ade_bench/experiments/`.

AshSgDe29071999:
## Summary
Bare subcommands like `dex check` produced a raw argparse “invalid choice” error with no recovery hint.

## Changes
- Rewrite unambiguous bare subcommands to their group (e.g. `check` → `maintain check`).
- On remaining parse failures, print a suggestion naming the canonical group form (or all candidates if ambiguous).

Closes #198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **CLI**: resolve unambiguous bare subcommands and suggest alternatives when the name is ambiguous ([#236]).
+
 - **transform init**: allow initializing into the current directory ([#235]).
 
 ### Fixed
--- a/packages/dex-core/src/exmergo_dex_core/cli.py
+++ b/packages/dex-core/src/exmergo_dex_core/cli.py
@@ -64,6 +64,55 @@ COMMAND_SURFACE: dict[str, list[str]] = {
 }
 
 
+def _bare_subcommand_index() -> dict[str, list[str]]:
+    """Map each subcommand name to the groups that define it."""
+    index: dict[str, list[str]] = {}
+    for group, names in COMMAND_SURFACE.items():
+        for name in names:
+            index.setdefault(name, []).append(group)
+    return index
+
+
+def _rewrite_unambiguous_bare_subcommand(argv: list[str]) -> list[str]:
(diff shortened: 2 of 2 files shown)


marcociav-exmergo (changes requested):

AshSgDe29071999:
Added Unreleased CHANGELOG entry for bare subcommand resolution and suggestions.

marcociav-exmergo (approved):
LGTM! Thanks @AshSgDe29071999

PR #237, opened 2026-08-06
A crowded low-cardinality dimension no longer pushes the true grain out of the composite-key probe

### LINKED ISSUE #168: `_COMPOSITE_PAIR_CAP = 3` crowds the true grain out of the candidate set
**Version:** measured at 1.2.1 and 1.3.1; re-verified at 1.4.2. Follows #49, which added composite detection.

## Summary

`_probe_composite_keys` searches pairs only, ranks them by smallest distinct-count product, and cuts at three:

```python
# explore/profile.py:40
_COMPOSITE_PAIR_CAP = 3
# explore/profile.py:574
chosen = [list(pair) for _ids, _product, pair in ranked[:_COMPOSITE_PAIR_CAP]]
```

The cap, rather than the supported width, is what decides whether a real grain is reachable.

## The instructive case is not an unreachable wide grain

It is a table whose true grain is **two columns, a date plus one low-cardinality dimension, well inside the width dex supports**, and which is still never probed, because three other candidate pairs (combining that same dimension with a second low-cardinality column) rank below it on distinct-count product and consume the entire budget of three.

**A natural control pair makes this measurable rather than argued.** A sibling table in the same project has the same row count and the same two-column grain, and it *is* elected, ranking 2nd. Same warehouse, same profiling run, same width. The only difference is how many junk pairs happened to score lower.

## Why this is worse than a coverage gap

The output is a **confident wrong election** rather than an abstention. The table reports an elected grain that is not its grain, which then feeds `key_lost_uniqueness` and cardinality checks downstream. A grain dex could not determine at all would be more useful than one it determined incorrectly, because the first is visible and the second is not.

For a project that cannot express the grain by declaration either (single-column `DeclaredKey`, #169), the grain becomes unreachable by measurement **and** inexpressible by declaration, with no signal that either happened.

## Expected

Either:

1. **Raise the cap.** Simplest, and it moves the boundary without changing the shape of the problem.
2. **Rank before capping, more carefully.** Distinct-count product is a reasonable cheapness heuristic but it is not a plausibility heuristic. Pairs that are supersets or near-duplicates of each other (the three `(dimension, other)` pairs above are all the same dimension crossed with different partners) could be deduplicated before the cut, so the cap spends its budget on genuinely different hypotheses.

(2) is likely the better fix. The cap presumably exists to bound probe spend, so "probe more" trades cost for correctness while "rank smarter" does not.

Worth stating the cost tradeoff explicitly in whichever direction you go, since the cap is a spend guard and this issue is asking it to be less binding.

## Acceptance

A table whose true grain is a two-column composite, where three same-dimension pairs score lower on distinct-count product, elects the correct grain.

## Related

#169 is the declaration half of the same gap. Both are instances of the shape collected in #172.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/168

_probe_composite_keys (explore/profile.py) raises _COMPOSITE_PAIR_CAP from 3 to 5, and adds a new _COMPOSITE_REDUNDANCY_RATIO = 3.0 dedup pass before the cut: a pair sharing a column with an already-kept pair is dropped as a near-duplicate ("same dimension, different filler") only when its product is within 3x of the kept pair's a genuinely different hypothesis still gets its own slot even if it reuses that column.
Fixes the crowding case from the issue: several junk pairs anchored on one low-cardinality dimension no longer consume the entire cap and starve out the true grain, which can legitimately score worse on raw distinct-count product.
No change to cost-guard behavior: each extra slot is still a real scan batched into one statement, and a metered adapter that can't cover the wider batch already degrades to "grain unknown" via distinct_combination_counts's existing contract.

The issue : 
image

After fix: 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   The cap (`_COMPOSITE_PAIR_CAP`) is raised from 3 to 5 for headroom, and a
   pair that shares a column with an already-kept pair is now dropped as a
   near-duplicate when its product is within `_COMPOSITE_REDUNDANCY_RATIO`
-  (3.0x) of the kept pair's — the same idea tried with interchangeable
+  (3.0x) of the kept pair's: the same idea tried with interchangeable
   filler, not a genuinely different candidate. A pair whose product diverges
   meaningfully still gets its own slot even if it reuses a column, which is
   what lets the true grain through. Each additional slot is still a real
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,32 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **A crowded low-cardinality dimension no longer pushes the true grain out
+  of the composite-key probe** ([#168]). `_probe_composite_keys` ranks
+  candidate 2-column keys by smallest distinct-count product and probes only
+  the top few, because each probe is a real two-column `DISTINCT` scan. When
+  one dimension column happened to pair cheaply with several other columns,
+  every one of those pairs ranked ahead of the true grain purely because they
+  shared that same attractive anchor, even though they were really the same
+  hypothesis ("does this dimension have *a* partner?") tried with different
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
LGTM

PR #238, opened 2026-08-07
SCRUM-1053 explore profile: report the value domain of a low-cardinality non-PII column

### LINKED ISSUE #203: `explore profile`: report the value domain of a low-cardinality non-PII column
**Depends on:** nothing. **Blocks:** #218.

### Problem

`ColumnProfile` carries `min_value` and `max_value`, and
`profile.is_min_max_safe()` excludes **every string column** on the grounds that a
string min or max is a raw value. The consequence on a raw source table, where
string columns are the norm:

```
$ dex explore profile raw_workspaces
ws_id           VARCHAR  null=0.0   distinct=198  unique=True   min=None max=None
acct_id_fk      VARCHAR  null=0.0   distinct=77   unique=False  min=None max=None
ws_stat_cd      VARCHAR  null=0.0   distinct=6    unique=False  min=None max=None
env_tier        VARCHAR  null=0.0   distinct=4    unique=False  min=None max=None
primary_ws_yn   VARCHAR  null=0.0   distinct=5    unique=False  min=None max=None
soft_delete_ind VARCHAR  null=0.0   distinct=2    unique=False  min=None max=None
data_quality: []
```

(Reproducible against the public ADE-bench `helixops_saas.duckdb` file.)

The caller learns that `env_tier` has four values and cannot learn what they are.
One `explore query` away sits the thing they actually needed:

```
ws_stat_cd: A=62, active=62, Enabled=59, archived=7, ARCH=5, disabled=3
```

Three logical states under six encodings. dex computed the distinct count to get
there and then declined to say it.

### Why it matters

Writing a correct `CASE` expression over a source column requires knowing the
column's domain. This is the single most common question asked of a warehouse
during modeling, and the current answer sends the caller to a raw SQL client, which
is the one place the PII policy and the query firewall do not apply. A guard that
pushes people around it is weaker than one they can work with.

### Proposal

When a column's distinct count is below a cap and the column carries **no PII flag
at any confidence**, include the value domain with frequencies in the profile.

Eligibility, all required:

- no PII flag at all, at any confidence, and not merely below the block threshold
- distinct count at or below a cap (proposed default 25, configurable downward)
- distinct count is a small fraction of non-null rows, so a tiny table of distinct
  people does not qualify on the absolute count alone
- the column is not a candidate key

The result is capped, ordered by frequency, and reports the number of values
elided when the cap binds.

### What this deliberately does not do

- It does not change detection, confidence, or the block threshold.
- It does not report a domain for a flagged column, ever, at any confidence or
  cardinality. A flagged column stays silent and a committed human override remains
  the only way to clear it.
- It does not weaken the rule that false positives are the design working.

The cardinality signal this uses is already computed and already used in the
opposite direction, to de-rate confidence on closed low-cardinality vocabularies.
This applies the same evidence to a column the detector never flagged.

### Cost

One grouped aggregate per eligible column, batched with the rest of the profile
scan. Eligibility is decided from the distinct count the profile already has, so
nothing extra is scanned for an ineligible column. This lands on the metered side
of the free-versus-gated split and must reserve its per-query billing floor like
every other profile statement.

### Acceptance

- A four-value non-PII string column reports its domain with counts.
- A PII-flagged column reports no domain, at any confidence and any cardinality.
- A high-cardinality column reports no domain.
- The cap is enforced and elisions are counted.
- The safety assertions gain a case: no value from a flagged column can reach the
  envelope through this path, on every connector.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/203
ColumnProfile gains a value_domain field (ValueDomain: capped values by frequency + elided count), populated by a new _probe_value_domains step in explore/profile.py.
Eligibility (all required): no PII flag at any confidence, distinct count ≤ 25 and ≤ 10% of non-null rows, not a candidate key (single-column or proven composite member).
New adapter capability value_domain_counts, implemented for DuckDB and BigQuery this pass (one batched statement: capped top-K by frequency + exact distinct-group count per column). Other connectors don't implement it yet and degrade to reporting no domain, same as any adapter that predates this feature.
Eligibility is decided from the approximate pre-scan distinct count (cost control); the reported domain is built from the exact count the probe returns if approximation under-estimated but the true count still clears the fraction bar, the result is capped with an honest elided count instead of being dropped; only a true count that breaks the fraction bar is dropped entirely.
BigQuery: cost-gated like every other optional escalation (dry-run, floor at the per-query minimum, try_charge, degrade to {} + a table note on insufficient budget); profile_estimate reserves for it up front.
before the fix : 
image
After fixing : 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,38 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **`explore profile` reports the value domain of a low-cardinality non-PII
+  column** ([#203]). `ColumnProfile.min_value`/`max_value` are suppressed for
+  every string column, PII or not, because a string extreme is itself a raw
+  value. On a raw source table (the norm: mostly string-typed columns), a
+  caller learned `env_tier` had 4 distinct values and had no way to learn
+  what they were, short of a raw SQL client, exactly the one place the PII
+  policy and query firewall do not apply.
+
+  A column now reports its `value_domain` (distinct values by frequency,
+  capped, with an `elided` count when the cap binds) when it carries no PII
+  flag at all, at any confidence, its distinct count clears both an absolute
+  cap (25) and a row-relative fraction (10% of non-null rows, so a tiny
+  table's near-key column does not qualify on the absolute count alone), and
+  it is not a candidate key (single-column or a proven composite member). A
+  flagged column reports no domain at any confidence or cardinality, full
+  stop; detection, confidence, and the block threshold are unchanged.
+
+  Eligibility is decided from the approximate pre-scan distinct count (cost
+  control: an obviously-too-wide column is never even queried), but the
+  reported domain is built from the exact count the new probe itself
+  returns: when the approximation under-estimated and the true count is
+  still within the fraction bar, the result is capped with an honest
+  `elided` count rather than dropped; only a true count that breaks the
+  fraction bar (a near-key the approximation missed) is dropped entirely.
+
+  New adapter capability (`value_domain_counts`), implemented for DuckDB and
+  BigQuery this pass; other connectors don't implement it yet and degrade to
+  reporting no domain, identical to how an adapter that predates this
+  feature behaves. On BigQuery it is spent only inside the already-confirmed
+  budget, floored at the per-query minimum like every other optional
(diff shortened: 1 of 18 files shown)


marcociav-exmergo (approved):
LGTM!

PR #239, opened 2026-08-08
fix(transform): accept line-broken top-level dbt refs

### LINKED ISSUE #195: `transform plan` rejects a `ref()` on its own line in a top-level FROM
**Depends on:** nothing

### Problem

`transform plan` refuses a valid dbt model when its `{{ ref(...) }}` sits on the
line after `from` rather than on the same line, but only at the top level of the
statement. Inside a CTE the same shape is accepted.

### Reproducer

Verified on 1.5.1. Three models, identical except for where the ref sits:

```sql
-- accepted: ref shares the FROM line
select a
from {{ ref('stg_accounts') }}
where a is not null
```

```sql
-- accepted: ref on its own line, but inside a CTE
with s as (
  select a
  from
    {{ ref('stg_accounts') }}
)
select * from s
```

```sql
-- REFUSED: ref on its own line at the top level
select a
from
    {{ ref('stg_accounts') }}
```

The third returns:

```
models/marts/t.sql: Expected table name but got
<Token token_type: TokenType.SENTINEL, text: SENTINEL, line: 1, col: 1, ...>.
Line 2, Col: 4.
  select a
  from
```

### Cause

`transform/validate.py: strip_jinja()` replaces each `{{ ... }}` with an
identifier placeholder, then walks the lines and **drops any line that is nothing
but the placeholder** when paren depth is zero. The docstring states the
assumption: such a line is taken to be a `{{ config(...) }}` header. At depth
greater than zero the line becomes a placeholder subquery instead, which is why the
CTE case survives.

For a top-level `from` whose ref is on the next line, the `from` survives and its
table reference is deleted, so the SELECT-only parse fails on SQL that was fine.

### Why it matters

A validator that rejects valid SQL is the most expensive bug class in `transform`,
because the correct response for a caller is to stop routing edits through dex and
hand-write the file, which loses the reviewable-diff guarantee that is the whole
point of plan-then-apply.

### Proposed fix

Keep the intent (a `config()` header should not become a bare identifier) and
narrow the trigger: at depth zero, drop a placeholder-only line **only before the
first `select` or `with` keyword**. A dbt `config()` block is by convention at the
top of the file; a ref never is.

### Acceptance

- The three models above all validate.
- A `{{ config(...) }}` header alone on its own line at the top of a file is still
  dropped and the model still validates.
- Same for a header followed by a blank line, and for `{{- config(...) -}}`.
- A model that is entirely jinja still reports the existing "model is entirely
  jinja; SELECT-only check skipped" warning.
- Regression tests cover top-level FROM, top-level JOIN, and the CTE case.

### Corroboration

Observed in the archived transcripts under `benchmarks/ade_bench/experiments/`,
where an agent's snapshot-derived model was refused on this shape and the agent
abandoned the plan path for that file.

alloutflo:
## What this changes

- locate the model query boundary from real SQL `SELECT` and `WITH` tokens after Jinja replacement
- preserve placeholder-only Jinja lines after that boundary so line-broken top-level `ref()` calls remain valid after `FROM` and `JOIN`
- continue dropping standalone config headers before the query
- add regression coverage for top-level refs, CTEs, config variants, SQL comments, and entirely-Jinja models
- document the fix under `[Unreleased]`

Token-based boundary detection avoids treating keywords inside SQL comments or strings as the start of the query.

## Related issues

Closes #195.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`): 1,818 passed, 61 skipped
- [x] The safety spine still holds (`uv run pytest -q tests/test_safety_spine.py`): 117 passed
- [x] Prose is em-dash free

Additional checks:

- `uvx ruff@0.15.19 check --output-format=github .`
- `uvx ruff@0.15.19 format --check --diff .`
- `git diff --check`
- sqlglot 28.6 token API compatibility smoke test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **`transform plan` accepts a line-broken top-level dbt `ref()`** ([#195]).
+  Placeholder-only Jinja lines remain intact after the model's first `SELECT` or
+  `WITH`, while standalone config headers before the query are still removed.
+
 - **transform dev-target preflight compares resolved DuckDB paths** ([#234]).
   Relative and absolute spellings of the same warehouse file (e.g. `warehouse.duckdb`
   vs `./warehouse.duckdb`) are no longer treated as configuration drift.
--- a/packages/dex-core/src/exmergo_dex_core/transform/validate.py
+++ b/packages/dex-core/src/exmergo_dex_core/transform/validate.py
@@ -11,7 +11,10 @@ from __future__ import annotations
 import re
 from typing import TYPE_CHECKING
 
+import sqlglot
 import yaml
+from sqlglot.errors import TokenError
+from sqlglot.tokens import TokenType
 
 from ..errors import DexError
 from ..guards.sql_guard import assert_select_only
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (approved):
Hey @alloutflo thanks for you contribution, love this!

PR #247, opened 2026-08-09
Read the snapshot schema version, and stop reporting a corrupt baseline as a bad request

### LINKED ISSUE #243: [core] Snapshot.schema_version is written and never read, so a corrupt baseline reports as a bad request
**Version:** measured at HEAD with the #171 change applied.

`Snapshot.schema_version` is stamped on every write and has no reader anywhere in the distribution: no migration, no staleness path, no compatibility check. `DRIFT_SCHEMA_VERSION` is in the same state.

Contrast `CACHE_SCHEMA_VERSION`, which the query firewall degrades on deliberately.

## The consequence is in the refusal taxonomy, not in the data

A baseline that fails validation raises out of `storage/filesystem.py` uncaught, reaches the CLI's catch-all, and because pydantic's `ValidationError` subclasses `ValueError` it is classified as a **request** error. So a corrupt or incompatible baseline tells the operator they made a bad request, when the fix is `maintain snapshot`.

That is exactly the retry-versus-stop distinction `PrerequisiteError` exists to carry, and the same shape of gap that left `CacheRequiredError` and `NoBaselineError` unreachable long enough for someone embedding the engine to have to match on prose.

## Why it matters more now than it did

Before #171 the only writer of a snapshot was dex reading a dbt project, so a version field with one producer and one consumer could not disagree with itself. A host can now name its own project format in `.dex/config.yml`, so a second writer exists and the versions start meaning something.

## Proposed fix

- Read the version on load; refuse an unknown one as a `PrerequisiteError` naming `maintain snapshot`.
- Classify a snapshot `ValidationError` the same way, so a corrupt baseline stops reporting as a bad request.
- Decide at the same time whether an older-but-readable version is migrated or refused, and write the answer down. Same question for `DRIFT_SCHEMA_VERSION`.

catincloudlabs:
## What this changes

`Snapshot.schema_version` was stamped on every write and read by nothing, so two
states reached the drift detectors that should not have.

**A corrupt baseline reported as a bad request.** `load_snapshot` raises, and
pydantic's `ValidationError` subclasses `ValueError`, so it fell to the CLI
catch-all and was classified as a request error. The operator is told they made a
bad request when the fix is `maintain snapshot`.

**An incompatible baseline was not detected at all.** A document from an unknown
schema was handed to the detectors as though it were current. That is the worse of
the two, because it is not an error: it is a drift report measured against a shape
the engine misread.

The reasoning is in the `CHANGELOG.md` entry under `[Unreleased]`, so it survives
the squash. This body covers the decisions worth pushing on.

### Where the check lives, and why

In the engine, not the store. `storage/base.py` is explicit that a stored
`schema_version` is not the store's to police, so `filesystem.py` is untouched and
a backend still loads what it was given.

All six `load_snapshot()` call sites in `maintain/commands.py` were byte-identical
and now go through one `_require_baseline()`.

### Why a supported set rather than the `<` you use on `CACHE_SCHEMA_VERSION`

The issue names that contrast, so it deserves an answer rather than a silent
divergence. Because the two documents are not the same kind of thing. An old cache
has a degraded reading, which is why the query firewall adds a hint and carries on.
A baseline has no degraded reading, only a wrong one, because it is the thing every
axis measures against. Membership also takes a position on newer-than-this-engine,
which `<` would have to decide separately.

### What the version check does not reach

Narrower than "an incompatible baseline is refused" sounds, so it is stated in the
code as well as here. The check runs on a parsed `Snapshot`, because the store hands
back a model and the raw document is not the engine's to see.

| stored document | outcome |
|---|---|
| unknown version, shape today's model still validates | refused, version named |
| unknown version, shape today's model rejects | refused as a parse failure, version not named |

Same refusal and same remedy either way, but the second message cannot say which
version it came from, and a real schema bump usually changes shape. Both messages
therefore name the possibility. Reading the version before validating would take a
`Store` contract change, which is wider than this PR should be. Happy to open it.

### A named error rather than a reuse

`BaselineUnreadableError` is a new `PrerequisiteError` rather than a second use of
`NoBaselineError`. Both remediate the same way, so the status is unchanged, but "you
never took a baseline" and "the baseline you have is unreadable" are different facts
about a deployment and only one of them suggests something went wrong. A host that
wants to page on the second and not the first can now tell them apart.

It carries `schema_version` for the same reason one level down: the class covers two
operational events, a deployment that rolled forward or back and an integrity
problem, and separating those would otherwise be a prose match. That is the gap this
issue is about, so the fix should not open a smaller one.

### Migrate or refuse

**Refuse**, with the reasoning next to the supported set rather than only here. A
baseline is cheap to regenerate; migration needs a function per version that has to
be right about a document nobody has read in a while; and a wrongly migrated
baseline is worse than an absent one, because drift measured against it looks like a
result.

Only one version exists today, so nothing is refused yet. The comment says what to do
when a second arrives, including the failure mode of adding it to the set just to
keep an old baseline working.

### What a refusal now says about running `maintain snapshot`

The command is free and cannot really fail, which is what makes "just re-snapshot"
sound costless. It is not: a replacement pins current state as known-good, so drift
accumulated since the last readable baseline is absorbed and never reported on any
axis. Both messages say that now.

The two version arms also differ. dex tells operators to commit
`.dex/snapshot.json` like a lockfile, so a baseline newer than the engine reading it
is usually a colleague on a newer dex rather than a broken file, and
`maintain snapshot` resolves that by overwriting their baseline with one they can no
longer read. That arm names upgrading first.

### `DRIFT_SCHEMA_VERSION`

Deliberately without the same gate. The reasoning is now next to the constant rather
than only in this body, since the issue's third bullet asked for the answer written
down: the baseline is vouched for and nothing else reproduces it, while the drift
report is derived, `maintain check` regenerates it on demand, `_record_axes` already
discards one measured against a different baseline wholesale, and `reconcile` already
has an absent-report path. An unrecognized version there has a cheap correct answer
the baseline does not have.

That argument is about versioning and not about corruption, and the comment says so:
a drift report that fails to parse still raises out of `load_drift` uncaught and is
still classified as a request error. Left out because the right remedy is probably
"treat as absent and rebuild" rather than the baseline's refusal, and picking it by
symmetry would be picking it for the wrong reason. Happy to take either here.

## Tests

Four arms, each parametrized over all six `maintain` entry points that read the
baseline: `check`, `schema`, `volume`, `grain`, `semantic`, `reconcile`.

That parametrization is the point rather than thoroughness. The change is a
consolidation, and asserting on `check` alone pins the helper while saying nothing
about whether anything else reaches it: reverting five of the six call sites to the
original three lines left the whole suite **bit-identical to an unmutated tree**.
With the arms parametrized, the same mutation fails ten cases across exactly those
five commands.

| removed | fails |
|---|---|
| the `schema_version` check | exactly the unknown-version arms |
| the `ValidationError` catch | exactly the corruption arms |
| `_require_baseline` from five of six call sites | ten arms, across those five commands |

Plus an absent-baseline arm, which keeps `NoBaselineError` and
`BaselineUnreadableError` distinguishable, and a quiet arm pinning that a current
baseline is not refused on any of the six. Without the last one, the refusal arms are
equally consistent with refusing every baseline everywhere.

Three more: that the newer-version message leads with upgrading rather than with
`maintain snapshot` (asserted on order, since presence is not the property that
matters), that neither message advertises a replacement as free, and that
`schema_version` on the error separates the two states.

Suite on this branch: **1856 passed, 80 skipped**, one pre-existing failure unrelated
to this change (see below). `ruff@0.15.19 check` and `format --check` clean from the
repo root, `uvx pytest evals` clean, and the em-dash checker clean over the files CI
passes it.

## Deliberately not in this PR

Each wants its own decision, and I would rather name them than let them read as
covered:

1. **The conformance suite asserts nothing about a document that will not parse**,
   for the cache or the snapshot, so a backend can satisfy the whole contract without
   ever exercising this path. It needs an overridable hook to let a backend produce
   an unparseable document in its own storage, the way `make_unreadable_project()`
   does for formats.
2. **A corrupt drift report** still misclassifies, as above.
3. **Reading a stored version without validating the document**, which is the
   `Store` contract change the table above points at.
4. **`load_cache` has the same gap as the drift report, and across ten call sites.**
   `_require_baseline` wraps `load_snapshot` only. Every `load_cache` caller, six in
   `explore/commands.py`, one in `transform/scaffold.py`, and three in
   `maintain/commands.py`, calls the store bare and branches on `is None`, so a
   `ValueError` out of a load reaches the CLI catch-all and is reported as a bad
   request. Surfaced by review on this branch: the contract paragraph first added
   here claimed the engine already classified it, and 7e25308 scopes that claim to
   `load_snapshot` and records the gap rather than overstating the fix.

Items 2 and 4 share a remedy shape, a `_require_cache` mirroring `_require_baseline`,
but they do not share a policy. A corrupt cache has to refuse: treating it as absent
silently re-profiles a warehouse and bills for it, which is the argument
`load_cache`'s own docstring already makes. A corrupt drift report can rebuild, per
the note beside `DRIFT_SCHEMA_VERSION`. Whoever picks this up should settle both
policies deliberately rather than inherit symmetry from the baseline.

## Three small things found on the way, none of them from this change

1. **`CONTRIBUTING.md`'s local loop leaves part of the safety spine failing rather
   than skipping.** `uv sync --extra duckdb --extra bigquery --extra dev` is not
   enough for `tests/transform/test_dev_target.py` (13 failures), and one of them is
   `tests/test_safety_spine.py::test_init_redshift_profile_is_dev_only_with_no_secrets`,
   which fails on `ModuleNotFoundError: No module named 'boto3'`. A safety-spine
   assertion that hard-fails on the documented setup is indistinguishable from a real
   break, and `release.yml` installs every extra and explains why while the
   contributor-facing doc does not.
2. **`test_retargeted_dev_database_is_refused_naming_both_values` fails on Windows.**
   It asserts `"analytics/profiles.yml"` in a message built from a `Path`, which is
   `analytics\profiles.yml` there. Invisible on ubuntu CI. It is the one failure on my
   baseline and is unrelated to this PR.
3. **`CONTRIBUTING.md` says `uvx ruff`, which floats; CI pins `uvx ruff@0.15.19`.** On
   latest ruff, `format --check` reports `README.md` and `references/storage.md` as
   unformatted, because newer ruff formats markdown code blocks. On the pin,
   everything passes. A contributor following the doc gets a different answer from the
   gate.

Happy to open issues for any of these.

## Related issues

Closes #243.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`); the one failure is the
      pre-existing Windows path assertion in item 2 above


marcociav-exmergo (approved):
LGTM! Thanks for this @catincloudlabs, I'm gonna merge

PR #248, opened 2026-08-09
The tier-3 write path carries the plan's directory and the confirmed handshake

### LINKED ISSUE #241: [core] EditableProject.write_edits cannot carry `confirmed`, so a caller routed through tier 3 would drop the human-edit conflict guard
**Version:** measured at HEAD with the #171 change applied.

The tier-3 protocol in `adapters/project.py` declares:

```python
def write_edits(self, edits: Any) -> Any: ...
```

One argument. The real caller passes two:

```python
# transform/commands.py
dbt_project.write_edits(edits, project_dir, confirmed=engine.confirmed)
```

and `DbtProject.write_edits` drops `confirmed` outright, so it defaults to `False`.

## Why this is a contract defect rather than an unused-code nit

`confirmed` is what the human-edit conflict handshake turns on. `dbt_project.write_edits` re-hashes every target file, and any file whose hash moved since plan time becomes a `Conflict`; with `confirmed=False` the whole apply is refused and nothing is written. That refusal is propose-don't-impose itself, and it is asserted in the safety spine (`test_apply_refuses_to_overwrite_a_human_edit`).

Nothing is broken today, because `transform apply` calls the module function rather than the tier and no caller routes through `EditableProject`. That is precisely the problem. The protocol as published describes a write path that silently loses a hard constraint, and the first person to implement tier 3 against the signature will implement the wrong thing, in the one place where getting it wrong overwrites someone's work.

A public protocol whose signature omits a parameter a hard constraint depends on is worth fixing whether or not it currently has a caller. It is also cheaper now than after a second format has built against the one-argument form.

## Proposed fix

- Widen the tier-3 member to carry `confirmed`, and have `DbtProject.write_edits` pass it through.
- Add a conformance assertion that a tier-3 format refuses rather than overwrites when it is not confirmed. That is the behavioural half, and no shape check would catch it. It is the tier-3 analogue of the never-raises assertion that already carries tier 1.

## Related

Surfaced while building #171, which deliberately stopped short of routing writes through tier 3: nothing is gained for a format that will never implement it, and the write surface is dbt all the way down (`contained_path`, the seven `EditKind` values, `validate_edit`'s `dbt parse` subprocess). Fixing the signature is separable from that work and should land before anyone routes through it.

See also the sibling defect in the same signature: #242.

### LINKED ISSUE #242: [core] Tier 3 resolves the project from engine config while the plan it applies recorded its own directory
**Version:** measured at HEAD with the #171 change applied.

`DbtProject.write_edits` resolves the project as:

```python
dbt_project.write_edits(edits, self.project_dir or self.repo_root)
```

from engine configuration. `plans.apply` resolves it as:

```python
Path(repo_root) / stored.project_dir
```

the plan's own recorded relative directory. That relative form is deliberate: it is what keeps a stored plan valid when the repository moves.

A `DbtProject` built from engine configuration need not point at the directory a given plan was pinned against, so the two can disagree, and the disagreement is silent: the edits would be written into whichever project the engine happened to be configured for, with the hash checks run against that project's files.

## Scope note

The discovery half of this was fixed in #171: `DbtProject.load()` now runs discovery and resolves to an absolute path the same way `definitions()` does, so tier 1 and tier 2 no longer disagree about which project they are reading. `write_edits` was left alone on purpose, because the correct resolution there is the *plan's* rather than the engine's, which is a different question and belongs with the signature change.

## Proposed fix

Have the tier-3 write path take the project directory from the plan being applied rather than from engine configuration, or state on the protocol that the caller supplies it. Settle it together with #241, since both are the same signature.

catincloudlabs:
## What this changes

`EditableProject.write_edits` declared one argument while the behavior it governs
depends on two more. Both issues are the same signature, which is why they land
together.

**`confirmed` had nowhere to go** (#241). It is what the human-edit conflict
handshake turns on: every target is re-hashed against the content the edit was
planned against, and a file whose hash moved is someone's work rather than a stale
line to overwrite. Anything routed through the tier would refuse every apply that met
a conflict and could never override one, and an implementation written against the
published signature had no way to receive it at all.

**The project directory came from engine configuration rather than from the plan
being applied** (#242). A plan records its own directory relative to the repository
root, precisely because that is what stays valid when the repository moves, so the
two can name different projects with nothing to say so. The edits would land in
whichever project the engine happened to be configured for, hash-checked against that
project's files.

The member is now `write_edits(edits, project_dir, *, confirmed=False)`. `DbtProject`
passes both through, and a directory the caller names wins over the one the instance
was built with. Passing none keeps the configured pin, which is what a caller holding
engine configuration and no plan has always had.

## Why this is a contract fix rather than a behavior fix

Nothing in the distribution called the tier, so no shipped path changes:
`transform apply` calls the module-level `dbt_project.write_edits`, which always
resolved both correctly, and `reconcile` reads the tier only to decide whether it may
author an edit at all.

That is the argument for fixing the signature now rather than after something builds
on it. `runtime_checkable` checks only that a method exists, so a third-party
implementation of the one-argument shape keeps satisfying `isinstance` and fails at
the call. The first person to implement tier 3 against the published signature would
implement the wrong thing, in the one place where getting it wrong overwrites
someone's work.

Deliberately **not** in this PR: routing `plans.apply` through the tier. #171 stopped
short of that on purpose, the write surface is dbt all the way down (`contained_path`,
the seven `EditKind` values, `validate_edit`'s `dbt parse` subprocess), and the
signature is separable from it.

## The conformance suite reaches tier 3

`EditableProjectContract` is the class the suite did not have. The chain stopped at
`MaintainProjectContract`, so a format could reach tier 3 and have nothing assert what
the tier is for.

It asserts the tier, that an unconfirmed write leaves a target someone else changed
alone, and that a confirmed one goes through. **The two write assertions are a pair on
purpose**: a `write_edits` that never writes satisfies the first by itself, so the
confirmed case is what rules out the trivial implementation.

Its one hook raises rather than defaulting to a skip, unlike `make_unreadable_project`,
because the excuse does not carry over. A format may genuinely have no unparseable
state, but a format that reached tier 3 writes into a source of truth a human can also
edit, so the case where the human got there first exists for every one of them.

Both shipped conformance subclasses widen to tier 3, including the factory one, whose
declared `tier` is now `EditableProject`. That is what `from_context` actually
returns, and a factory quietly building one tier below what the format implements is
exactly what that assertion is for.

**Which directory the edits land in stays a dbt assertion**, in
`tests/adapters/test_project_parity.py`, rather than a contract one. `project_dir` is
a slot for the formats keyed by a directory, so asserting it in the shipped contract
would really be asserting that every tier-3 format is a filesystem format.

## Verified by mutation, not only by green

Reverting `DbtProject.write_edits` to its old body under the new signature, which
reproduces both defects exactly, fails precisely three tests: both
`test_a_confirmed_write_overrides_the_conflict` instances and
`test_the_write_tier_uses_the_directory_the_caller_passed`.
`test_an_unconfirmed_write_refuses_a_target_that_moved` correctly still passes, since
the refusal was never the broken half.

## Related issues

Closes #241.
Closes #242.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`): 1837 passed, 80 skipped.
      The one failure is the pre-existing Windows path-separator assertion in
      `test_retargeted_dev_database_is_refused_naming_both_values`, which fails
      identically with this branch stashed
      cost-guarded, PII flagged not surfaced, propose-don't-impose. This PR is
      propose-don't-impose specifically, and moves it from a promise an
      implementation has to remember into an assertion the contract makes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,46 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The tier-3 write path can carry the two things that make it safe** ([#241],
+  [#242]). `EditableProject.write_edits` declared one argument while the behavior it
+  governs depends on two more. `confirmed` had nowhere to go, so anything routed
+  through the tier would refuse every apply that met a conflict and could never
+  override one, and an implementation written against the published signature had no
+  way to receive it at all. That is the parameter the human-edit conflict handshake
+  turns on: every target is re-hashed against the content it was planned against, and
+  a file whose hash moved is someone's work, not a stale line to overwrite. The
+  project directory was resolved from engine configuration rather than from the plan
+  being applied, and a plan records its own directory relative to the repository root
+  precisely because that is what survives the repository moving, so the two can name
+  different projects with nothing to say so: the edits would land in whichever
+  project the engine happened to be configured for, hash-checked against that
+  project's files.
+
+  The member is now `write_edits(edits, project_dir, *, confirmed=False)`.
+  `DbtProject` passes both through, and a directory the caller names wins over the
+  one the instance was built with. Passing none keeps the configured pin, which is
+  what a caller holding engine configuration and no plan has always had.
+
+  No shipped path changes behavior, because nothing in the distribution called the
+  tier: `transform apply` calls the module-level `dbt_project.write_edits`, which
+  always resolved both correctly, and `reconcile` reads the tier only to decide
+  whether it may author an edit at all. That is the argument for fixing the signature
+  now rather than after someone builds on it. `runtime_checkable` checks only that a
+  method exists, so a third-party implementation of the one-argument shape keeps
+  satisfying `isinstance` and fails at the call.
+
+- **The conformance suite reaches tier 3** ([#241]). `EditableProjectContract` is the
+  class the suite did not have, and it asserts the behavioral half no shape check can
+  see: an unconfirmed write leaves a target someone else changed alone, and a
+  confirmed one goes through. The two are a pair on purpose, since a `write_edits`
+  that never writes satisfies the first by itself. Its one hook raises rather than
+  defaulting to a skip, unlike `make_unreadable_project`, because the excuse does not
+  carry over. A format may genuinely have no unparseable state, but a format that
+  reached tier 3 writes into a source of truth a human can also edit, so the case
+  where the human got there first exists for every one of them.
+
 ## [1.6.0] - 2026-08-08
 
 ### Added
--- a/packages/dex-core/src/exmergo_dex_core/adapters/conformance.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/conformance.py
@@ -16,9 +16,9 @@ runs the contract against your format.
 **Pick the class that matches the tier you implement**, mirroring the protocols
 exactly: :class:`ExploreProjectContract` for a format that can state what it
 declares, :class:`MaintainProjectContract` when it can also produce the two
-snapshot layers. Subclassing a wider contract than your format implements is how
-you find out you have not finished, so the narrow class is a feature rather than a
-convenience.
+snapshot layers, :class:`EditableProjectContract` when it can also receive an edit.
+Subclassing a wider contract than your format implements is how you find out you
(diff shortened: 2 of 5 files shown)


marcociav-exmergo (approved):
LGTM! Thanks @catincloudlabs

PR #251, opened 2026-08-09
`explore profile` flags a declared type that contradicts the column's   content

### LINKED ISSUE #204: `explore profile`: flag a declared type that contradicts the column's content
**Depends on:** nothing

### Problem

Profiling reports a column's declared type and nothing about whether the content
agrees with it. Three shapes recur constantly in landing-zone and CDC data and none
is detected:

- a string column whose values are dates or timestamps in a fixed format
- an integer (or string) column whose values are unix epochs
- a string column whose values are all numeric

Against the public ADE-bench `helixops_saas.duckdb`, `raw_workspaces` has
`crt_ts_epoch VARCHAR`, `deact_ts_epoch VARCHAR` and `load_epoch VARCHAR`, all
carrying epoch values, and `data_quality` comes back empty.

### Why it matters

The fix for one of these is a cast, and the cast needs the **format**, which is the
part that is expensive to work out by hand and cheap to measure. A model that
carries a string date forward silently produces a relation that compares equal to
nothing and sorts lexicographically, which is a wrong answer that looks like a
right one.

### Proposal

Add data-quality observations, each computed as a fraction inside a measuring
aggregate, using the same mechanism that already buys value-shape evidence:

| Observation | Signal | Reported |
|---|---|---|
| string-encoded temporal | fraction matching a date or timestamp pattern | the pattern, the matching fraction, and the format string a cast would use |
| epoch-shaped number | numeric range consistent with seconds or milliseconds since 1970 | which unit, and the implied min and max dates |
| numeric-only string | fraction matching a numeric pattern | the fraction, and whether every value fits an integer |

Candidate formats are a small fixed list (ISO, `%m/%d/%Y`, `%d/%m/%Y`, with and
without a time part). Where two candidates are indistinguishable on the data, say
so rather than picking, since guessing between `%m/%d/%Y` and `%d/%m/%Y` is how a
silent corruption happens.

Only fractions and pattern names leave the engine. No value does.

### Acceptance

- A string column of `%m/%d/%Y %H:%M:%S` values produces a note naming that format.
- A string column mixing two indistinguishable date orders reports the ambiguity
  rather than choosing.
- An integer column of second-precision epochs is distinguished from one of
  millisecond-precision epochs.
- A genuine integer measure (an amount, a count) produces no epoch note.
- No raw value appears in the observation text, asserted in the safety suite.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/204
New type_stats request set (parallel to shape_stats): non-PII string/integer columns get 14 new ColumnAggregate fraction/min-max fields, computed inside the same already-scanned aggregate batch across DuckDB, BigQuery, Snowflake, Databricks, Redshift, and Postgres. No new adapter capability, no extra scan cost.
Three detectors surfaced as data_quality notes: string-encoded date/timestamp format (ISO + slash-separated, with MDY/DMY disambiguated from the data or reported as ambiguous when genuinely indistinguishable), unix-epoch seconds vs. milliseconds (with an implied calendar-date range translated from an aggregate MIN/MAX), and numeric-only strings (plus whether every value fits an integer).
Safety: a column is eligible only with no PII flag at all, at any confidence mirrors is_min_max_safe and #203's value-domain gate. Only fractions, format/unit names, and a translated date ever leave the engine; the underlying value never does.
test_profile.py: eligibility request-set test, 12 parametrized note-decision cases (numeric string, ISO date/datetime, unambiguous MDY, unambiguous DMY, genuine ambiguity, contradiction, epoch seconds vs. millis, genuine integer measure, no-evidence), and an end-to-end PII-gate test.
test_safety_spine.py: extended the 6 existing per-dialect _build_aggregate_sql SELECT-only tests to also assert the new ts_* aliases, plus a new end-to-end test asserting concrete epoch integers never appear in note text only the translated date does.

image


marcociav-exmergo (approved):
LGTM, nice work @emnarahmouni-exmergo

PR #255, opened 2026-08-09
Expired BigQuery credentials report a prerequisite

### LINKED ISSUE #252: Expired credentials classify as `reason: internal`, so a fixable setup problem reads as an engine bug
## Summary

When a connector's credentials are *present but stale*, the refusal comes back
with `reason: internal`. That reason is documented as "unclassified: not a
deliberate dex refusal", so a host branching on it sees an engine defect rather
than a setup problem the user can fix in one command and retry.

## Repro

1. Authenticate a billed connector normally, then let the credentials go stale
   (on BigQuery, an Application Default Credentials session that has aged out;
   equivalently, revoke or expire the token the connector discovered).
2. Run any command that reaches the warehouse, for example `maintain check`.

## Actual

```json
{ "status": "error",
  "reason": "internal",
  "errors": ["Reauthentication is needed. Please run `gcloud auth application-default login` to reauthenticate."] }
```

The message is excellent. The classification contradicts it: the message names a
setup command and promises that retrying afterwards will work, which is the
definition of a *deliberate* refusal.

## Expected

`reason: prerequisite`, which `envelope.py` documents as "run a named setup
command first, then retry". That is exactly this situation.

## Why this matters

#170 added `reason` so a host could branch on retry/setup/stop without parsing
prose. `internal` is the one value that says "do not bother retrying, this is our
bug". An expired token is the single most common recoverable failure in a long
session, and it is currently the one case where the field actively misleads. A
wrapper that surfaces "dex hit an internal error" for an expired login sends the
user to the tracker instead of to `gcloud`.

## Cause

Credential *discovery* is handled. `connect.py` catches the absent-credentials
error and re-raises it as `CredentialDiscoveryError`, which maps to
`configuration`, with a comment noting it "will not appear on a retry".

Expiry is a different exception raised at a different time. It surfaces lazily
when the client first refreshes a token, which is inside the adapter's own
request path, well past discovery. Nothing catches it, so `reason_for` falls
through every entry in `_reason_overrides()` and returns `INTERNAL`.

This is not connector-specific. Grepping the adapters for expiry or refresh
handling turns up nothing on any of them, so every connector with a token that
can age out has the same gap. Fixing one connector is a complete, useful change;
the rest can follow the same shape.

## Suggested fix

Translate the expiry exception at the adapter boundary into a typed dex error
that maps to `PREREQUISITE`, the way discovery already translates the
absent-credentials case.

Deliberately *not* `configuration` (the bucket its discovery-time sibling uses):
that one is documented as the case that will not change on a retry, and re-auth
plus retry is precisely what fixes this. Deliberately not `connection` either;
the backend was reachable, and the credential was the problem.

Prefer translating in the adapter over adding the provider's exception class to
`envelope.py`'s override table, so the envelope layer does not grow an import
from a connector SDK.

## Scope for a first contribution

Small and self-contained. One connector, one translation, one test that raises
the provider's refresh error from the existing test fake and asserts the reason.
Worth confirming the `prerequisite` choice with a maintainer before writing it,
since the reason taxonomy is a published contract.

nightcityblade:
## What this changes

Translates Google Auth `RefreshError` failures at the BigQuery scope-resolution boundary into the existing public `PrerequisiteError`. This preserves the provider reauthentication message while changing the envelope reason from `internal` to `prerequisite`, so hosts can direct users to reauthenticate and retry.

The request wrapper drains lazy dataset iterators inside the translation boundary and also covers configured-dataset probes. A regression test raises the real provider exception lazily and asserts both the reason and unchanged remediation message.

Validation:
- BigQuery adapter tests: 51 passed
- Error, envelope, and BigQuery suites: 85 passed
- Safety-spine and dev-target suites: 166 passed
- Eval scoring-core: 26 passed
- Ruff lint and format checks passed
- Full core run: 1,626 passed and 129 skipped before the remaining optional connector extras were installed; every initially failing group was then rerun successfully with all extras

## Related issues

Closes #252

## Checklist

      cost-guarded, PII flagged not surfaced, propose-dont-impose

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- Expired BigQuery credentials now report a retryable prerequisite instead of
+  an internal error.
+
 ## [1.6.1] - 2026-08-09
 
 ### Fixed
--- a/packages/dex-core/src/exmergo_dex_core/adapters/bigquery.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/bigquery.py
@@ -16,11 +16,12 @@ simply calls no mutating client API, and the docs recommend read-only roles
 
 from __future__ import annotations
 
+from collections.abc import Callable
 from typing import Any
 
 from ..config import BigQueryTarget
 from ..envelope import Paradigm
-from ..errors import ConnectorError
+from ..errors import ConnectorError, PrerequisiteError
 from ..guards.cost_guard import CostGate, OverCeilingError
 from ..guards.sql_guard import assert_select_only
 from .base import (
@@ -104,6 +105,7 @@ class BigQueryAdapter:
         # [bigquery] extra; only this adapter pulls it in.
         try:
             from google.api_core import exceptions as api_exceptions
+            from google.auth import exceptions as auth_exceptions
             from google.cloud import bigquery
         except ImportError as exc:
             raise RuntimeError(
@@ -112,6 +114,7 @@ class BigQueryAdapter:
             ) from exc
         self._bq = bigquery
         self._api_exceptions = api_exceptions
+        self._auth_exceptions = auth_exceptions
         self._client = client or bigquery.Client(
             project=project, credentials=credentials
         )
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (approved):
Thanks for this @nightcityblade !
LGTM! Will merge now

PR #256, opened 2026-08-09
Accept semantic query metrics positionally

### LINKED ISSUE #253: `explore semantic query` takes its metrics as a repeated flag while every sibling verb takes its object positionally
## Summary

`explore semantic query` requires `--metric <name>`. Every other `explore` verb
takes the thing it operates on as a positional argument. The inconsistency is
easy to trip over, and the failure is an argparse error that does not point at
the fix.

## Repro

```
dex explore semantic query <metric-name> --group-by <dimension>
```

## Actual

```
dex: error: unrecognized arguments: <metric-name>
```

The usage line printed alongside it lists only the top-level flags, so it does
not show that `--metric` is what was wanted.

## Expected

The metric is the object of the verb, so it should be accepted positionally:

```
dex explore semantic query <metric-name> --group-by <dimension>
```

## Why this matters

The surrounding verbs set an expectation this one breaks: `explore query` takes
its SQL positionally, `explore cluster` takes its object positionally, and
`explore profile` takes one or more objects positionally. `explore semantic
query` is the only one where the object moved into a flag, and the error it
produces when you follow the house pattern does not name `--metric`.

Small, but it is the first thing you type when you reach for the command, and
#135 established that this class of ergonomic fix on this exact surface is
welcome.

## Suggested fix

Accept metrics positionally *after* the explicit `query` mode, and keep
`--metric` working so nothing breaks.

One wrinkle worth knowing before starting: the subcommand already has an
optional positional (`mode`, with `choices=["list", "query"]` and a default of
`list`), so a second optional positional makes bare `explore semantic <name>`
with no mode resolve `<name>` against the mode choices and fail with an invalid
choice. That is acceptable, since `query` is an explicit mode, but the
implementation should not try to make the mode-less form work, and a test should
pin the behavior so the ambiguity is recorded rather than rediscovered.

Failing that, the cheaper fix is to make the error name the flag.

ShlokShar:
Accept metric names as positional arguments after the explicit `query` mode, while keeping `--metric` working for existing callers.

Add parser and command-handler coverage for positional, flag, and mixed forms, and document the mode-less ambiguity.

Fixes #253


marcociav-exmergo (changes requested):
This looks good @ShlokShar.
Could you also add a brief summary of your changes on the CHANGELOG.md?
Make sure you rebase first, we have a lot of incoming PRs today.

Once that is done I will approve and merge!

ShlokShar:
Done — rebased onto the latest main and added a brief entry under Unreleased in CHANGELOG.md. I reran the focused semantic/CLI suites (85 passed, 7 skipped) and Ruff; both are clean.

marcociav-exmergo (approved):
Thanks for this, awesome work!

Issue #258, opened 2026-08-09
A tier-3 format has no way to say where a declaration lives


catincloudlabs:
### Area

DexEngine / command contract

### Problem

Reconcile's mechanical edits are pinned to literal paths:

```python
model_path = f"models/staging/stg_{table}.sql"
path = f"models/staging/stg_{table}.yml"
```

`Edit` itself is already format neutral, which is the encouraging part: a path, new content, the hash it was planned against, and an op. Nothing about that shape is dbt specific, and `write_edits(edits, project_dir, *, confirmed=False)` is likewise a signature any format can implement.

What is dbt specific is the path construction, and it happens before the format is consulted. So even with the class check in the sibling issue resolved, the edits handed to a second format name files it does not have. The format's only options are to refuse them or to guess a mapping, and a guessed mapping that is wrong writes into the wrong file with a hash check that passes because the file was not there.

Worth stating plainly because it is easy to assume otherwise: this is not a request to let a format author its own edits. It is that reconcile currently decides *where* before asking *who*.

### Proposed solution

A path resolution seam on the write tier: given a model name and the kind of artifact being proposed, the format says where that lives, and reconcile builds the `Edit` against the answer. A directory-keyed format returns what it returns today, so nothing changes for dbt.

That is deliberately the smaller of the two designs I can see. The larger one, where the format authors the edit content as well, is a much bigger ask and I do not think the evidence justifies it yet.

### Alternatives considered

**Format-authored edits.** Reconcile hands over findings and the format returns edits. More flexible, and it moves the "what does a staging model look like" question out of the engine. I am not proposing it: it multiplies what a format has to implement to reach tier 3, and the tier is already the one most formats should decline.

**A configurable path template.** Cheaper than either, and wrong in the way config usually is here: the mapping is a property of the format, not of the deployment, so putting it in `.dex/config.yml` invites two projects of the same format to disagree about their own layout.

### Context worth having

I maintain a Dagster asset graph format for dex, in its own distribution, currently at tier 2 and passing the shipped conformance suite. It declines tier 3 today, and until recently I justified that with "a project reduced from a running graph cannot receive an edit."

[...]

Issue #257, opened 2026-08-09
reconcile's write gate asks for a class, not a tier


catincloudlabs:
### Area

Reconcile

### Problem

The project seam is tiered, and `maintain reconcile` reads the tier before it proposes an edit. That is the right shape, and it is why I can decline the write tier honestly rather than relying on my directory names not colliding with the scaffold convention.

But after the tier check passes, `maintain/commands.py` narrows again:

```python
editable = engine.editable_project()
if editable is None:
    ...advisory, format declined the write tier...
elif not isinstance(editable, DbtProject):
    ...advisory, "dex cannot yet author them for another format"...
```

So a format that implements `EditableProject` in full, passes `EditableProjectContract`, and honours both `project_dir` and `confirmed`, gets exactly what a format that declined the tier gets. The tier is asked and then overruled by a class check.

I am not arguing the current behaviour is wrong today: the mechanical edits really are dbt artifacts, and writing them into another format's tree would be worse than refusing. The problem is that the refusal is expressed as `isinstance(x, DbtProject)`, which makes it permanent and makes the tier decorative for everyone else. There is no way for a second format to become eligible, and nothing in the conformance suite fails to tell an implementer that.

### Proposed solution

Make the second gate a question the format can answer, so implementing the tier can eventually mean something. The narrow version does not require deciding how a second format authors an edit: keep refusing, but refuse on a capability rather than on a type, and let a format that can accept the edit say so.

The warning text is already right. It is only the predicate that hard-codes a class.

### Alternatives considered

Leaving it. That is defensible while there is one format, and it is what I would do if the tier did not already exist. Given that it does, an implementer can pass the whole shipped contract and discover the tier bought them nothing, which is the discovery I would rather they not make after doing the work.

### Context

I maintain a Dagster asset graph project format in its own distribution, currently at tier 2 and passing the shipped conformance suite including `ProjectFactoryContract`. This is the first of two issues about what stands between such a format and the write path; the second is about where a proposed edit lands.

catincloudlabs:
Sibling issue: #258, on where a proposed edit lands. These are independent, and resolving either alone leaves the write path closed to a second format: this one decides *whether* a format may receive an edit, #258 decides *where* the edit goes.

catincloudlabs:
Two things I got wrong in the framing here, both found by building it rather than reading it. Draft PR #263 has the code; this is the part worth reacting to without opening a diff.

[...]

PR #260, opened 2026-08-09
docs(references): describe the tiers as the extension seam, not ProjectAdapter

catincloudlabs:
`references/canonical-model.md`'s section titled "The extension seam (more formats
later)" still names a single `ProjectAdapter` protocol with `load()` and
`write_edits()` as the live seam, with no mention of the three tiers.

That protocol is superseded, unreferenced anywhere in the distribution, and
satisfies no tier. A class written against it fails `isinstance(project,
ExploreProject)` and `construct_project` refuses it as *"not a project: it is
missing definitions"*.

Which makes this the worst place in the tree for that particular staleness: it is
the document someone reads *before* `references/project.md`, so following it
produces exactly the class the engine rejects.

## What changed

The section now names the three tiers with a small table, says why they are tiers
rather than a capability flag, and hands off to `references/project.md` for the
contract itself rather than restating it. `ProjectAdapter` gets a closing note
instead of silence, since a reader who has seen it elsewhere needs to know it is
not the thing to implement.

The section's actual argument, that supporting more formats does not require a
neutral internal model, was correct and is kept, with the reason sharpened: it is
`ProjectDefinitions` being reused rather than a parallel vocabulary that would have
to be mapped at every call site.

## Verification

Docs only. `scripts/check_no_em_dashes.py` clean, LF endings, no code touched and
no test asserts anything about this file.

Found while writing a second format against the seam. Two related issues are filed
separately (#257, #258), and this one is independent of both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Changed
+
+- `references/canonical-model.md`'s extension-seam section described a single
+  `ProjectAdapter` protocol with `load()` and `write_edits()` as the live seam. That
+  protocol is superseded, unreferenced, and satisfies no tier, so a format written
+  against it is refused as "not a project", and this is the document a reader reaches
+  before writing one. It now describes the three tiers and points at
+  `references/project.md`.
+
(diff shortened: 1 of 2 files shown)


marcociav-exmergo (approved):
LGTM!
Docs drifting was my bad, thanks for this @catincloudlabs

PR #261, opened 2026-08-09
docs(project): decline tier 3 on the artifact, not on the project's origin

catincloudlabs:
`EditableProject`'s docstring, `references/project.md` and `CONTRIBUTING.md` all
state the tier-3 decline as an absolute:

> A project reduced from a running graph cannot receive an edit, because its source
> of truth is the code that produced the graph.

That is right about the reduction and wrong as a general rule, and the two come
apart more often than the graph example suggests.

## Why it is wrong as stated

An asset graph carries neither column names nor join keys. So a format over one
*cannot* read its declared keys, joins or semantics from the graph. It reads them
from somewhere else, and that somewhere is usually a hand-authored file that nothing
regenerates.

Such a file is a real source of truth, and it is precisely the shape reconcile's
highest-value mechanical edit already targets: adding a `unique` test to a column in
schema YAML. A format holding one may honestly reach tier 3 for that channel while
still refusing to author a model.

So the test is **which artifact an edit lands in and whether anything rewrites it**,
not where the project came from. Deciding on origin alone declines a tier the format
could serve, and the docs were teaching exactly that.

## How I found it

By writing a second format and getting it wrong in that direction. I maintain a
Dagster asset graph format at tier 2, and I justified declining tier 3 with the
sentence above for months. It is false for half my project: the models are a
reduction, and the declarations are hand-written YAML that only ever gets read.

The `CONTRIBUTING.md` copy is the one I would flag, since it sits in the section a
format author reads before starting.

## Scope

Docs and one docstring. No behavior changes, and no test asserts the old claim.

Two other sites survive review and are deliberately unchanged, because both phrase
it conditionally rather than universally: `maintain/commands.py:687` describes *a
format that cannot receive an edit*, and `test_safety_spine.py`'s `_NotEditableProject`
is one such format by construction.

## Verification

`ruff check` and `ruff format --check` clean, `tests/adapters` 352 passed 6 skipped,
`check_no_em_dashes.py` clean, LF endings.

Related: #257 and #258, which are about what stands between a tier-3 format and
reconcile actually reaching it. This PR only corrects the reasoning and does not
depend on either.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Changed
+
+- **Declining tier 3 is a question about the artifact an edit lands in, not about
+  where the project came from.** `EditableProject`, `references/project.md` and
+  `CONTRIBUTING.md` all stated the decline as an absolute: a project reduced from a
+  running graph cannot receive an edit. That is right about the reduction and wrong
+  as a general rule, and the two come apart more often than the graph example
+  suggests. An asset graph carries neither column names nor join keys, so a format
+  over one reads its declared keys, joins and semantics from somewhere else, and that
+  somewhere is usually a hand-authored file that nothing regenerates. Those files are
+  a real source of truth and they are the shape `reconcile` already proposes edits
+  to, so a format holding one may reach tier 3 for that channel while still refusing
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
LGTM!
Docs drifting was my bad, thanks for this @catincloudlabs

PR #262, opened 2026-08-09
feat(conformance): assert the content a format declares, not only its shape

### LINKED ISSUE #259: The conformance suite cannot see three ways a project format is wrong
### Area

DexEngine / command contract

### Problem

I ran the shipped conformance suite against a second format and passed it, then found three defects the suite structurally cannot reach. Each is a way to be green while losing something a caller depends on.

**1. `declared_composite_keys` is never asserted.** `DeclaringProjectContract` checks a single-column key and a join. Nothing in the suite touches the composite field, even though it is a distinct type from `DeclaredKey` with different multiplicity. A truncated composite key is the expensive version of this, because it does not read as a missing declaration. It reads as a declared grain that is simply narrower than the truth, so everything downstream runs against a grain the author never claimed.

Worth noting for whatever fixture this gets: two columns is not enough. A format that handles a composite key by special-casing the pair passes a two-column fixture and fails a four-column one.

**2. `test_a_declared_join_carries_both_sides` cannot fail for the right reason.** It takes the model, column, target model and target column from the implementer's own fixture. If both ends of that fixture are spelled the same, an implementation that reads the source column and copies it onto the target satisfies the assertion exactly.

That is not hypothetical. In my own suite, mutating `to_columns=(field,)` to `to_columns=(column,)` left everything green until I added a case whose two ends were named differently. dbt's `relationships` test exists precisely to express a foreign key whose ends are spelled differently, and a format that mirrors would join on a column the target may not even have, surfacing as a wrong answer rather than as a failure.

**3. A semantic layer mapped entirely to `None` passes the whole suite.** This is the one that cost me a real check, and it is the reason I am filing.

`MaintainProjectContract` asserts that an *empty* project produces an empty semantic layer. Nothing looks at a populated one. So a format that reads every dimension and measure name and drops the physical column behind each is green.

`SemanticModelDef` keys every field to a warehouse column, and `maintain/drift.py` skips any field whose column is `None`, correctly, since it cannot resolve what it was not given. A layer of all `None` therefore validates, serializes, and compares clean forever: the check does not fail, it never runs.

Measured on my own format rather than reasoned about. With the column mapped, dropping a warehouse column yields **1 finding at severity `high`**. With `None`, **0**. Third arm run as a discriminator: mapped to a column that still exists also yields 0, so the detector is not merely firing whenever a column is present. Eleven mappings were affected in my project, plus two `categorical_dimensions`.

### Proposed solution

Two optional hooks on `DeclaringProjectContract` and one new opt-in mixin. Optional rather than mandatory, because an existing implementer's green suite should not turn red on an upgrade, and a format may genuinely have no multi-column grain to express: dbt's own `unique` test is column level.

- `a_project_declaring_a_composite_key()`, returning `(project, model, columns)`, asking for more than two columns.
- `a_project_declaring_a_join_with_differently_named_sides()`, which refuses a mirrored fixture, since a mirrored one cannot detect mirroring.
- `SemanticProjectContract`, whose hook returns the field-to-column mappings the format expects to arrive, and which also checks that `categorical_dimensions` holds only real columns. That mapping takes `str` rather than `str | None`, so a field that is categorical *and* unresolved has to be left out rather than given an invented column, and the two properties have to stay independent.

Where a hook defaults to skipping, the skip message should name what goes unchecked. A silent skip is how a format arrives at "the suite is green" without knowing what it did not run.

I have this implemented and mutation-tested against `DbtProject`, which can express all three: truncating a composite key, mirroring a join target, and dropping the semantic column each turn exactly the intended assertion red. Happy to open a PR if the shape is acceptable, or to adjust it first.

### Alternatives considered

**Making them mandatory rather than opt-in.** Cleaner, and a skip really is a weak signal. Rejected because it is a breaking change to a published contract for every existing implementer, and because the composite-key case has a legitimate "my format cannot express this" answer.

**Leaving it and documenting the gaps in `references/project.md`.** Prose does not fail a build. The whole argument for shipping a conformance suite rather than a checklist is that the checklist gets read once.

catincloudlabs:
Implements #259. Three ways a project format can pass the shipped suite and still be
wrong, each costing a check a caller depends on.

Found the ordinary way: by writing a second format, passing the suite, and then
finding the defects anyway.

## What is added

**Two optional hooks on `DeclaringProjectContract`.**

`a_project_declaring_a_composite_key()` reaches `declared_composite_keys`, a distinct
type from `DeclaredKey` that nothing in the suite touched. It asks for more than two
columns on purpose: a format that special-cases the pair passes a two-column fixture
and fails a four-column one. A truncated composite key is the expensive shape here,
because it does not read as a missing declaration. It reads as a declared grain that
is simply narrower than the truth, so everything downstream runs against a grain the
author never claimed.

`a_project_declaring_a_join_with_differently_named_sides()` covers what
`test_a_declared_join_carries_both_sides` structurally cannot. That assertion takes
both sides from the implementer's own fixture, so if the fixture names them the same,
an implementation that copies the source column onto the target satisfies it exactly.
This hook refuses a mirrored fixture rather than accepting one that cannot detect
mirroring. In my own suite, mutating `to_columns=(field,)` to `to_columns=(column,)`
left everything green until a differently-named case existed.

**`SemanticProjectContract`, new, and the largest hole.**

`MaintainProjectContract` only ever looks at an *empty* semantic layer. So a format
that reads every dimension and measure name and drops the physical column behind each
passes the whole suite. `SemanticModelDef` keys every field to a column and
`maintain/drift.py` skips any field whose column is `None`, correctly, since it cannot
resolve what it was not given. A layer of all `None` therefore validates, serializes
and compares clean forever: the check does not fail, it never runs.

Measured rather than reasoned about, on the format that hit it: a dropped warehouse
column yields **1 finding at severity `high`** with the column mapped, and **0**
without. Discriminating third arm: mapped to a column that still exists also yields 0,
so the detector is not merely firing whenever a column is present.

## Why opt-in

Mandatory would be a breaking change to a published contract for every existing
implementer, and the composite-key case has a legitimate "my format cannot express
this" answer, since dbt's own `unique` test is column level. Where a hook defaults to
skipping, the skip names what goes unchecked: a silent skip is how a format arrives at
"the suite is green" without knowing what did not run.

The two questions I flagged on #259 are still the ones worth a maintainer view, and I
am happy to change either:

1. opt-in versus mandatory;
2. `SemanticProjectContract` as a separate mixin rather than folded into
   `MaintainProjectContract`. Separate because a format that genuinely declares no
   semantics should not have to opt out of anything, and its empty layer is already
   correct under the tier contract.

## Verification

`DbtProject` implements all three in `tests/adapters/test_project_parity.py`, so the
shipped format is held to them too.

Mutation-tested rather than read. Each turns exactly the intended assertion red, and
the restore is green:

| mutation | result |
|---|---|
| composite key truncated to one column | 1 red |
| FK target column mirrored from source | 2 red |
| semantic field column dropped | 1 red |

`ruff check` and `ruff format --check` clean, `tests/adapters` 356 passed 6 skipped,
`check_no_em_dashes.py` clean, LF endings. Rebased on `main` after #260 and #261.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,40 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **The conformance suite reaches the content a format declares, not just its
+  shape** ([#259]). Three assertions, each covering a way a format can pass the
+  existing suite while being wrong in a way that costs a real check.
+
+  `DeclaringProjectContract` gains two optional hooks.
+  `a_project_declaring_a_composite_key()` reaches `declared_composite_keys`, which
+  is a separate field from `declared_keys` and which nothing in the suite touched
+  before; it asks for more than two columns, because a format that special-cases the
+  pair passes a two-column fixture and fails a four-column one. A truncated
+  composite key does not read as a missing declaration, it reads as a narrower grain
+  that is simply wrong, which is what makes it worth its own assertion.
+  `a_project_declaring_a_join_with_differently_named_sides()` covers the case
+  `test_a_declared_join_carries_both_sides` structurally cannot: if a fixture names
+  both ends of a join the same, an implementation that copies the source column onto
+  the target satisfies it exactly. The contract refuses a mirrored fixture here.
+
+  `SemanticProjectContract` is new, and it closes the largest hole.
+  `MaintainProjectContract` asserts that an empty project yields an empty semantic
+  layer and never looks at a populated one, so a format that reads every dimension
+  and measure name and drops the physical column behind each passed the whole suite.
+  `SemanticModelDef` keys every field to a column and the drift detector skips any
+  field whose column is `None`, so a layer mapped entirely to `None` validates,
+  serializes, and compares clean forever: the check does not fail, it never runs, and
+  a dropped warehouse column that should raise `dangling_reference` at high severity
+  raises nothing. The absence is indistinguishable from agreement.
+
+  All three are opt-in, and the two on `DeclaringProjectContract` skip with a message
+  naming what goes unchecked rather than silently: an existing implementer's green
+  suite must not turn red on an upgrade, and a format may genuinely have no
+  multi-column grain to express. `DbtProject` implements all three in
+  `tests/adapters/test_project_parity.py`, so the shipped format is held to them too.
+
 ## [1.6.2] - 2026-08-09
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,13 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   a dropped warehouse column that should raise `dangling_reference` at high severity
   raises nothing. The absence is indistinguishable from agreement.
 
+  `SemanticProjectContract` asserts tier 2 before anything else, against the project
+  its own hook returns. `semantic_layer` is a tier-2 member, so a format mixing the
+  contract in beside `ExploreProjectContract` would otherwise fail with
+  `AttributeError: no attribute 'semantic_layer'`, naming the missing attribute
+  rather than the tier it belongs to. Checking the hook's project rather than
+  `make_project()` keeps the mixin dependent only on the one fixture it declares.
+
   All three are opt-in, and the two on `DeclaringProjectContract` skip with a message
   naming what goes unchecked rather than silently: an existing implementer's green
   suite must not turn red on an upgrade, and a format may genuinely have no
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -293,6 +293,18 @@ raises there turns an ordinary state into an outage. Override
 `make_unreadable_project()` to get those assertions running against your format
 instead of skipped.
 
+Mix in the contracts covering what your format *declares*, beside the one for your
+tier. `DeclaringProjectContract` checks that a declared key and a declared join
+arrive, and carries two further hooks worth supplying: a composite key of more than
+two columns (a format that special-cases the pair passes a two-column fixture), and
+a join whose two ends are spelled differently (if both ends of your fixture share a
+name, an implementation that mirrors one side onto the other satisfies it exactly).
+`SemanticProjectContract` checks that each semantic field keeps the warehouse column
+behind it, which is the one no tier assertion can see: the tier contract only looks
+at an *empty* semantic layer, and a format that reads every field name and drops the
+columns maps everything to `None`, which validates, serializes, and compares clean
+forever while the drift check silently never runs.
+
 Mix `ProjectFactoryContract` in front of your tier contract if dex will build your
 format from a name rather than be handed an instance, which is what a host reaching
 dex as a subprocess needs. Naming one is the same shape as naming a storage backend:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -303,7 +303,9 @@ name, an implementation that mirrors one side onto the other satisfies it exactl
 behind it, which is the one no tier assertion can see: the tier contract only looks
 at an *empty* semantic layer, and a format that reads every field name and drops the
 columns maps everything to `None`, which validates, serializes, and compares clean
-forever while the drift check silently never runs.
+forever while the drift check silently never runs. It checks your format reaches
+tier 2 before any of that, so mixing it in beside the tier-1 contract fails with a
+sentence naming the tier rather than with a missing-attribute error.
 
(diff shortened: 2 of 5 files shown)


catincloudlabs:
Wired all three hooks up in a real out-of-tree format before asking you to merge this, since the hook names become surface that formats pin against. Verified against this branch as an actual dependency (`git+...@feat/conformance-declared-content`), not a local checkout: **26 assertions become 30, none skipped.** All three pass.

So the shapes work. Two things the wiring surfaced that are worth your call before merge rather than after.

**1. `SemanticProjectContract` calls a tier-2 method while inheriting no tier contract.**

Its assertions call `project.semantic_layer()`, but the class is a bare mixin. A format that mixes it beside `ExploreProjectContract` gets `AttributeError: 'MyProject' object has no attribute 'semantic_layer'` rather than something naming the tier. `MaintainProjectContract` would have caught that with its `tier_of(...) >= 2` assertion.

Three ways out, and I do not have a strong preference:

- have it inherit `MaintainProjectContract` (loses the pure-mixin property, but semantics genuinely presuppose tier 2);
- keep it a mixin and add a `tier_of(project) >= 2` assertion to it;
- leave it and document the requirement.

I lean toward the second: it keeps the mixin standalone for a format that wants only the content assertions, while failing with a sentence instead of an attribute error.

**2. `a_project_declaring_a_semantic_model()` invites something a real format may not be able to supply.**

The guidance says to include a field mapping to `None` "if you can produce one", and the hedge turns out to be load-bearing. My fixture builders are all-or-nothing: one produces a complete declaration where every field carries its `expr`, the other produces three fields that resolve to nothing, and neither is mixed. So I supply the fully-mapped one, and the `None` case goes unexercised **through your contract** even though my own suite covers it directly.

Not a defect, and I am not asking for a change: the mapping type already accepts `str | None`, so a format whose fixtures happen to be mixed can supply one. Flagging it because the hedge is doing real work, and I would keep it rather than tighten the wording into something implementers cannot satisfy.

---

Consumer branch is `catincloudlabs/catincloud-labs-vps@test/upstream-declared-content-hooks`, held deliberately: it fails at collection against the released 1.6.1 with `no attribute 'SemanticProjectContract'`, so it merges with the pin bump after whatever release carries this, not before.

Happy to push either change to `SemanticProjectContract` here if you tell me which you prefer.

marcociav-exmergo:
@catincloudlabs thanks for this!

For the open decision: let's keep it a mixin and add a tier_of(project) >= 2 assertion to it, as you recommend.

Also please rebase on main, we had a lot of contributions today!

catincloudlabs:
Both done: rebased on `main` (through #251), and `SemanticProjectContract` now asserts the tier as you chose.

**The assertion is on the project the contract's own hook returns, not on `make_project()`.** That was the one judgement call left in your answer, and I went this way because it keeps the standalone property that made option 2 preferable in the first place: the mixin ends up depending only on `a_project_declaring_a_semantic_model()`, the single fixture it declares, rather than quietly acquiring a second requirement on a method it does not define. It is also the object the failure actually happens on, since every assertion in the contract calls `semantic_layer()` on the hook's project.

Proven both ways against a tier-1 format rather than assumed. With the guard:

> a format declaring semantics has to reach tier 2, because `semantic_layer()` is a `MaintainProject` member and every assertion in this contract calls it. Implement `transform_layer()` and `semantic_layer()`, or drop this mixin if the format declares no semantics: an empty layer is already correct under the tier contract

Without it, the same fixture raises `AttributeError: '_Tier1Only' object has no attribute 'semantic_layer'`, which is the error the guard replaces.

`CHANGELOG.md`, `CONTRIBUTING.md` and `references/project.md` all note it, including why the check reads the hook's project.

## On the rebase

Only `CHANGELOG.md` conflicted. You had cut `1.6.2` and emptied `[Unreleased]`, so this entry moved into the now-empty `[Unreleased]` and your `1.6.2` section is untouched. The other four files did not overlap.

One note on the local run: `tests/transform/test_dev_target.py::test_retargeted_dev_database_is_refused_naming_both_values` fails on Windows, on `main` as well as on this branch, because the message carries `analytics\profiles.yml` while the assertion looks for `analytics/profiles.yml`. Nothing to do with this PR and your CI is Linux, so it is green there. Mentioning it only because it is a real path-separator assumption in a test, if you want it filed separately.

## Point 2, unchanged

The `None`-mapped field stays a hedge and I am still not asking for a change. Repeating the substance so it does not get lost in the diff: our fixture builders are all-or-nothing, so we supply the fully-mapped model and the `None` case goes unexercised through your contract even though our own suite covers it directly. The wording is doing real work as it stands.

marcociav-exmergo (approved):
LGTM!

PR #263, opened 2026-08-09
A project format says where a proposed edit lands, and which paths it owns

### LINKED ISSUE #257: reconcile's write gate asks for a class, not a tier

### LINKED ISSUE #258: A tier-3 format has no way to say where a declaration lives

catincloudlabs:
## What this changes

Both issues, and the two gates behind them that neither describes. Marco's call on the design question this PR was opened to ask: containment widens so a format can declare its own editing surface, and dbt's stays intact.

**The two issues are one seam.** #257 is a class check deciding *whether* a format may receive an edit, #258 is hard-coded paths deciding *where*. They separate less than they look: the paths reconcile builds are not filesystem paths, they are keys into the view the format's own `load()` returned (`reconcile.py` looks up `model_path not in view.files` and `view.files.get(path)`). So a format that answers "where does an edit of this kind for this table go" supplies both. `PlacingProject` asks that question, and the gate stops asking `isinstance(editable, DbtProject)`.

**It sits beside `EditableProject`, not on it.** The tiers are `runtime_checkable`, so a method added to tier 3 demotes every format that has not implemented it yet: `tier_of` starts answering 2 where it answered 3, and the write path closes for exactly the implementers who were already passing.

## The two further gates, both now open

**Containment, at plan time.** `plans.plan` validated every edit against **dbt's** `model_paths` whatever format produced it, so a second format placing a declaration sidecar was refused even with #257 and #258 resolved exactly as proposed. `editing_surface()` is where a format declares the region its edits may land in, and containment reads that declaration. It stays a safety property and stays mandatory; only the authority moved. A format that declares no surface validates against dbt's, as before.

It is a second method rather than a widened `edit_path` because `transform plan` validates edits an *agent* authored: there is no `(kind, model)` pair to ask placement about and no prior answer to compare against, only a path and the question of whether the format admits to owning it.

**The pin has to come from the same view as the surface.** Reconcile already built the sidecar's *content* from the format's view while `plans.plan` read `view.files` off dbt's. Widening containment alone leaves those two disagreeing, which is not a refusal but a quiet defect: the sidecar pins `old_content_hash=None`, the reviewable diff renders an existing hand-written file as a whole-file create, and the apply that follows reports a conflict on a file nobody touched. Both halves now come from the format.

**A fourth gate, at apply time.** `plans.apply` wrote every plan with `dbt_project.write_edits`, which resolves each edit under the dbt project and re-hashes what it finds on disk. So the plan a format can now store would have been refused one stage later, and the tier-3 `write_edits` it implemented would never have been called. A plan a format can store and not apply is not a write path, so `transform apply` routes through the format the plan was planned against.

That one comes with a widening of the tier-3 contract, which is the change here most worth your eye: `EditableProject.write_edits` now documents that its return has to report `written` and `conflicts`, and `EditableProjectContract` asserts it. The apply path reads both to tell a refusal from a write, and a result answering neither fails in both directions at once. A format returning `dbt_project.ApplyResult`, or anything exposing those two, already satisfies it.

## dbt is unchanged, by construction

`edit_path` returns the scaffold convention reconcile hard-coded. `editing_surface` returns the configured model and macro paths containment already read. The root manifests are still allowed by name, the macro and root-config kind checks still run, `validate_deletions` and the missing-macro warnings still run, and a plan planned against dbt is still written by dbt's writer.

Which validations apply is asked of the **view's shape**, not of the format's class: a format handing back a `DbtProjectView` is offering dbt's surface and gets dbt's checks. That keeps the seam from growing back the `isinstance(project, DbtProject)` gate it exists to remove.

## Two smaller findings, both fixed rather than flagged

**The kinds are not equally reachable, and `None` per kind is load-bearing.** `SCHEMA_YML` is a mutation of the format's own file. `MODEL_SQL` carries dbt SQL that `model_edits` generates alongside the path, so placement alone cannot open that channel; a format that places a staging model elsewhere is refused rather than written into. This is the argument for a seam over a boolean: a single flag would have to be wrong about one of the two kinds for the graph-derived format that motivated #258.

**The convention leaked through content, not only paths.** `_grain_test_edits` matched `model.get("name") == f"stg_{table}"` inside the YAML, so a sidecar naming its model `orders` was missed even when the path resolved. It now reads the model from the file the format chose, which is the same string for dbt (`models/staging/stg_<table>.yml` has stem `stg_<table>`). The fixture no longer needs the `stg_` prefix to get that far, which is what pins it. The miss also has its own warning now instead of sharing a silent `continue` with the already-alerting case.

## Conformance

`PlacingProjectContract` ships beside the others, and `TestDbtProject` mixes it in so the shipped format runs what a third party runs. Its one assertion worth having is the cross-check neither method can make alone: every path `edit_path` returns has to be inside `editing_surface()`. A format placing outside its own declared surface builds a proposal the plan store then refuses, and from the outside that reads as dex declining rather than as the format contradicting itself.

## Related issues

Closes #257. Closes #258.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`): 1996 passed, 80 skipped

On the safety box, since this PR moves a guard. Containment was the safety property under discussion and it is still enforced on every edit, still ahead of any write, and still refuses escapes (absolute paths, `..`) without consulting the format at all. What changed is that dbt no longer answers the question on another format's behalf. Propose-don't-impose is likewise unchanged and now reaches further: a second format's plan goes through the same hash-pinned, conflict-refusing apply, which is what the widened `write_edits` return exists to keep honest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,15 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   change as a whole-file create, and the apply that follows reports a conflict on
   a file nobody edited.
 
+  That holds for agent-authored edits too, which is the whole write surface
+  outside reconcile: `transform plan`, `transform macro`, and every
+  `semantic define|update|plan` share one call, and it asked the engine for dbt's
+  directory unconditionally. A format declaring a surface now supplies the
+  directory from its own view, so a repository with no `dbt_project.yml` can
+  reach those commands at all, and one with a dbt project elsewhere in the tree
+  no longer pins an edit against a file the apply will not write. dbt is on the
+  same path either way: its view loads the directory the engine would have named.
+
 - **A plan is applied through the format it was planned against** ([#257]).
   `transform apply` wrote every plan with dbt's writer, which resolves each edit
   under the dbt project and re-hashes what it finds on disk, so a plan a second
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,81 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **A project format can say where its edits land, and which paths it owns**
+  ([#257], [#258]). `maintain reconcile` read the project seam's write tier and
+  then narrowed again on `isinstance(editable, DbtProject)`, so a format that
+  implemented tier 3 in full and passed the shipped conformance suite got exactly
+  what a format declining the tier got. Underneath that, the paths reconcile
+  proposed were literals (`models/staging/stg_<table>.sql` and its `.yml`) built
+  before the format was consulted, so a second format was handed edits naming
+  files it does not have. The two are one seam: those paths are keys into the
+  view the format's own `load()` returned, so a format that answers where an edit
+  goes supplies both halves.
+
+  `PlacingProject` is that seam, beside `EditableProject` rather than on it: the
+  tiers are `runtime_checkable`, so a method added to tier 3 would demote every
+  format that has not implemented it yet, closing the write path for exactly the
+  implementers who were already passing. `edit_path(kind, model)` answers where
+  an edit of a kind lands and may answer `None` to decline that kind, which is
+  the honest answer for a format whose models are reduced from a running graph
+  (no authored staging model) but whose declared keys are hand-written files
+  (a `unique` test lands fine). `editing_surface()` declares the region those
+  paths must stay inside. The write gate now asks for the capability instead of
+  the class.
+
+  dbt reaches identical behaviour by the new route: `edit_path` returns the
+  scaffold convention reconcile hard-coded, and `editing_surface` returns the
+  project's configured model and macro paths, which is what containment checked
+  directly before.
+
+### Changed
+
+- **Containment validates an edit against the surface its own format declared**
+  ([#257], [#258]). `transform plan` validated every edit path against dbt's
+  `model_paths` whatever format produced it, so a second format placing a
+  declaration sidecar was refused at plan time even with the two gates above
+  resolved. Containment stays a safety property and stays mandatory; what moved
+  is who declares the surface. A format that declares none is unaffected and
+  validates against dbt's as before, and dbt's own path is unchanged, including
+  the root manifests allowed by name and the checks that a macro and a model each
+  live where dbt will find them. Escapes (absolute paths, `..`) are refused ahead
+  of the surface and are not a format's to permit.
+
+  The file an edit is pinned against now comes from the same view as the surface
+  it is checked in. Those two disagreeing is not a refusal but a quiet defect:
+  an existing file hashes as absent, the reviewable diff renders a one-line
+  change as a whole-file create, and the apply that follows reports a conflict on
+  a file nobody edited.
+
+- **A plan is applied through the format it was planned against** ([#257]).
+  `transform apply` wrote every plan with dbt's writer, which resolves each edit
+  under the dbt project and re-hashes what it finds on disk, so a plan a second
+  format could now store would still have been refused one stage later and the
+  `write_edits` that format implemented to reach tier 3 would never have been
+  called. Plans planned against dbt are written by dbt exactly as before.
+
+  `EditableProject.write_edits` now documents that its return has to report
+  `written` and `conflicts`, and `EditableProjectContract` asserts it. The apply
+  path reads both to tell a refusal from a write, and a result answering neither
+  fails in both directions at once: a plan recorded as applied that wrote
+  nothing, or a conflict that never reaches the person it was raised for. This is
+  a widening of the tier-3 contract, and a format returning
+  `dbt_project.ApplyResult` (or anything exposing those two) already satisfies
+  it.
+
+- **Reconcile matches the model a format declares, not dbt's spelling of it**
+  ([#258]). The `stg_` convention leaked through file contents as well as
+  through paths: the `unique` test edit looked for a model named `stg_<table>`
+  inside the YAML, so a format that placed its sidecar correctly and named its
+  model `orders` was missed one line before the edit was built, silently. The
+  model is now read from the file the format chose, which is the same string for
+  dbt, whose scaffold path is `models/staging/stg_<table>.yml`. A file declaring
+  no matching column entry now says so in `warnings` instead of skipping in
+  silence, which was indistinguishable from dex deciding the test was already
+  there.
+
 ## [1.6.3] - 2026-08-10
 
 ### Changed
--- a/packages/dex-core/src/exmergo_dex_core/adapters/conformance.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/conformance.py
@@ -41,6 +41,12 @@ declared join actually reach the engine, which is the whole reason a project is
 read on the explore path: those two arrive through ``definitions()`` and through no
 other channel.
 
+**If you reach tier 3 you also want** :class:`PlacingProjectContract`, because
+placement is what carries a proposal from reconcile to your write path, and its two
+methods have to agree with each other: a format that places an edit outside the
+surface it declares builds a proposal the plan store then refuses, which reads as
+dex declining rather than as the format contradicting itself.
+
 **If dex is going to build your format rather than be handed one**, mix
 :class:`ProjectFactoryContract` in front. Construction is a separate contract from
 the tiers for the same reason it is separate in storage, and a format that passes
@@ -74,6 +80,7 @@ __all__ = [
     "EditableProjectContract",
     "ExploreProjectContract",
     "MaintainProjectContract",
+    "PlacingProjectContract",
     "ProjectFactoryContract",
     "SemanticProjectContract",
 ]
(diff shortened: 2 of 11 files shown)


marcociav-exmergo:
@catincloudlabs let me know when this one is ready!

catincloudlabs:
@marcociav-exmergo Ready for review. Rebased onto `main` at 1554d24 (v1.6.3), all nine checks green, and the body's test section is corrected: the local failures I mentioned were a missing `dbt` CLI in my venv rather than anything in the tree, and the one real defect among them landed separately as #264.

The "draft" framing in the description is now the only stale thing, so let me replace it with the actual ask. The diff is complete and I am not holding anything back. What this needs from you is not a pass on the code but a call on the third gate:

**Does `plans.plan`'s containment check widen so a format can declare its own editing surface, or does it stay closed?**

That question is why I opened this rather than just filing the shape. Right now containment validates every edit against dbt's `model_paths` regardless of which format produced it, so a second format placing a declaration sidecar is refused even with #257 and #258 both resolved as proposed. I deliberately did not touch it, because widening it is a design decision about who owns the editing surface and this seam should not answer that alone.

Either answer merges. The end state is pinned by a strict xfail (`test_a_placing_format_receives_the_test_edit_in_its_own_sidecar`), so if you decide containment stays closed I flip that to a plain assertion of current behaviour, and if it widens the xfail turns green when the follow-up lands. The suite states which one is true either way.

Two things I would want your read on alongside that:

- `PlacingProject` sits beside `EditableProject` rather than on it, because the tiers are `runtime_checkable` and adding a method to tier 3 would demote every format that has not implemented it yet.
- `CHANGELOG.md` is intentionally unchecked. If you want this to land as it stands I will add the entry; I held off because an entry would claim the seam is finished when gate three says it is not.

If you would rather this wait on the #257 and #258 discussion, say so and I will flip it to draft. Otherwise it is yours to review.

marcociav-exmergo:
I'd say let's try to solve both https://github.com/exmergo/dex/issues/257 and https://github.com/exmergo/dex/issues/258 in one elegant PR, this one.
plans.plan's containment check widen so a format can declare its own editing surface.

And clearly, the default dbt containment check should stay intact after this PR!

That's my pov @catincloudlabs . If you agree, let's go forward!

catincloudlabs:
@marcociav-exmergo Agreed, on all three points, and that is what is on the branch.

Taking your clauses in order:

**Both issues in one PR.** They are one seam. #257 is the gate deciding *whether* a format may receive an edit; #258 is hard-coded paths deciding *where*. The paths reconcile builds are not filesystem paths, they are keys into the view the format's own `load()` returned, so a format that can answer "where does an edit of this kind for this table go" supplies both halves at once. `PlacingProject.edit_path(kind, model)` asks exactly that, and `reconcile`'s gate stops asking `isinstance(editable, DbtProject)`.

It sits **beside** `EditableProject`, not on it. The tiers are `runtime_checkable`, so a method added to tier 3 would silently demote every format that has not implemented it yet — `tier_of` starts answering 2 where it answered 3, and the write path closes for exactly the implementers who were already passing. Beside it, a format that has not implemented placement is "tier 3, placement unknown", which is true and is what the caller warns about.

**`plans.plan` containment widens.** `editing_surface()` is where a format declares the region its edits may land in, and `contained_key` matches by path segment rather than resolving anything, since these are keys and not paths. `declarations` admits `declarations/orders.yml` and not `declarations_backup/orders.yml`. Absolute paths and `..` are refused *ahead* of the surface, because an escape is not a format's to permit — a format listing one would be declaring the rest of the filesystem.

**And dbt's stays intact.** This is the clause I most wanted to get right, so here is the mechanism rather than the assurance.

The branch is keyed on the **view's shape**, not on the project's class:

```python
dbt_shaped = isinstance(view, DbtProjectView)
```

A format handing back a `DbtProjectView` is offering dbt's surface and gets dbt's checks — `contained_path` against `view.model_paths` / `view.macro_paths`, plus all three kind-and-location rules (a `macro_sql` outside the macro paths, a non-macro inside them, a root-config edit pointed at the wrong manifest). None of that moved. Asking the *view* is also what keeps this from quietly re-introducing the `isinstance(project, DbtProject)` gate the seam exists to remove.

A format declaring no surface is not routed through the new seam at all: it takes the pre-seam path, loading the dbt project and validating against dbt's paths. Deliberate fallback, not an oversight — such a format cannot place an edit either, so the edits reaching there are dbt-shaped and dbt's view is the right thing to pin them against.

What proves it is `test_the_no_format_fallback_answers_exactly_what_dbt_answers`, which asserts `_placed(None, kind, model) == DbtProject().edit_path(kind, model)` **for every `EditKind`** — and then asserts the two kinds reconcile actually proposes still resolve to real paths, so the parity is not both sides having gone silent. `test_the_format_declaring_its_surface_does_not_widen_it` is the negative arm on the other side.

Rebased on `main` at `1554d24`, head `28bd412`, nine checks green. One thing worth flagging since it is adjacent: I have just widened #274 to a second site, so if you take these in either order there is no interaction — they touch different modules.

marcociav-exmergo (approved):
LGTM! Gonna merge this one while we gear up towards a new release

PR #264, opened 2026-08-09
A dev-target refusal should not mix path separators mid-sentence

catincloudlabs:
## What this changes

`transform`'s dev-target refusal names the config file beside the profile it disagrees with. The first half of that sentence is built from a literal and always used forward slashes; the second came from `_relative`, which returned the platform separator. So on a Windows checkout one sentence read:

```
.dex/config.yml and analytics\profiles.yml disagree about the dev target:
```

Both halves are repo-relative labels a reader matches against paths written in configuration, not paths anything opens, so both use forward slashes now. `_relative` has a single call site, so the change is contained to this message.

The other half of the same bug is that `test_retargeted_dev_database_is_refused_naming_both_values` asserts `"analytics/profiles.yml" in message`, so it passes on your Linux CI and fails on a Windows checkout. I left the assertion as written rather than loosening it to accept either separator, on the grounds that the message really should not vary by platform here. If you would rather keep the native separator and relax the test instead, say so and I will flip it.

Found while running the suite on Windows against `81d57d2`. Mentioned in passing on #263; sending it separately as offered there.

## Related issues

None. Unfiled, and small enough that a PR seemed better than an issue first.

## Checklist

On the local run: 1868 passed, 34 failed, and all 34 need the `dbt` CLI on PATH, which this box does not have. That count was 35 before this change and the one that went away is the assertion above. No safety path is touched; the box is ticked because the refusal itself is a guardrail and it still refuses, with the same two values named.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,18 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   multi-column grain to express. `DbtProject` implements all three in
   `tests/adapters/test_project_parity.py`, so the shipped format is held to them too.
 
+### Fixed
+
+- **A dev-target refusal no longer mixes path separators mid-sentence on
+  Windows.** The message names the config file beside the profile it disagrees
+  with, and the first half is built from a literal (`.dex/config.yml`) while the
+  second took the platform separator, so a Windows reader got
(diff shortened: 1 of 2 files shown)


marcociav-exmergo (approved):
LGTM!

PR #272, opened 2026-08-10
SCRUM-1055-explore profile flags a candidate-key column that mixes value shapes

### LINKED ISSUE #205: `explore profile`: flag a heterogeneous key column
**Depends on:** nothing

### Problem

A string id column carrying two different kinds of value (numeric ids and opaque
hashes, or two id schemes from a migration) profiles as an ordinary unique string
column. Nothing distinguishes it from a homogeneous one.

### Why it matters

This is the shape that survives a partial migration or a merged upstream, and the
failure mode is specific and severe: downstream code casts the column to an
integer, or filters to what parses, and **silently drops exactly the rows that came
from the other scheme**. The row count falls by a few percent, every test still
passes, and the loss is invisible until someone reconciles a total.

### Proposal

For a key-shaped column (a candidate key, or one participating in a detected
relationship), report the fraction of values matching each of a small set of
shapes: all-numeric, fixed-length hexadecimal, UUID, and other. When more than one
shape holds a meaningful share, emit a data-quality observation naming the shapes
and their fractions, and warning that a cast or a numeric comparison will drop one
group.

Fractions only. No value, and no example of either shape.

### Acceptance

- A column of 90% numeric and 10% md5-shaped values produces the observation with
  both fractions.
- A homogeneous numeric or homogeneous UUID column produces none.
- A free-text column is not treated as key-shaped and produces none.
- The observation names the consequence (a cast drops one group), not only the
  statistic.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/205
New key_shape_stats request set (string-only, non-PII): ColumnAggregate gains uuid_string_fraction, hex_string_fraction, hex_string_min_length, hex_string_max_length, computed in the same aggregate batch across DuckDB, BigQuery, Snowflake, Databricks, Redshift, Postgres. No new adapter capability, no extra scan cost.
Every non-null value partitions into exactly one of numeric / UUID / hex / other by construction (the hex bucket explicitly excludes anything the numeric pattern already claimed), so #204's existing numeric_string_fraction is reused unchanged for the numeric bucket rather than recomputed.
A data_quality note fires only for a candidate-key column (single-column proven unique, or a composite member) when 2+ shapes each clear a 5% minority-share floor, naming the fractions, the hex length (md5/sha1/sha256 recognized by length), and the consequence: a cast or numeric comparison will silently drop the non-numeric group(s).
Homogeneous keys and non-key free-text columns correctly produce no note.
before: 
image
after:
image


marcociav-exmergo (approved):
LGTM! Thanks @emnarahmouni-exmergo

Issue #226, opened 2026-08-05
`maintain verify`: report row loss and fanout against a model's driving parent


marcociav-exmergo:
**Depends on:** #224. Overlaps **#164**.

### Problem

A model that quietly returns fewer rows than the parent it is built from is the
single most common silent defect in a dbt project. An inner join where a left join
was meant, a filter that excludes NULLs by accident, a deduplication that keys on
the wrong column: each drops rows, none raises an error, and uniqueness and
not-null tests all still pass.

Fanout is the same defect mirrored: a join on a non-unique key multiplies rows, and
every downstream sum is inflated.

**#164** reports that `maintain snapshot` already fingerprints transform-layer
counts and never diffs them, which is the drift-framed half of the same idea.

### Proposal

For each selected model, identify its **driving parent** (the relation in the FROM
clause, as distinct from joined relations) from the compiled SQL, compare row
counts, and report:

- `row_loss` when the model has materially fewer rows than its driving parent and
  the SQL contains no filter, aggregation, or deduplication that accounts for it
- `row_fanout` when it has materially more, naming the join and the key

The qualifier is what keeps this from firing on every aggregate. A model with a
`group by`, a `where`, a `qualify`, or a `distinct` has an explicit reason to have
fewer rows and should not be reported unless the loss is far larger than the filter
plausibly explains. Getting this threshold right is the substance of the issue, and
it should start conservative: report only what is unambiguous, since a detector
that cries wolf on every aggregate will be turned off.

Row counts are free metadata on most connectors, so most of this is free.

### Acceptance

- A model whose only change from its parent is an inner join, and which loses 20%
  of rows, is reported with the join named.
- A `group by` model that legitimately has fewer rows is not reported.
- A model that fans out on a non-unique join key is reported, and the key is named.
- The finding states the counts, so a reader can judge the threshold themselves.
- Coordinate with #164 so the drift and no-baseline paths share one implementation.

PR #273, opened 2026-08-10
SCRUM-1056: `explore profile` reports temporal continuity for date and timestamp   columns

### LINKED ISSUE #206: `explore profile`: report temporal continuity for date and timestamp columns
**Depends on:** nothing. **Blocks:** #219.

### Problem

A temporal column reports a null fraction, a distinct count, and a min and max.
Nothing says whether the range is **continuous**. A table covering 3,000 days with
2,900 distinct days looks identical to one covering 3,000 days with 3,000.

### Why it matters

Two common failures share this signature and neither shows up in a row count:

- a **half-failed load**, where one day or one partition is missing from an
  otherwise healthy table. The total row count moves by a fraction of a percent and
  the drift threshold never fires, but every daily aggregate for that period is
  wrong.
- a **date-spine defect**, where a rollup is built from observed dates rather than
  a calendar, so periods with no activity vanish from the output instead of
  appearing as zero.

Row count measures the total; continuity measures the shape, and the shape is the
better signal.

### Proposal

For temporal columns, add to the profile:

- `span`, the range between min and max at a detected granularity (day is the
  default; detect month or hour where the data is clearly at that grain)
- `distinct_periods`
- `missing_periods`, and the largest single gap

Computed as `count(distinct <truncated>)` against the span, both bounds of which
are already fetched for temporal columns.

Where a model's grain is a date, a non-zero `missing_periods` is a candidate
finding for #226's sweep, not only a profile statistic.

### Acceptance

- A daily table missing 100 days reports 100 missing periods and the largest gap.
- A complete daily table reports zero.
- A column that is genuinely sparse by nature (an event timestamp on a rare event)
  reports the numbers without being characterized as broken. The statistic is
  neutral; the interpretation belongs to whoever asked.
- Granularity detection is reported, so a caller can see that "missing periods" was
  counted in days rather than hours.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/206
New temporal_stats request set (non-PII date/timestamp columns): ColumnAggregate gains two alignment fractions (day_aligned_fraction/month_aligned_fraction, decide the reported granularity) plus three parameterized (distinct_periods, largest_gap) pairs for day/month/hour, all computed in the same aggregate batch across DuckDB, BigQuery, Snowflake, Databricks, Redshift, Postgres.
ColumnProfile gains temporal_granularity/temporal_span/temporal_distinct_periods/temporal_missing_periods/temporal_largest_gap, decided in explore/profile.py: granularity picks hour when day-truncation would lose real sub-day variation, month when values sit cleanly on month boundaries, day otherwise.
largest_gap uses a correlated scalar subquery with LAG() OVER (ORDER BY period) spliced into the existing flat aggregate SELECT (the same pattern the composite-key probe already uses) rather than a new row-returning adapter capability zero extra round trip, one bounded scalar per column. Confirmed with the user as the preferred design over a new uncapped row-returning call.
Hour granularity is skipped for a bare DATE column (nothing to truncate to an hour, and BigQuery's DATE_TRUNC doesn't even accept that unit) via a new temporal_units_for/is_date_only_type classifier, applied uniformly across all 6 adapters.
BigQuery needed type-aware truncation/diff functions (DATE_TRUNC/TIMESTAMP_TRUNC/DATETIME_TRUNC, reversed argument order vs. every other dialect); Postgres has no native DATEDIFF, so its diff is EXTRACT(EPOCH FROM ...) arithmetic for day/hour and explicit year/month field subtraction for month.
The statistic is neutral by design: a genuinely sparse column (an event timestamp on a rare event) reports its numbers without being flagged as broken interpretation is left to the caller, explicitly including a future drift-sweep detector (#226) this only lands the raw statistics for.
The issue : 
image
After fixing : 
image


marcociav-exmergo (approved):
LGTM! Thanks @emnarahmouni-exmergo

PR #274, opened 2026-08-10
`explore query` and `explore cluster` decide from the cache without opening a connection

catincloudlabs:
## What this changes

`explore query` opened a warehouse connection before the PII firewall ran. One line:

```python
named = list(dict.fromkeys(name for names in reads.values() for name in names))
if named:
    shared.adapter = engine._adapter("explore query")   # unguarded
    gap = _object_gap(shared.adapter, cache, named)
```

`named` is non-empty for essentially every real query, so this is the ordinary path rather than an edge case. A caller holding a profiled cache and no reachable connector now gets a connector error where 1.6.2 returned a refusal.

**Why that is more than an inconvenience.** The guard decides from cached PII flags. It needs no warehouse to reach its answer, so gating it on a connection converts a policy decision into a connectivity one, and it fails in the direction that matters: the firewall is closed in exactly the environments that cannot bill for a mistake. CI, an offline box, anything holding only a cache.

**The fix is a consistency fix rather than a new tolerance.** `_object_gap` already declines to conclude anything from a connection it could not read:

```python
try:
    _meta, columns = adapter.table_metadata(identifier)
except Exception:  # noqa: S112 -- an unreadable signature settles nothing
    continue
```

That tolerance exists one level below the acquisition and was missing at it. "No adapter at all" is the same category of doubt as "unreadable signature", and it now falls through the same way.

**Nothing is swallowed, and this is the half worth checking.** Only a refusal returns without a connection. A statement that *passes* the guard reaches `shared.adapter or engine._adapter(...)` below and raises there, which is where a caller about to run SQL wants to hear that the warehouse is unreachable. The second test asserts exactly that.

Drift detection is untouched: wherever a connection exists, `_object_gap` runs as before and #269 keeps everything it was built for. The change is only about what happens when there is no connection to be had.

## The second site: `explore cluster`

Added after the first review pass, by asking what *else* #269 changed rather than
what its symptom was. #269's own title names two commands, and `cluster` took the
identical unguarded acquisition:

```python
if auto:
    adapter = engine._adapter("explore cluster")   # unguarded
    gap = _object_gap(adapter, cache, [obj])
```

It sits above the two things `cluster` decides from the cache alone, and both are
refusals: `no exploration cache yet` and `'<obj>' is not a profiled object`. At
1.6.2 both returned before any connection; at 1.6.3 both are connector errors.
`auto_profile` defaults to `true`, so this is the ordinary path and not an opt-in
one, which leaves `--no-auto-profile` as the only way to reach either refusal
offline.

**The claim here is narrower than the one above, and deliberately so.** This half
is not the PII firewall. `cluster` drops PII-flagged columns during feature
selection, but that only matters for an object that *is* profiled, which needs a
connection anyway to build its sample. What regressed is a prerequisite refusal,
not a policy one. Both are the same defect in the same change, so they are fixed
together, but only the query half closes a safety hole.

Measured rather than read, on the installed wheels:

| | 1.6.2 | 1.6.3 |
|---|---|---|
| `explore cluster` on an object not in the cache, warehouse unreachable | `reason: prerequisite` | connector failure |

The fix is the query path's fall-through applied unchanged, and its limit is the
same: a profiled object still reaches the opener at the bottom of `cluster` and
raises there. `_unreachable_warehouse` moves to `conftest.py` as
`unreachable_warehouse`, since two test files now need it.

## Scope

The batching rework is not implicated; the probe simply landed in front of the guard. I have deliberately **not** reordered `inspect_query` ahead of `_object_gap`, which would also turn the symptom green but at the cost of the drift check that probe exists to perform.

## How this was found

Downstream, on a bump to 1.6.3. `catincloud-labs-vps` carries a control asserting that a PII refusal is free and offline, built on a synthetic cache and deliberately no credentials, and it went red on the bump. The relevant part of its contract:

> `explore/commands.py::query` reads the cache, masks overridden columns, and calls `inspect_query` (the firewall) **before** `engine._adapter(...)` ever opens a connection. A refusal returns without credentials, network, or billing.

Reproduced two ways, which is what ruled out a local environment: with no credentials it asked for ADC, and on CI with no BigQuery extra it asked for the connector. Two messages, one cause.

## Verification

Proven red before green, since a guard that cannot fail is the thing this repo keeps finding:

- With the test in place and the acquisition unguarded, `test_a_pii_refusal_needs_no_connection` **fails**: `assert 'query refused' in 'the connector extra is not installed'`. That is the regression, reproduced in-tree.
- With the fix, both new tests pass.
- The cluster half was proven the same way: `test_a_cluster_prerequisite_refusal_needs_no_connection` fails as `assert 'not a profiled object' in 'the connector extra is not installed'` with the acquisition unguarded, and the guard was then **reviewed by mutation**: deleting it turns the test red again and restoring it green.
- Full suite on Windows, all CI extras synced: **2070 passed, 65 skipped, zero failures**. Re-measured after rebasing onto `d61e4db`; the count moved from 2056 because #263 landed in between, and none of the arrivals fail here.
- `ruff check` and `ruff format --check` clean on both touched files.

The control test passes both with and without the fix, which is intended: it guards the direction this change must not break.

## Related

Refs #269, which added the probe. No issue filed; the defect is against the code's own stated tolerance rather than a design question, so it seemed shorter to bring the fix.

## Checklist

On the safety box: this is a safety path and it moves toward the spine rather than across it. The firewall now refuses in cases where it previously could not run, and no case that refused before stops refusing. The change adds no new way for flagged data to be surfaced, since the only behaviour that changes is whether a refusal is reachable without a connection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,17 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   reaches the same opener afterwards and raises there, which is where a caller
   about to run SQL wants to hear it.
 
+- **`explore cluster` refuses from the cache again without a connection.** The
+  same probe was added to `cluster` in the same change, in front of the two
+  things that command decides from the cache alone: that there is no cache at
+  all, and that the named object is not in it. Both refusals sat below the
+  acquisition, so both became connector errors. `auto_profile` defaults to
+  `true`, which made this the ordinary path rather than an opt-in one, and
+  `--no-auto-profile` the only remaining way to reach either refusal offline.
+  The same fall-through applies, with the same limit: an object that *is*
+  profiled still needs a connection to build its sample, so it reaches the
+  opener at the bottom of `cluster` and raises there.
+
 ## [1.6.3] - 2026-08-10
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,26 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   silence, which was indistinguishable from dex deciding the test was already
   there.
 
+### Fixed
+
+- **The PII firewall decides again without a warehouse connection.** Since 1.6.3,
+  `explore query` opened a connection before the guard ran: the object-gap probe
+  added with the auto-profiling work took an already-open adapter, and the
+  acquisition ahead of it was unguarded. Every query naming a relation therefore
+  needed a reachable warehouse before the firewall could refuse anything, so a
+  caller holding a profiled cache and no connector got a connector error where
+  1.6.2 returned a refusal. That reaches further than an inconvenience: the guard
+  reads cached PII flags and needs no warehouse to decide, so gating it on a
+  connection turns a policy decision into a connectivity one and closes the
+  firewall in exactly the offline environments that cannot bill for a mistake.
+
+  The acquisition is now as tolerant as the use one level below it, where an
+  unreadable column signature already "settles nothing" and falls through. A
+  failed open settles nothing either. Drift detection is unchanged wherever a
+  connection exists, and nothing is swallowed: a statement that passes the guard
+  reaches the same opener afterwards and raises there, which is where a caller
+  about to run SQL wants to hear it.
+
 ## [1.6.3] - 2026-08-10
 
 ### Changed
--- a/packages/dex-core/src/exmergo_dex_core/explore/commands.py
+++ b/packages/dex-core/src/exmergo_dex_core/explore/commands.py
@@ -1828,7 +1828,22 @@ def cluster(
     warnings: list[str] = []
     profiled_names: list[str] = []
     if auto:
-        adapter = engine._adapter("explore cluster")
+        try:
+            adapter = engine._adapter("explore cluster")
+        except DexError:
+            # The same tolerance the query path applies, at the second site the
+            # object-gap probe was added to. `cluster` decides two things from
+            # the cache alone -- "there is no cache" and "that object is not in
+            # it" -- and both refusals sit below this acquisition, so gating it
+            # here made them unreachable wherever no connector is installed.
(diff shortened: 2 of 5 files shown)


catincloudlabs:
Scope note, not a nudge: this PR now covers **two** sites rather than one, and the title changed to match.

I went back and asked what else #269 changed rather than what its symptom was. Its own title names two commands, and `explore cluster` took the identical unguarded acquisition in front of the same probe — above the two things `cluster` decides from the cache alone (`no exploration cache yet`, and `'<obj>' is not a profiled object`). `auto_profile` defaults to `true`, so both refusals became connector errors on the ordinary path, leaving `--no-auto-profile` as the only way to reach either one offline.

Measured on the installed wheels rather than read off the diff:

| | 1.6.2 | 1.6.3 |
|---|---|---|
| `explore cluster` on an object not in the cache, warehouse unreachable | `reason: prerequisite` | connector failure |

**One thing worth being precise about, because it changes how you weigh this half.** The cluster site is *not* the PII firewall. `cluster` does drop PII-flagged columns during feature selection, but that only matters for an object that is already profiled, which needs a connection anyway to build its sample. So what regressed there is a *prerequisite* refusal, not a policy one. Same defect, same change, same fix — but only the `query` half closes a safety hole, and I did not want the shared title to imply otherwise.

The guard is the query path's fall-through applied unchanged, reviewed by mutation: deleting it turns the new test red with the connector error, restoring it green. `_unreachable_warehouse` moved to `conftest.py` since two files need it now. Full suite with every CI extra synced: 2056 passed, 65 skipped, zero failures; ruff clean.

Happy to split the cluster half back out into its own PR if you would rather keep this one scoped to the firewall — it is one commit (`dd2991c`) and lifts cleanly.

marcociav-exmergo (approved):
LGTM! Thanks @catincloudlabs. Merging

Issue #276, opened 2026-08-11
`data.bytes_billed` is present on `transform build` and absent on `maintain check`


marcociav-exmergo:
**Environment:** dex 1.6.3, BigQuery connector.

### Problem

The same concept, "what did this command bill", lives in different envelope keys
depending on the command:

| Command | `data.bytes_billed` | `data.spend.bytes_billed` |
|---|---|---|
| `transform build` | 4750049280 | 4750049280 |
| `maintain check` | **absent** | 891289600 |

### Why it matters

A caller that reads `data.bytes_billed` and defaults a miss to zero reports that
`maintain check` was free. It was not; it had just spent 0.89 GB.

This is not hypothetical. It happened in a real session: the wrong figure was
reported to a human, and it survived into the first draft of a written cost
report before being caught by reconciling the envelopes against
`.dex/spend.jsonl`.

Silently under-reporting spend is the one class of envelope defect that
undermines the cost-governance guarantee rather than merely annoying the caller.
An agent reading one key and getting nothing has no way to distinguish "this
cost nothing" from "look in the other key", and both readings are plausible.

#136 is the same family (an envelope whose shape misrepresents what happened) on
a different command, which suggests the fix belongs at the envelope contract
rather than being patched per command.

### Proposal

Make `data.spend` the single canonical location, present on every billed command,
carrying `bytes_billed` and `session_spent_today`.

Then either populate `data.bytes_billed` on every command that bills, or remove
it entirely. Do not leave it present-sometimes: a key that exists on one command
and not another is worse than a key that never exists, because its absence reads
as a value.

### Acceptance

- Every command that can bill returns `data.spend.bytes_billed`.
- No command returns a `data.*` spend key that another billed command omits.
- A contract test asserts key parity across `transform build`, `maintain check`,
  `explore query`, `explore map`, and `explore profile`.

PR #284, opened 2026-08-11
Fix null entry on transform build spend-ledger rows

### LINKED ISSUE #277: Spend-ledger rows carry `entry: "settlement"` on some commands and `null` on `transform build`

### Problem

`.dex/spend.jsonl` mixes record shapes. Settlement rows written by
`maintain check` and `explore query` carry `"entry": "settlement"`. Rows written
by `transform build` carry `"entry": null`, while still holding a correct
`billed_bytes`.

Observed distribution of `(command, entry)` in one repo's ledger:

```
('explore map',     None)          1567
('maintain check',  'settlement')   170
('maintain check',  None)           117
('transform build', None)            20
('explore query',   'settlement')     2
```

### Why it matters

The ledger is a documented artifact that external tooling is invited to read,
and `entry` looks exactly like the field you would filter on to get settled
spend.

Doing so drops **every `transform build` row**, which is the single largest
spender in a normal session. In one day's ledger that filter returned 0.90 GB
against a true 6.60 GB: an 85% undercount, concentrated precisely on the command
most worth watching.

dex's own `session_spent_today` accumulates both shapes correctly, so the
accounting is right. The defect is in the artifact's self-describability.

### Proposal

Give every row an explicit `entry` drawn from a closed vocabulary
(`reservation`, `settlement`, `release`), and document that vocabulary alongside
the ledger format. If `transform build` rows are settlements, label them so.

### Acceptance

- No row written to `.dex/spend.jsonl` has a null `entry`.
- Summing `billed_bytes` over `entry == "settlement"` reproduces
  `session_spent_today` for the same UTC day.

lesbass:
Fixes #277.

## Problem

`transform build` writes its spend directly to `.dex/spend.jsonl` through `_record_build_spend`, bypassing the cost gate's `_append` which always includes `"entry": kind`. As a result, every `transform build` row carried `"entry": null` while `maintain check` and `explore query` rows correctly carried `"entry": "settlement"`.

Filtering on `entry == "settlement"` therefore drops every `transform build` row, which is the single largest spender in a normal session.

## Fix

Add `"entry": "settlement"` to the dict passed to `append_spend_log` in `_record_build_spend` so all settlement rows are self-describing.

## Files changed

- `packages/dex-core/src/exmergo_dex_core/transform/commands.py`: +1 line

## Verification

```
uv run pytest tests/transform/test_build.py tests/transform/test_apply.py tests/transform/test_deps.py tests/transform/test_init.py tests/transform/test_parse.py tests/transform/test_validate.py -x
```
123 passed, 7 skipped.

### DIFF
--- a/packages/dex-core/src/exmergo_dex_core/transform/commands.py
+++ b/packages/dex-core/src/exmergo_dex_core/transform/commands.py
@@ -767,6 +767,7 @@ def _record_build_spend(
             "at": datetime.now(UTC).isoformat(),
             "connector": connector,
             "command": "transform build",
+            "entry": "settlement",
             field: float(billed),
             "job_id": None,
             "statement_sha256": None,


lesbass:
Hi @exmergo, just checking in on this PR. It has been open for about 2 weeks now.

The fix is a single-line change that addresses a data integrity issue where `transform build` spend-ledger rows carry `entry: null` instead of `entry: "settlement"`. This breaks filtering on `entry == "settlement"` which excludes all `transform build` rows.

Let me know if you need any changes or if there are concerns about the approach. Happy to iterate if needed.

marcociav-exmergo (changes requested):
@lesbass thanks for this.
Can you add a small CHANGELOG.md entry?
It's our standard practice, even though the fix is small.

catincloudlabs:
Thanks for the fix — this closes #277's immediate symptom (build rows with no
`entry`), and the placement in `_record_build_spend` is right.

One shape concern worth deciding deliberately before this merges: everywhere
else in the ledger, `entry` and `reservation_id` travel together —
`CostGate._append` stamps both on every row it writes. This change introduces
a third row shape: gate rows carry `entry` + `reservation_id`, pre-fix build
rows carry neither, and post-fix build rows carry `entry: "settlement"` with
no `reservation_id`. A reader that keys settlements by `reservation_id` (the
natural join if settlement-keyed reconciliation lands — #276 is open in that
area) would silently skip or mis-join these rows.

Two ways to keep the shape closed — dealer's choice:

1. Route the build's write through the gate's append path, so the stamping
   logic stays in one place — though `_record_build_spend`'s docstring is
   explicit that a build settles outside any gate, so this may be a bigger
   change than #277 warrants.
2. Keep the direct write but stamp `reservation_id: None` explicitly, so the
   row *declares* it settled outside a reservation rather than leaving the key
   absent — absent and null are different claims to a reader.

Related naming question, since a build "settles outside any gate": reusing the
gate's `"settlement"` kind makes these rows indistinguishable from gate
settlements except by the missing rid. A distinct kind (e.g. `"build"`) would
keep `entry` meaning "which writer produced this row" instead of overloading
it.

What I'd do regardless: add a small regression test pinning the full row
shape `_record_build_spend` writes — nothing currently asserts it, and it
would keep the next writer honest.

PR #287, opened 2026-08-11
SCRUM-1057-flag joins with catastrophic orphan rates as findings

### LINKED ISSUE #207: A join with no value overlap is only a demoted confidence, not a finding
**Depends on:** nothing. **Blocks:** #228.

### Problem

`explore relationships --verify` measures orphan fractions correctly and then says
nothing about the ones that are catastrophic. Against the public ADE-bench
`intercom.duckdb`:

```
company_tag_history_data.tag_id -> conversation_tag_history_data.tag_id
    conf=0.45 verified=True orphan=1.0
contact_tag_history_data.contact_id -> conversation_contact_history_data.contact_id
    conf=0.4  verified=True orphan=1.0
conversation_part_history_data.conversation_id -> conversation_contact_history_data.conversation_id
    conf=0.3  verified=True orphan=1.0
notes:
  - inference examined 24 id-shaped column(s) across 13 profiled object(s)
  - verified 9 inferred join(s) with aggregate overlap probes
```

Verified on 1.5.1. Three of nine edges have **complete** orphan rates: the two
columns share a name and share no values at all. The only consequence is a lower
`confidence`, and nothing in `warnings`, `data_quality`, or `notes` says so.

### Why it matters

An orphan fraction of 1.0 is not a weak signal, it is a strong one pointing the
other way: these two columns are named alike and are not the same key. That is
precisely the finding a caller needs, and it is being communicated as the absence
of a finding. A reader skimming nine edges has to notice that a number they may not
have looked at means "never join these".

The failure this leads to is specific: a model joins on the same-named column,
returns rows whose parent-side attributes are all NULL, and looks like it worked.

### Proposal

- Emit a finding for each verified edge whose orphan fraction is at or above a
  high threshold, naming both sides and the measured fraction, and stating that the
  shared column name is not evidence of a shared key.
- Add a summary line to `notes` when any such edge exists, so a caller reading only
  the notes still sees it.
- Consider a matching data-quality observation on the child dataset, so it survives
  into the cache and is visible to anything reading profiles rather than
  relationships.

No behavior change to inference or to the confidence arithmetic. This is about
saying what was already measured.

### Acceptance

- A verified edge at orphan 1.0 produces a finding naming both sides.
- A verified edge at orphan 0.0 produces none.
- An unverified inferred edge produces none, since nothing was measured.
- The threshold is a named constant with a stated rationale.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/207
New _ORPHAN_FINDING_THRESHOLD = 0.9 and orphan_findings() in explore/relationships.py: returns (Relationship, text) pairs for every verified inferred join whose measured orphan_fraction clears the threshold, naming both sides (from_dataset.column -> to_dataset.column) and the fraction, mirroring declared_relationships's existing arrow-label convention.
Wired into both command handlers that call verify_relationships (relationships() and map()): the finding text is appended to notes (visible in the command's own envelope) and mirrored onto the child dataset's data_quality (persisted into the cache, so it's visible to anything reading profiles later, not just this one command's output).
Declared joins and anything not verified never qualify (nothing was measured). 90% is set well above the existing 20% confidence-demotion tier, so this fires only for the catastrophic case the issue describes, not every weakened guess.
No change to inference logic or confidence arithmetic purely surfacing what verify_relationships already measures.

The issue: 
image
The fix:
image


marcociav-exmergo (approved):
Yeah this is good, we weren't classifying orphans properly. LGTM!

PR #298, opened 2026-08-11
SCRUM-1058-Entity matching strips table-name suffixes and layer prefixes

### LINKED ISSUE #208: Entity matching is defeated by table-name suffixes and layer prefixes
**Depends on:** nothing. **Blocks:** #220.

### Problem

Relationship inference matches a foreign key to a parent by entity name, and it
works well when the parent table is named for the entity. Against the public
ADE-bench `analytics_engineering.duckdb` it produces sixteen edges, including
`inventory_transactions.product_id -> products.id` at confidence 0.95 with a
measured orphan fraction of 0.0.

It fails when the parent table's name carries a suffix. Against
`intercom.duckdb`, the edge
`conversation_part_history_data.conversation_id -> conversation_history_data.id`
is **never proposed**, because singularizing `conversation_id` yields
`conversation` and the parent is `conversation_history_data`. Instead, inference
proposes a child-to-child edge between two tables that both merely *hold* a
`conversation_id`, and that edge measures at orphan 1.0.

### Why it matters

These suffixes and prefixes are not exotic. They are the default output of
change-data-capture history modes, of landing-zone conventions
(`_data`, `_raw`, `_stg`), of versioned tables (`_v2`), and of layered warehouse
naming (`stg_`, `dim_`, `fct_`, `int_`). A warehouse with messy naming is the
exact case where a caller most needs the relationship map, and it is the case the
matcher is weakest on.

### Proposal

During entity matching, compare the normalized entity name against the parent's
name with a small, ordered set of affixes stripped:

- suffixes: `_history`, `_data`, `_raw`, `_snapshot`, `_current`, `_v<n>`
- prefixes: `stg_`, `src_`, `raw_`, `dim_`, `fct_`, `fact_`, `int_`, `base_`

A match that required stripping scores lower than an exact one, so the ranking
still prefers the unambiguous case, and the note explaining what inference examined
should say when a match was made this way.

The affix list should be configurable, since house conventions vary, and it should
be small by default rather than exhaustive.

### Acceptance

- `conversation_id -> conversation_history_data.id` is proposed on the intercom
  file, and survives `--verify`.
- Existing exact matches keep their current confidences, so nothing already working
  is re-ranked.
- A stripped match ranks below an exact match to the same key.
- Where stripping produces two candidate parents, both are proposed and
  `--verify` decides, rather than the matcher picking one.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/208
Relationship inference matched a foreign key to a parent by comparing the singularized column stem to the parent table's name (layer prefix stripped), so a CDC history suffix, a landing-zone suffix (_data/_raw/_stg), or a versioned table name defeated the match entirely.
Adds a new, lower-confidence match tier that strips a small configurable set of prefixes/suffixes (entity_affixes in .dex/config.yml) plus a table-version suffix, and retries the entity-name comparison.
A stripped match never outranks an exact match to the same key, and explore relationships/explore map's notes say when a join relied on it. Where stripping resolves more than one candidate parent, both are proposed and --verify decides.
Before : 
image
After: 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,33 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   silence, which was indistinguishable from dex deciding the test was already
   there.
 
+### Fixed
+
+- **Entity matching survives table-name suffixes and layer prefixes**
+  ([#208]). A foreign key matched a parent by comparing the singularized
+  column stem against the parent table's name, layer prefix stripped, so a
+  parent named plainly for its entity matched
+  (`inventory_transactions.product_id -> products.id`), but a parent whose
+  name also carried a CDC history suffix or a landing-zone suffix did not:
+  singularizing `conversation_id` yields `conversation`, and nothing
+  recovered `conversation` from `conversation_history_data`, so inference
+  fell back to a child-to-child edge between two tables that merely shared
+  the FK's name, measured at an orphan fraction of 1.0.
+
+  Entity matching now also compares against the parent's name with a small,
+  ordered set of prefixes (`stg_`, `src_`, `raw_`, `dim_`, `fct_`, `fact_`,
+  `int_`, `base_`) and suffixes (`_history`, `_data`, `_raw`, `_snapshot`,
+  `_current`) stripped, repeated until none remain, plus a table-version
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
LGTM!

PR #300, opened 2026-08-12
SCRUM-1064-transform plan warns on authored/declared column drift

### LINKED ISSUE #214: `transform plan`: warn when authored columns diverge from the declared contract
**Depends on:** nothing. **Blocks:** #230, #223.

### Problem

A model with a `schema.yml` entry declaring its columns can be authored with a
different set, and `transform plan` does not compare them. The declaration is the
closest thing a dbt project has to a contract, and it is not consulted at the one
moment it would be cheapest to check.

### Proposal

At plan time, when the model being authored has declared columns, compare the
authored SELECT list to the declared set and warn on both directions: declared but
not produced, produced but not declared.

A **warning**, never a refusal. The declaration is frequently the thing that is out
of date, and the caller is often deliberately changing the model. What matters is
that the divergence is visible in the same envelope as the diff, at the moment the
caller can act on it.

Where the SELECT list cannot be resolved statically (a `select *`, a macro-generated
list), say so rather than guessing.

### Acceptance

- Authoring a model that omits a declared column warns, naming it.
- Authoring one that adds an undeclared column warns, naming it.
- A model with no declared columns produces no warning.
- A `select *` reports that the comparison could not be made.
- The plan still validates and stores. This never blocks.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/214

transform plan compared nothing against a model's declared columns: in schema.yml, so an authored SELECT list could silently diverge from the project's own contract.
Adds column_contract_warnings() in plans.py, run for every MODEL_SQL upsert actually in the plan (not the whole project): warns when a declared column isn't produced, and when a produced column isn't declared.
Declared columns are read with this plan's own SCHEMA_YML edits overlaid on the project, so a model and its doc edited together in one plan are compared against each other, not a stale on-disk file.
Where the SELECT list can't be resolved statically (select *, t.*, or an unaliased macro standing in for a column), the warning says so instead of guessing; an aliased macro call still resolves by its alias.
Always a warning, never a refusal: the plan still validates and stores regardless.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,28 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   project's configured model and macro paths, which is what containment checked
   directly before.
 
+- **`transform plan` warns when a model's authored columns diverge from its
+  declared schema.yml contract** ([#214]). A model's schema.yml entry is the
+  closest thing a dbt project has to a column contract, and plan time is the
+  cheapest moment to check it, but nothing did: a model authored with a
+  different column set than its declaration passed silently. `transform plan`
+  now compares the authored SELECT list against the declared columns for
+  every model actually being planned, and warns in both directions: a
+  declared column the SELECT does not produce, and a produced column the
+  declaration does not name.
+
+  The check never refuses. The declaration is often the side that is stale,
+  and the caller is often deliberately changing the model, so a warning in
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
LGTM!

PR #303, opened 2026-08-13
SCRUM-1065-transform test scaffolds dbt unit tests from a model's inputs

### LINKED ISSUE #215: `transform`: scaffold dbt unit tests from a model's inputs
**Depends on:** nothing. **Blocks:** #232.

### Problem

dbt unit tests are the right tool for verifying transformation logic against fixed
inputs, and writing one by hand means restating every input model's column set in a
`given` fixture with correctly typed values. The structural work is mechanical and
it is most of the effort, which is a large part of why unit tests are rare in real
projects.

### Proposal

`transform test --scaffold <model>` derives from the model's compiled SQL: its
input `ref()`s, the columns of each that the model actually reads, and their types.
Emit a `unit_tests:` skeleton with correctly shaped `given` blocks and a stub
`expect`, as a normal `schema_yml` plan.

Two decisions worth stating in the issue:

- **Do not invent expected output.** Fixture *inputs* can be scaffolded; the
  expectation is the thing the author is asserting and inventing it would produce a
  test that passes by construction, which is worse than no test.
- **Only the columns the model reads**, not every column of every input. A fixture
  restating forty columns to test three is why people abandon unit tests.

### Acceptance

- Scaffolding a model with two inputs produces a `given` block per input, with
  only the columns read, correctly typed.
- The scaffold parses under dbt.
- The stub expectation is obviously a stub and fails until filled in.
- Column types come from the cache or the warehouse, not from a guess.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/215

New transform test --scaffold <model> command derives a dbt unit_tests: skeleton from a model's own ref()/source() inputs.
A given block per input, holding only the columns that model actually reads (traced through CTEs and joins via a new sqlglot-based column-lineage pass in transform/test_scaffold.py), typed from the exploration cache, never guessed.
expect: is always a deliberately empty stub with a TODO comment, since inventing the expected output would produce a test that passes by construction.
A select */t.* over a single resolvable source expands against the cache (the common staging-model shape); over more than one joined source, or an unqualified column ambiguous across sources, it's refused rather than guessed at.
Planned like any other schema.yml edit: dbt's own parser gates it before the plan is stored (same layering as transform macro), and nothing touches the project until transform apply.

image
transform test --scaffold customer_orders returned status: "ok" with a given block per input holding only the columns the model reads (customers → {id: 1}, no email; orders → {customer_id: 1, id: 1}, no total) and the expected empty expect: stub with its "fails until you fill in" warning.


marcociav-exmergo (approved):
LGTM!

PR #304, opened 2026-08-13
SCRUM-1049-Use the one DuckDB file in an unconfigured run directory

### LINKED ISSUE #199: Refuse twice before reading a single unambiguous DuckDB file
**Depends on:** nothing

### Problem

With exactly one `*.duckdb` file in the working directory and no `.dex/config.yml`
anywhere up the tree, the first two commands a new user runs both refuse:

```
$ dex explore inventory
no .dex/config.yml found searching from '.' up to the git root: run inside your
dex project, pass --repo-root, or pass --connector/--path for an ad-hoc read

$ dex explore inventory --connector duckdb
no DuckDB path: pass --path or set duckdb.path in .dex/config.yml
```

Verified on 1.5.1.

### Why it matters

The rule these refusals implement is right and should stay: dex must not invent a
connection target, because reading a phantom DuckDB path is how a command silently
answers about the wrong data. But one real file, sitting in the directory the
command was run from, is not a phantom target. It is the single most likely thing
the user meant, and it is the first thirty seconds of the zero-credential on-ramp.

### Proposal

When there is no config and no explicit `--connector`/`--path`, and the run
directory contains **exactly one** `*.duckdb` file, use it, and say so in an
envelope warning that names the file and the config entry that would make the
choice explicit.

Deliberately narrow, so the original rule is preserved:

- Exactly one candidate. Two or more refuses, and lists them.
- The run directory only, no recursive search and no walk up.
- Never applies when a config exists, even a config that resolves to something
  else. An explicit setting always wins.
- Never applies to any connector other than DuckDB, where a target is a
  credentialed remote and guessing is not a thing to do.
- Always warns. It must never be silent, since a warning is what makes a wrong
  guess visible.

### Acceptance

- One `.duckdb` present, no config: the command runs and warns, naming the file.
- Two present: refused, both listed.
- Zero present: the existing refusal, unchanged.
- Config present: unchanged, even if it names a different file.
- `--path` present: unchanged.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/199
With no .dex/config.yml anywhere and nothing explicit passed, a run directory holding exactly one *.duckdb file is now used automatically instead of refused, matching the "single most likely target" case the issue describes.
The choice always warns, naming the file and the --path/duckdb.path that would make it explicit going forward.
Two or more candidates still refuses, but now names every one instead of leaving the caller to guess why.
Zero candidates, a config (even one naming a different file), and an explicit --connector/--path all stay exactly unchanged — the exception is deliberately as narrow as the rule it sits inside.
Along the way, fixed a real keyword-argument evaluation-order bug in DexEngine._adapter() that would have silently made the whole fix a no-op (connector=self.connector was being read before config=self._config_for_open() ran and mutated it).

image
Testing with  real demo199/ directory with an actual DuckDB file : a lone lone.duckdb returns status: "ok" with a real inventory result and a warning naming the file ("no connector configured; using the one DuckDB file in this directory (lone.duckdb)..."); adding a second file (second.duckdb) flips it back to status: "error", now naming both candidates and telling the user to pass --path to disambiguate. Screenshot on hand showing both runs back to back.


marcociav-exmergo (approved):
LGTM!

PR #285, opened 2026-08-11
fix(maintain): populate per-axis findings

floze-the-genius:
## What this changes

`maintain check` now retains each axis result in the command model and serializes `run_at`, `scope`, `finding_count`, and `findings` under `data.axes`.

The top-level `data.findings` remains the globally ranked view. A contract regression verifies that its complete finding multiset matches the concatenated per-axis lists, so an empty axis list can no longer hide findings held elsewhere in the same envelope.

Fixes #279

## Validation

- `uv run pytest -q`: 2,052 passed, 65 skipped
- `uv run pytest tests/maintain -q`: 116 passed
- `uvx pytest evals -q`: 26 passed
- `uvx ruff@0.15.19 check .`
- `uvx ruff@0.15.19 format --check --diff .`
- prose em-dash check
- `git diff --check`

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`)
- [x] Eval scoring-core tests pass (`uvx pytest evals`)
- [x] Safety spine remains green as part of the full suite
- [x] `CHANGELOG.md` updated under `[Unreleased]`
- [x] Prose is em-dash free
- [x] No credentials, secrets, or raw warehouse rows added

Issue #163, opened 2026-07-29
A declared relationship can never be verified, so `maintain grain`'s join-fanout half cannot run


marcociav-exmergo:
**Version:** mechanism read at 1.3.1, re-verified at 1.4.2; live consequence observed on a project that declares all its joins.

## Summary

Two skips chain, and each is locally reasonable:

```python
# explore/relationships.py:403
if rel.kind is not RelationshipKind.INFERRED:
    continue                      # so verified = True is unreachable for a declared join

# maintain/drift.py:426
if not rel.verified or rel.orphan_fraction is None:
    continue                      # so fanout_pairs is built only from verified ones
```

A **declared** join is therefore unverifiable at any budget, by any flag, from any caller. `fanout_pairs` is permanently empty for such a project, and `join_fanout_grew` can never fire.

## `--verify` is not the fix, which is worth stating up front

The obvious remedy (expose `--verify` on the refresh path, or schedule it) is a **non-fix**, and one integrator spent a cycle on it before tracing a level deeper. The skip happens by `kind`, upstream of the flag, before any measurement question arises. Adding `--verify` spends budget for zero additional coverage.

## Evidence

On a project that declares its joins rather than relying on inference, every relationship in the baseline is:

```json
{"kind": "declared", "confidence": 1.0, "verified": false, "orphan_fraction": null}
```

10 of 10. That is the whole set, not a sample. Inference finds none there, because the id columns do not match parent tables by name (the log reads "examined 2 id-shaped column(s) across 12 profiled object(s); no id-shaped column matched a parent table by name").

So `maintain grain` returns `ok` with uniqueness findings only, which is **indistinguishable from a clean join graph**. An operator reading the capability description would reasonably believe a join whose fanout silently grew (duplicate rows after a join, one of the more common real warehouse failures) would be caught. It would not be, and nothing errors.

## The inversion worth naming

The declared channel is weakest exactly where it adds most. A project that **states** its joins, the cooperative case, loses the axis that validates them.

A declared join is perfectly probeable: the overlap SQL does not care how the relationship was learned. The model in `cache.py` only ever contemplates the inferred case, documenting `verified` / `orphan_fraction` as set by the opt-in probe because *"an **inferred** join stays a name-based guess until measured"*. A declared join is not a guess, so it is treated as needing no measurement, and then `grain_plan` refuses it for lacking one.

## Question, before any fix

[...]

PR #305, opened 2026-08-13
Classify an unreadable cache, and rebuild an unreadable drift report

### LINKED ISSUE #249: [core] A corrupt cache or drift report still reports as a bad request, which `_require_baseline` fixed only for the baseline
**Version:** measured at HEAD with the #243 change (PR #247) applied.

`_require_baseline` gave `load_snapshot` the classification it needed: a `ValueError`
out of a load becomes a `PrerequisiteError` naming the command that rebuilds the
document, so a corrupt baseline stops telling the operator they made a bad request.

The other two loads did not get it, and both are still wrong in the same direction.

## `load_cache`, across ten call sites

Every caller invokes the store bare and branches only on `is None`:

- `explore/commands.py` lines 287, 423, 651, 767, 1053, 1207
- `transform/scaffold.py` line 53
- `maintain/commands.py` lines 250, 758, 1068

So a `ValueError` propagates to the CLI catch-all, where `envelope.py` maps
`(ValueError, Reason.REQUEST)` and the operator is told they typed something wrong.
Pydantic's `ValidationError` subclasses `ValueError`, so this is the shipped path for
a corrupt cache, not only a third-party one.

## `load_drift`

Same shape, already recorded in the comment beside `DRIFT_SCHEMA_VERSION` in
`maintain/drift.py`, which names it as a real gap rather than a decision and says why
it was left out of #243: the right remedy differs.

## The remedy is shared; the policy is not

Both want a `_require_cache` mirroring `_require_baseline`. What they must not share
is what happens next, and inheriting symmetry from the baseline would be the wrong
answer for at least one of them.

**A corrupt cache has to refuse.** Treating it as absent would silently re-profile a
warehouse and bill for it, which is the argument `load_cache`'s own docstring already
makes, and the query firewall resolves every table and column reference against the
cache, so an empty read is a security-relevant state rather than an inconvenient one.

**A corrupt drift report can rebuild.** It is derived rather than vouched for:
`maintain check` regenerates it from the baseline on demand, `_record_axes` already
discards a report measured against a different baseline wholesale, and `reconcile`
already has an absent-report path naming the command that produces one.

## Why this is worth its own issue

The `ExploreStore.load_cache` docstring claimed the engine already performed this
classification. It does not, and PR #247 scoped the claim back to `load_snapshot`
rather than overstate the fix, so the contract is now honest about the gap and the gap
is unfixed. A third-party backend author implementing to the contract raises a
`ValueError` today and gets a request error for it.

## Proposed fix

- Add a `CacheUnreadableError(PrerequisiteError)` and a `_require_cache` helper
  mirroring `_require_baseline`, and route the ten `load_cache` call sites through it.
- Settle `load_drift` separately in the same change: treat an unparseable report as
  absent and rebuild, if that survives review, rather than refusing by analogy.
- Extend the storage conformance suite so a backend that cannot produce a parseable
  document is exercised for the cache as well as the snapshot. This is item 1 of
  "Deliberately not in this PR" on #247: the suite currently asserts nothing about an
  unparseable document for either, so a backend can satisfy the whole contract without
  ever reaching this path. It needs an overridable hook letting a backend produce an
  unparseable document in its own storage, the way `make_unreadable_project()` does
  for formats.
- Update the `load_cache` contract paragraph in `storage/base.py` and
  `references/storage.md`, both of which currently document the gap.

catincloudlabs:
## What this changes

`load_cache` raises on a document it cannot parse, and pydantic's `ValidationError`
subclasses `ValueError`, so an unreadable cache fell through to the CLI catch-all
and was classified as `reason: request`. That tells an operator they typed
something wrong, and tells a host to retry with different arguments, when the fix
is a command nobody has run.

This is the same defect `_require_baseline` fixed for the baseline, on the loads
the codebase had already flagged as unfixed. Both notes are quoted rather than
paraphrased, because they are what this PR implements:

> `storage/base.py`, on `load_cache`: *"This load has no such wrapper yet, so what
> a backend raises here reaches the engine's catch-all either way; raise a
> `ValueError` so the load is classifiable when it gets one, not because it is
> classified today."*

> `maintain/drift.py`, beside `DRIFT_SCHEMA_VERSION`: *"What is not yet handled,
> and is a real gap rather than a decision: a drift report that fails to parse
> still raises out of `load_drift` uncaught and is classified as a request error
> [...] the right remedy differs (treat it as absent and rebuild, rather than
> refuse)."*

So this applies an existing pattern in two places, and takes the drift remedy that
note already chose rather than proposing a new one.

### It classifies; it does not require

**The issue asked for a `_require_cache` mirroring `_require_baseline`, and that
shape does not fit.** `_require_baseline` can *require* because all six
`load_snapshot` call sites are identical and every one needs a baseline. Absence
is **legal** at most cache call sites: `explore profile`, `explore relationships`
and `explore map` read a prior cache only to merge pre-run state and a first run
has none; `maintain snapshot` falls back to a metadata capture; `_baseline_warnings`
just skips a warning. A `_require_cache` refusing on `None` would break five
commands.

`readable_cache` therefore returns `None` unchanged and every caller keeps the
absence policy it already had. Only the unparseable case gains a class.

### Thirteen call sites, and one deliberate exclusion

All thirteen engine sites across `explore`, `maintain` and `transform` now go through
it. The issue listed ten; `transform/row_attribution.py` has since added two more, and
`transform test --scaffold` (#303) added a thirteenth while this sat in review, which
the rebase onto main picked up. It types every value in a `given` row from the
exploration cache, so an unreadable cache reached the catch-all there too.

🔴 **`explore/semantic/local.py` is deliberately not routed through it.** Its bare
`except Exception: return None` is documented as intentional (a metric query is
governed by dimension name before any SQL exists, so a repo that never ran
`explore map` can still query metrics), and routing it would make
`explore semantic query` refuse where it currently degrades.

### Where it lives, and why not beside `CacheRequiredError`

`storage/base.py` rather than `explore/commands.py`. Three packages load a cache
and **none of them imports `explore.commands`** today, so the natural home would
have added `transform -> explore` and `maintain -> explore` edges at three modules.
Storage is the neutral home they already share, it is where the contract asking for
the raise lives, and `spend_total` is the precedent for a module-level helper there.
Happy to move it if you would rather it sat next to its sibling.

### The drift half takes the opposite remedy, on purpose

`_stored_drift` treats an unparseable report as **absent**. Both callers already
had that path: `_record_axes` discards a report measured against a different
baseline wholesale, and `reconcile` raises the `NoBaselineError` naming
`maintain check`. No new error class, because neither caller needed one.

`CacheUnreadableError` carries no `schema_version`, unlike `BaselineUnreadableError`.
The asymmetry is the one already reasoned out in `maintain/snapshot.py`: the cache's
version drives a `<` comparison that *degrades*, where the baseline's is a
membership test that *refuses*. A degrading version leaves no refusal to carry.

## Reviewed by mutation

| mutation | result |
|---|---|
| one cache site back to `store.load_cache()` | exactly that command's two arms red; the operator sees the raw `1 validation error for DexCache ...`, which is the defect itself |
| `_record_axes` back to `store.load_drift()` | exactly its two arms red, `reconcile`'s stay green |

⚠️ **The drift corruption fixture carries a positive control, and it earned it.**
An earlier version set a `findings` key that `DriftReport` does not have. The
document still parsed, every arm exercised a perfectly healthy report, and the
refusal test passed for the wrong reason. It now asserts the induced document
actually fails `model_validate_json`, so a fixture that quietly stops corrupting
cannot look like a fix that works.

The quiet arms are the other half: `test_a_readable_cache_is_not_refused` and
`test_a_readable_drift_report_is_not_discarded`. Without them, every assertion
above is equally consistent with *every* cache being refused and *every* report
being discarded.

## Related issues

Closes #249.

⚠️ **#285 is adjacent**: it rewrites `_record_axes` around the `load_drift` line
without touching the line itself. A textual conflict is likely and trivial; happy
to rebase whichever lands second.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`) - **2118 passed, 65 skipped**, from 2094 before
- [x] Eval scoring-core tests pass (`uvx pytest evals`) - 26 passed
- [x] If a safety path is touched, the spine still holds: read-only against data, cost-guarded, PII flagged not surfaced, propose-don't-impose - refusal-path only; nothing queries, and the refusal names that `explore map` **bills**, which the previous behaviour did not
- [x] Prose is em-dash free (checked in CI) - `check_no_em_dashes.py` clean; `ruff check` and `ruff format --check` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,68 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   new persisted state and a new cross-command annotation pass. Filed as a
   follow-up rather than folded in here.
   
+- **`CacheUnreadableError`**, exported from the package root and from
+  `exmergo_dex_core.storage`. The sibling of `CacheRequiredError` that
+  `BaselineUnreadableError` is of `NoBaselineError`: both remediate the same way,
+  so the status is identical, but "nothing has been explored here" and "what was
+  explored will not parse" are different facts about a deployment and only one of
+  them suggests something went wrong. A host can page on the second without
+  matching on prose.
+
+  It carries no `schema_version`, unlike the baseline's, and the asymmetry is the
+  one already reasoned out in `maintain/snapshot.py`: the cache's version drives a
+  `<` comparison that *degrades*, where the baseline's is a membership test that
+  *refuses*. A degrading version leaves no refusal for the attribute to carry.
+
+### Fixed
+
+- **A corrupt exploration cache no longer reports as a bad request** ([#249]).
+  `load_cache` raises on a document it cannot parse, and pydantic's
+  `ValidationError` subclasses `ValueError`, so an unreadable cache fell through
+  to the CLI catch-all and was classified as `reason: request`. That tells an
+  operator they typed something wrong, and tells a host to retry with different
+  arguments, when the fix is a command nobody has run. It is the same defect
+  `_require_baseline` fixed for the baseline in 1.6.3, on the load the storage
+  contract had already flagged: *"this load has no such wrapper yet [...] raise a
+  `ValueError` so the load is classifiable when it gets one."* All thirteen
+  engine call sites, across `explore`, `maintain` and `transform`, now go through
+  `readable_cache`, including the one `transform test --scaffold` added above:
+  it types every value in a `given` row from the cache, so an unreadable one
+  reached the catch-all there too.
+
+  `explore/semantic/local.py` is deliberately **not** routed through it. Its bare
+  `except Exception: return None` is documented as intentional, because a metric
+  query is governed by dimension name before any SQL exists and a repo that never
+  ran `explore map` can still query metrics. Routing it would make
+  `explore semantic query` refuse where it currently degrades.
+
+- **An unreadable drift report is rebuilt instead of refused** ([#249]). Same
+  root cause, deliberately opposite remedy, and the one the note beside
+  `DRIFT_SCHEMA_VERSION` pre-committed to: a baseline is *vouched for* and
+  nothing else reproduces it, while a drift report is *derived* and
+  `maintain check` regenerates it from the baseline on demand. `_stored_drift`
+  treats a document that will not parse as absent, so `_record_axes` rebuilds it
+  exactly as it already did for a report measured against a different baseline,
+  and `reconcile` raises the `NoBaselineError` naming `maintain check` that it
+  already raised for a missing one. No new error class, because neither caller
+  needed one.
+
+### Changed
+
+- **`readable_cache` classifies rather than requires**, which is why there is no
+  `_require_cache` mirroring `_require_baseline`. Every `load_snapshot` caller
+  needs a baseline, so that helper can refuse on `None`. Absence is *legal* at
+  most cache call sites: `explore profile`, `explore relationships` and
+  `explore map` read a prior cache only to merge pre-run state and a first run
+  has none, `maintain snapshot` falls back to a metadata capture, and
+  `_baseline_warnings` merely skips a warning. `None` is returned unchanged and
+  every caller keeps the absence policy it already had.
+
+- The refusal names the cost. `maintain snapshot` is free on every connector, so
+  the baseline's remedy can say "just re-run it"; `explore map` re-profiles the
+  warehouse and **bills**. An operator choosing between investigating a corrupt
+  document and replacing it needs that said before they run it.
+
 ## [1.6.4] - 2026-08-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,52 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   already raised for a missing one. No new error class, because neither caller
   needed one.
 
+- **A declared join is measured, and the measurement does not revise the
+  declaration** ([#163]). A relationship the project declares could not be
+  verified at any budget, by any flag, from any caller: verify selected its
+  probes by `kind`, and the skip happened *upstream* of `--verify`, so asking for
+  verification spent nothing extra and covered nothing extra. `fanout_pairs` was
+  permanently empty for a project that declares its joins, and `maintain grain`
+  returned a result indistinguishable from a clean join graph. The same `kind`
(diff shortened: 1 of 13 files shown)


marcociav-exmergo (commented):
@catincloudlabs looks very good, thanks for this.
I think you can rebase and then we can merge (just some CHANGELOG.md conflicts).

While you're at it, could you also please add some CHANGELOG notes for this PR of yours? https://github.com/exmergo/dex/pull/306

catincloudlabs:
Rebased onto main and the #306 notes are in, as two separate commits so they read apart.

The rebase turned up something worth flagging rather than burying in a force-push. #303 landed a thirteenth `load_cache()` call site while this sat in review: `transform test --scaffold` types every value in a `given` row from the exploration cache. Git auto-merged around it cleanly, so this PR would have gone on claiming "all twelve call sites" while the defect it fixes quietly survived at the new one. It is routed through `readable_cache` now, and the count is corrected in the commit message, the body above, and the changelog.

Verified by mutation, same as the rest: reverting that one line turns nine of the ten scaffold tests red, and `test_scaffold_without_cache_is_a_clean_error` stays green, which is the arm showing `None` still passes through unchanged there.

That is a durability gap rather than a defect in this PR. Nothing caught the thirteenth site, and nothing will catch the fourteenth either. Happy to file a follow-up for a test asserting no raw `store.load_cache()` outside `readable_cache` and the documented `explore/semantic/local.py` exclusion, so the next one turns red instead of reproducing this. Say the word and I will open it rather than widen this diff.

On #306: the entry is under `[Unreleased]` / `Fixed`, keyed to #163 rather than the PR, matching the surrounding entries. It covers the measurement/confidence split, `probe_candidates` as the single selector and why that matters for the estimate, the one extra probe per declared edge under `--verify`, and the two scope limits a reader would otherwise hit by surprise: composites stay excluded, and it is not retroactive, since `grain_plan` reads the baseline snapshot and existing snapshots hold declared joins at `verified: false`.

2158 passed, 65 skipped on the new base, up from 2118. `ruff check`, `ruff format --check` and the em-dash gate are clean.

One flag: #285 still overlaps `_record_axes` if it lands first. Happy to rebase again.

marcociav-exmergo (approved):
Awesome, thanks for this @catincloudlabs!
Merging

Issue #275, opened 2026-08-11
`maintain check` pays BigQuery's 10 MB floor once per probe


marcociav-exmergo:
**Environment:** dex 1.6.3, BigQuery connector.

### Problem

One `maintain check` against a mid-sized BigQuery project settled **86 separate
queries**, each billed at BigQuery's 10,485,760-byte minimum, for **0.89 GB**
total. The volume and grain axes issue their probes per relation.

The data actually interrogated is a small fraction of that. These are row counts
and distinct counts, the cheapest aggregates there are; the bill is almost
entirely floor, not scan.

### Why it matters

`maintain check` is the most frequently run command in the loop, the one a team
is meant to schedule. Its cost is currently a function of relation count rather
than data size, which is exactly backwards: a project with many small tables
pays more than one with a few large ones. At 86 relations it is 0.89 GB a run.
At 300 it is over 3 GB a run to ask questions whose answers are kilobytes.

dex already knows the floor exists, and says so in its own notes:

> BigQuery bills at least 10,485,760 bytes (10 MB) per query; every number here
> already reflects that floor

It also already batches elsewhere: profiling aggregates are documented as
"batched to keep statements bounded". The maintain probes do not get the same
treatment.

### Proposal

Apply the profiling batcher to the maintain probes.

- Volume is `SELECT COUNT(*)` per relation, which batches trivially into one
  statement per dataset with `UNION ALL`.
- Grain is a distinct-count per candidate key and batches the same way.

Expected effect at the observed shape: 86 statements to roughly 8, and 0.89 GB
to roughly 0.1 GB. A ~10x reduction on the loop's most-run command.

### Acceptance

- A `maintain check` over N relations in one dataset issues O(1) statements for
  the volume axis rather than O(N).
- Billed bytes for an unchanged project drop by roughly the ratio of relations
  to datasets.
- Findings are byte-identical to the unbatched implementation.

PR #306, opened 2026-08-13
A declared join is measured, and the measurement does not revise the declaration

### LINKED ISSUE #163: A declared relationship can never be verified, so `maintain grain`'s join-fanout half cannot run

catincloudlabs:
# A declared join is measured, and the measurement does not revise the declaration

Closes #163.

> The design question this issue asks first is answered in
> https://github.com/exmergo/dex/issues/163#issuecomment-5286753646 and reproduced below.
> If maintainers land the other way, this reduces to the docs half: the `cache.py` docstring
> and the `--verify` contract line. Nothing else here would survive that call, so it is worth
> settling before review goes deep.

## The defect

A declared relationship could not be verified at any budget, by any flag, from any caller. The
skip happened by `kind`, upstream of `--verify`, so adding the flag spent budget for zero
additional coverage. `fanout_pairs` was permanently empty for a project that declares its joins,
and `maintain grain` returned a result **indistinguishable from a clean join graph**.

The rule had also outgrown the axis the issue describes. `orphan_findings` landed in #287
(`a340b78`, one commit behind main at the time of writing) gating on `INFERRED` as well, which
made the catastrophic orphan-rate finding that PR shipped unreachable for exactly the cooperative
case it was written for. On the project #163 reports from, that is 10 of 10 relationships,
because inference finds none there.

## The split this turns on

`cache.py` said `verified` and `orphan_fraction` are set only for an inferred join because it
*"stays a name-based guess until measured"*. That is correct about credibility and a non sequitur
about measurability. A declaration is a claim about the data, and the overlap SQL does not care
how the relationship was learned.

- **Measurement** (`verified`, `orphan_fraction`) applies to both kinds.
- **Confidence arithmetic stays inferred-only.** Demoting a declared 1.0 on a 0.2 orphan rate
  would report a data defect as though dex had grown less sure of an edge the project stated.

The disagreement surfaces through `orphan_findings` instead, with its own wording: "the project
and the warehouse disagree" is a different and more actionable claim than a shared name that
turned out not to be a shared key.

## Three things the fix required beyond lifting the filter

**1. Ordering.** Lifting the `kind` filter alone changes nothing on `explore relationships` and
`explore map`: `declared_relationships()` is called *after* the verify handshake, so declared
edges did not exist yet when probing ran. Both commands now verify the merged set. That also
removes any risk of the "declared wins over the same inferred edge" merge rule discarding a
measurement, since at merge time nothing has been measured yet.

**2. One selector.** `probe_candidates` is now the single definition of what verify runs on,
shared by `verify_relationships`, `probe_statements`, and `_verify_estimate`. Those three
previously agreed only by each hard-coding the same filter. Pricing N probes and issuing N+M
under-reports spend *before* it happens, which is the one thing the cost guardrail exists to
prevent, and it is the same failure mode #275 and #276 describe.

**3. A composite-key defect the filter had been hiding.** `_overlap_probe_sql` joins on
`from_columns[0]`/`to_columns[0]` only. That was total coverage while inference was the sole
source, since it emits single-column edges by construction, and stops being so once declared
edges qualify. Probing the first column of a composite key answers about a *different*
relationship and reports its orphan count as the join's, so composites stay unverified until the
probe itself spans a key.

## Verification

`2101 passed, 65 skipped`. `ruff check` and `ruff format --check` clean at the pinned 0.15.19;
the em-dash gate clean.

New coverage, one test per claim:

- `probe_statements` and `verify_relationships` select the same set, asserted against the shared
  selector, so the estimate/run desync cannot regress silently.
- A declared join measured at 0.6 orphans keeps `confidence == 1.0`. This is the design split,
  and the thing a future refactor is most likely to quietly undo.
- The finding text differs by kind, and the declared text does not contain the shared-name
  disclaimer.
- End to end on the grain axis: with the FK declared and the map rebuilt, the cache carries
  `declared, verified=true, orphan_fraction=0.0, confidence=1.0`, and a fanout regression on that
  edge raises `join_orphans_increased` where it previously returned clean.

`test_orphan_findings_decisions` had a parametrized case pinning *"a declared join never produces
a finding"*, which is the defect itself. Migrated rather than loosened.

## Cost

Probing declared joins raises `--verify` spend by one probe per declared edge on a billed
connector. That is covered by the existing handshake precisely because the estimate and the run
now select through the same function.

## Notes

- The declared channel only exists under `--use-project`. Without it `defs` is empty and every
  edge is inferred, which is why this is invisible in any fixture that maps without the flag.
- **The fix is not retroactive.** `grain_plan` reads the baseline snapshot, and existing snapshots
  hold declared joins at `verified: false`; `join_orphans_increased` needs a before-value. Fanout
  drift on a declared join needs a fresh `maintain snapshot` taken after the join was declared.

## Not in this PR

- A probe that spans a composite key. Composites are excluded explicitly, with the reasoning in
  `probe_candidates`, rather than left to look like an oversight.
- Whether a declared join's orphan finding deserves a higher severity than the inferred one. It
  is arguably a stronger signal; left alone because severity is a separate contract.


marcociav-exmergo (approved):
LGTM! Thanks @catincloudlabs

Issue #136, opened 2026-07-23
`maintain semantic` returns its free reference/definition findings inside a `needs_confirmation` envelope


marcociav-exmergo:
**Environment:** dex 1.3.1, BigQuery connector.

## Summary

The reference/definition half of `maintain semantic` is free and completes immediately, but because the dimension-cardinality half is paid, the whole response is returned as `needs_confirmation`. To read the free findings you have to accept a response shaped like a pending paid action and filter the findings out of it. Getting the free answer should not look like a pending charge.

## Detail

Running `maintain semantic` with no `--confirm` returns `status: needs_confirmation` with a `cost.estimate` for the cardinality scan. The `data.findings` array (the free reference/definition results) is already fully populated inside that same envelope, and the notes confirm "the definition and reference checks are free and already complete". So the findings are final, but the envelope status implies nothing has run yet.

## Expected

Either the free findings come back as a normal `ok` result with the paid axis offered separately, or the envelope makes explicit that the findings present are final and only the cardinality scan is pending.

## Proposed fix

A flag to run only the free reference/definition axis (e.g. `--no-scan` / `--refs-only`) that returns `ok`, or an envelope shape that separates "completed free findings" from "confirm to add the paid scan".

marcociav-exmergo:
**Version:** dex 1.4.0, BigQuery connector. Same pathology, wider blast radius than this issue's title suggests.

This issue scopes the problem to `maintain semantic`. It applies identically to **`maintain check`**, and matters more there, for two reasons.

First, `check` is documented as "the everyday entry point" — the first command a maintenance session runs. Second, three of its four axes (schema, volume, and the reference/definition half of semantic) are free and complete on the first call; only grain and dimension-cardinality scan. So the *majority* of what `check` produces arrives inside an envelope shaped like a pending paid action.

Concretely: this session's opening `check` returned **373 findings** — the entire triage, everything needed to understand the state of the project — inside a `needs_confirmation` whose `estimated_bytes` referred to work I had not yet agreed to and, on a first pass, did not want. Reading the free answer meant treating a refusal-shaped response as a result and parsing findings out of it.

[...]

PR #308, opened 2026-08-14
SCRUM-1047-Skip the confirmation handshake where nothing can bill

### LINKED ISSUE #197: A cost handshake on a free connector gates nothing
**Depends on:** nothing. Subsumes **#136**.

### Problem

`transform build --target dev` against DuckDB returns:

```json
{"status": "needs_confirmation",
 "data": {"command": "transform build", "target": "dev",
          "hint": "review the cost, then re-run with --confirm ..."},
 "cost": {"paradigm": "free_local", "estimate": 0.0, "ceiling": null}}
```

Verified on 1.5.1. The estimate is zero, the paradigm is `free_local`, and nothing
is billable, so the handshake asks the caller to confirm spending nothing. Every
DuckDB build costs a wasted round trip.

**#136** reports the same shape from the other side: `maintain semantic` returns
its free reference and definition findings inside a `needs_confirmation` envelope.

### Why it matters

The argument for keeping the hosted semantic layer outside the handshake is that a
handshake which gates nothing trains people to click through handshakes that do.
That argument applies unchanged here, and it applies to an interactive human as
much as to an automated caller. A confirmation prompt is a scarce attention budget;
spending it where there is nothing to confirm devalues it everywhere else.

### Proposal

State it as one rule and apply it everywhere: **a confirmation handshake is
emitted only where spend is possible.**

- On `free_local`, a command that would otherwise raise `needs_confirmation`
  proceeds, and the envelope carries a warning saying the handshake was skipped
  because the connector cannot bill.
- On any metered paradigm, nothing changes.
- Where a command returns a mix of free and billed work, the free half returns as
  `ok` with its findings and the billed half carries the ask, which is exactly
  what #136 requests.

Add a test asserting that no `free_local` path can produce
`status: needs_confirmation`, so the rule cannot regress quietly.

### Open question for the maintainer

Whether to keep an opt-in strict mode for callers that want a uniform handshake
regardless of paradigm. The recommendation is no: a uniform handshake that is
sometimes meaningless is the thing this issue is removing.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/197
transform build --target dev on DuckDB returned needs_confirmation for an estimate of 0.0 on a free_local connector, asking the caller to confirm spending nothing.
guards/cost_guard.py's three gate points (preflight(), CostGate.preflight_command(), CostGate.preflight_phase()) now skip the confirmation and ceiling-required checks when the paradigm is FREE_LOCAL; an unconfirmed run proceeds instead of stopping.
Over-ceiling enforcement stays unconditional across every paradigm, exactly as before: an estimate that contradicts an explicitly configured ceiling is still refused even on a free connector, since that's the caller's own contradiction, not a spend question.
The envelope now carries a warning naming why nothing was asked ("no confirm handshake: this connector cannot bill..."); passing --confirm anyway stays silent about it, since nothing was actually skipped worth remarking on.
Verified maintain semantic/maintain check were already correct on DuckDB (no adapter attaches a cost gate to a free connector, so their two-phase confirmation never triggers there) no change needed for those.
Closes #197. Closes not #136: that issue's own scenario (a mixed free/billed result on a billed connector still reading needs_confirmation even with real findings already attached) lives in results.to_envelope()'s unconditional pending-confirmation handling, a separate, larger change not made here.
Before: 
image
After:
image


marcociav-exmergo (approved):
LGTM!

PR #318, opened 2026-08-15
A statement the warehouse refuses is classified, names its object, and reports what it spent; profiling a non-PII string column no longer kills the statement on Redshift

### LINKED ISSUE #310: The skill wrappers require `uv` but never say so, and fail with a raw traceback when it is missing
All three skills shell out through `subprocess.call(["uv", "run", ...])` with no
guard (`skills/{explore,transform,maintain}/scripts/run.py:169,181`, byte-identical
in all three). On a machine without `uv` on `PATH`, here is what a user gets:

```
$ PATH=/usr/bin:/bin python3 skills/explore/scripts/run.py explore inventory
Traceback (most recent call last):
  File ".../skills/explore/scripts/run.py", line 185, in <module>
    sys.exit(main())
  File ".../skills/explore/scripts/run.py", line 181, in main
    return subprocess.call(cmd, env=env)
  ...
FileNotFoundError: [Errno 2] No such file or directory: 'uv'
```

Reproduced on a clean clone of `main`.

The `SKILL.md` instructions are worse still, because they tell the agent to invoke
`uv run "${CLAUDE_SKILL_DIR}/scripts/run.py"` (`skills/explore/SKILL.md:19`,
`skills/transform/SKILL.md:17,62`, `skills/maintain/SKILL.md:40`), so on a machine
without `uv` the failure is a bare `command not found: uv` from the shell before
our code runs at all.

## Why this is a blocker and not a nice-to-have

- Anthropic's [Software Directory Policy](https://support.claude.com/en/articles/13145358-anthropic-software-directory-policy)
  requires that developers "keep software operational" and "handle errors gracefully
  with helpful feedback". A stack trace is neither.
- The person hitting this is a Claude Code user, not necessarily a Python developer.
  Claude Code does not install `uv`. There is a meaningful chance someone installing
  the plugin does not have it.
- Nothing anywhere in the plugin, the manifest, the README's Claude Code install
  section, or the skill frontmatter mentions that `uv` is required. Our own README's
  Claude Code path is two `/plugin` commands and then "the skills appear and
  auto-trigger". A user who follows exactly that gets a traceback.
- **The failure is silent to us.** dex collects no telemetry, by design, so every
  user who hits this churns invisibly and we never learn about it. That makes it
  exactly the class of first-run bug worth eliminating pre-emptively rather than
  waiting for a report that will never arrive.

## Fix

In all three `scripts/run.py`, guard the exec and emit the same sanitized JSON
envelope shape the engine emits, so the agent can read the failure and tell the user
what to do rather than pasting a traceback:

```python
if shutil.which("uv") is None:
    print(json.dumps({
        "ok": False,
        "error": {
            "code": "uv_not_found",
            "message": (
                "dex runs its engine through uv, which was not found on PATH. "
                "Install it with: curl -LsSf https://astral.sh/uv/install.sh | sh "
                "(or `brew install uv`, or `pipx install uv`), then re-run."
            ),
        },
    }))
    return 1
```

Note that the wrappers are stdlib-only and currently import neither `shutil` nor
`json`, so both need adding. That is fine: the wrapper's constraint is that it runs
before the engine is installed, not that it stays minimal.

Then add one line to each `SKILL.md` under "How to drive it" stating the `uv`
prerequisite, and one line to the README's Claude Code install block.

Roughly two hours including tests. Worth a safety-spine style test that asserts the
guard exists in all three wrappers, since it is the kind of thing that gets
refactored away.

## Out of scope, deliberately

Whether the wrapper should fall back to `python -m exmergo_dex_core` when the engine
is already importable and `uv` is absent. That is a larger change with its own
questions (which interpreter, which extras, how the pinned version is honoured), and
it should not hold up the guard. A clear refusal that tells the user what to install
is a complete fix for the reported failure; the fallback is a separate enhancement.

spascarella:
## What this changes

`explore map --confirm` and `explore profile --confirm` fail against Redshift on
any table with a non-PII string column, and have since 1.6.2. Three defects sit
on that path and each one hid the next. This fixes all three.

### 1. The profiling aggregate no longer depends on lazy CASE evaluation

`type_contradiction_expressions` guarded every `CAST` behind a length-bounded
shape predicate inside a `CASE`, on the premise that a dialect without
`TRY_CAST` would then never evaluate the cast for a row the `WHEN` excluded.
Postgres honors that premise. Redshift does not: it evaluates the branch for
rows it never selects, so one ordinary varchar status or category column was
enough to fail the whole statement with `Invalid digit, Value 'p', Pos 0,
Type: Long`.

Every `CAST` the probe builds is now total. Its argument is a `CASE` that
yields a digit-only string on every row, so no evaluation order can reach a
cast with something it cannot parse:

    CAST(CASE WHEN "status" ~ '^-?[0-9]{1,10}$' THEN "status" ELSE '0' END AS BIGINT)

The sentinel is rejected by every predicate built on the cast (the epoch ranges
start in 2000, the slash-component test asks for `> 12`), so a row that reaches
the cast only because the cast is unconditional is never counted as evidence.
The "shaped versus not shaped" distinction the fractions' denominators need
moved to a separate uncast expression, which carries a string and so cannot
raise. One change in the shared expression builder covers all six adapters and
removes the standing assumption about lazy `CASE` evaluation everywhere.

Measurements are unchanged, denominators included.

### 2. A statement the warehouse refuses is now classified and named

A server-side SQL error escaped the adapters untranslated. Not being a
`DexError`, it fell through every entry in `_reason_overrides()` and arrived as
`reason: internal` ("not a deliberate dex refusal") with `data: {}`. Same class
of misclassification as #252.

Every adapter now raises a typed `WarehouseQueryError`, exported from the
package root and mapped to `reason: execution_failure`, carrying the server's
own message and error code trimmed to one line and capped. Redshift reads `M`
and `C` out of the driver payload rather than stringifying the whole dict.
Postgres gates on `sqlstate` and Databricks on `ServerOperationError`, so a
connection that died mid-statement is not reported as SQL the server rejected.
Profiling attaches the object name at the one place that knows it, since a map
run has a dozen statements in flight.

### 3. A failed billed command reports what it spent

`_record_elapsed` runs in a `finally`, so the wasted compute reached
`.dex/spend.jsonl` correctly, but the error envelope carried `data: {}`. On a
metered connector a caller reading only the envelope saw a failure that
appeared to cost nothing. Any error envelope now reports the settled spend,
as the budget-exhaustion path already did. Nothing billed means no spend block,
so a refusal that never reached the warehouse still says nothing rather than
claiming a zero.

### Also

The live suites assert through a new `assert_ok(rc, envelope)` helper that
prints `errors`, `reason`, and `warnings`. `assert rc == 0, envelope` is why
this took sixteen CI runs to diagnose: pytest elides the middle of a long
assertion message, and the middle of a profiling envelope is where `errors`
sits, so the Redshift message never once reached a CI log. 35 call sites across
all five connectors.

## Evidence

The offline suite passed before this change too, so the tests are the argument
here as much as the code is.

**The fix reproduces the bug offline.** A new test lifts each `CAST` argument
out of the generated statement and casts it over every row unconditionally,
which is exactly the work Redshift does eagerly. Under the old shape it raises
on DuckDB; under the new one it cannot raise anywhere. Alongside it,
`assert_every_cast_is_total` asserts the structural invariant on all six
adapters' generated SQL. Reverting the expression change fails 7 tests.

**The fix is behavior-preserving.** Old and new expressions were run side by
side on DuckDB across 12 fixtures (pure text, mixed epoch and text, MDY, DMY,
ambiguous slash dates, boundary epochs, all-null, empty table). All 14 aliases
agree on every fixture. End to end, `dex demo` plus `explore map` produces a
cache byte-identical to the pre-change run apart from `updated_at`, including
both of the demo warehouse's designed type-contradiction findings.

**The envelope defects are pinned.** A Redshift adapter test drives a real
`Invalid digit` payload through profiling and asserts the failure is typed,
carries the message and SQLSTATE, names the object, and still bills its
seconds to the ledger. A CLI-level test drives a refusal mid-profile on a
metered connector and asserts `reason`, the server's words, the object, and
`data.spend`, with its counterpart asserting that a refusal which never reached
the warehouse reports no spend at all.

## Not done here

**This needs the live Redshift suite to confirm it.** I have no credentials for
that workgroup, so the fix is unverified against a real Redshift. The offline
evidence above is as close as it gets. The residual risk is narrow: it is that
Redshift also pushes a cast down into `CASE` branches, which would defeat the
sentinel shape too. That is not its documented behavior and this is the
standard workaround for it, but only a green Redshift job proves it.

Separately, and left alone deliberately: `epoch_seconds_fraction` is measured
over the epoch-shaped rows, so a column that is half digits and half `pending`
reports "100% of values fall in the plausible unix-epoch range". That reads
wrong, but it is pre-existing and identical before and after this change, and
moving a detection threshold is not this fix's job. Worth its own issue.

## Related issues

Closes #310.

## Checklist

- [x] Tests pass (`uv run pytest` in `packages/dex-core`): 2197 passed, 65
      skipped. The 4 deselected `test_packaging.py` cases install wheels from
      PyPI and GitHub and cannot run on this machine (TLS trust store); the
      other 13 packaging tests pass.
- [x] If a safety path is touched, the spine still holds: read-only against
      data, cost-guarded, PII flagged not surfaced, propose-don't-impose. The
      generated SQL is still SELECT-only in every dialect, the aggregates still
      emit fractions rather than values, and spend now appears on more
      envelopes rather than fewer.
- [x] No credentials, secrets, or raw warehouse rows in code, tests, or
      fixtures. One judgment call worth review: `WarehouseQueryError` passes the
      server's message through to the envelope, and a server message can quote
      the fragment of a value that offended it (`Invalid digit, Value 'p'`).
      That matches what an agent SQL failure already surfaces, and suppressing
      it is what made this class of failure undiagnosable. The class docstring
      states the tradeoff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,29 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **Profiling a non-PII string column no longer kills the statement on
+  Redshift** ([#310]). The declared-type-vs-content probe (#204) guarded every
+  `CAST` behind a length-bounded shape predicate inside a `CASE`, on the
+  premise that a dialect without `TRY_CAST` would then never evaluate the cast
+  for a row the `WHEN` excluded. Postgres honors that premise; Redshift does
+  not, and evaluates the branch for rows it never selects, so one ordinary
+  varchar status or category column was enough to fail the whole profiling
+  statement with `Invalid digit, Value 'p', Pos 0, Type: Long`. The probe
+  shipped in 1.6.2, so on Redshift that took down `explore profile`,
+  `explore relationships`, and `explore map`, plus the commands that profile
+  the object they name before running (`explore query`, `explore cluster`).
+
+  Every `CAST` the probe builds is now total: the argument is a `CASE` that
+  yields a digit-only string on every row of the column, so no evaluation
+  order can reach a cast with something it cannot parse, and the sentinel it
+  falls back to is rejected by every predicate built on the cast. The
+  measurements are unchanged, denominators included; the "shaped versus not
+  shaped" distinction the fractions need now comes from a separate uncast
+  expression instead of from the cast result being NULL. The fix is
+  dialect-agnostic and lands in the shared expression builder, so the standing
+  assumption about lazy `CASE` evaluation is gone from all six adapters, and an
+  offline invariant test asserts the shape on every one of them.
+
 - **`maintain check` carries each axis's findings in the command envelope**
   ([#279]). The top-level `data.findings` ranking could report drift while the
   adjacent per-axis result did not carry those findings, making an axis look
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,21 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   assumption about lazy `CASE` evaluation is gone from all six adapters, and an
   offline invariant test asserts the shape on every one of them.
 
+- **A statement the warehouse refuses is classified, names its object, and
+  reports what it spent** ([#310]). A server-side SQL error escaped the
+  adapters untranslated. Not being a `DexError`, it fell through every reason
+  override and arrived as `reason: internal` ("not a deliberate dex refusal")
+  with `data: {}`: nothing to branch on, no object named, and no spend, on a
+  connector that had already billed the seconds the statement ran before it
+  died. Every adapter now raises a typed `WarehouseQueryError` (exported from
+  the package root, `reason: execution_failure`) carrying the server's own
+  message and error code, trimmed to one line and capped; profiling names the
+  object the refused statement was reading; and an error envelope from a
+  metered connector reports the spend the ledger recorded, as the
+  budget-exhaustion path already did. The live suites now assert with a helper
+  that prints `errors` rather than an envelope repr pytest truncates, which is
+  what kept the Redshift message out of sixteen CI logs.
+
 - **`maintain check` carries each axis's findings in the command envelope**
   ([#279]). The top-level `data.findings` ranking could report drift while the
   adjacent per-axis result did not carry those findings, making an axis look
--- a/packages/dex-core/src/exmergo_dex_core/__init__.py
+++ b/packages/dex-core/src/exmergo_dex_core/__init__.py
@@ -109,6 +109,7 @@ _EXPORTS = {
     "StoreContext": "storage",
     "StoreFactory": "storage",
     "StoreRequiredError": "errors",
+    "WarehouseQueryError": "errors",
     "generate_demo_warehouse": "demo",
     "render_er_mermaid": "explore.diagram",
     "to_envelope": "results",
@@ -143,6 +144,7 @@ if TYPE_CHECKING:  # what a type checker and an IDE see; never run
         RepoRootRequiredError,
         RequestError,
         StoreRequiredError,
+        WarehouseQueryError,
     )
     from .explore.cluster import ClusterDependencyError, ClusterError
     from .explore.commands import CacheRequiredError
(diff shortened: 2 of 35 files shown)


marcociav-exmergo (approved):
LGTM! nice breakdown

PR #325, opened 2026-08-17
SCRUM-1067-A PII refusal names the override entry that would clear it

### LINKED ISSUE #217: A PII refusal should emit the override entry that would clear the column
**Depends on:** nothing

### Problem

The query firewall refuses a projection that would carry values from a flagged
column:

```
query refused: the projection would carry values from PII-flagged column(s):
dim_accounts.account_name (name). Use a measuring aggregate over them
(COUNT, APPROX_COUNT_DISTINCT, AVG(LENGTH(...))), or drop them from the output.
PII values never cross the envelope.
```

The refusal is correct and the policy is right. False positives are the design
working, and `account_name` on a table of companies, or a team name in a sports
dataset, are exactly the columns a name detector should flag and a human should
clear.

But the message offers only two routes, both of which mean not getting the answer.
The third route, a reviewed override, exists and is not mentioned. In practice the
caller does the fourth thing: opens a raw SQL client, where neither the firewall
nor the PII policy applies at all.

### Why this is not a request to loosen detection

It is the opposite. The override mechanism exists precisely so that clearing a
column is a recorded human decision rather than a tuning knob, and the reason it is
underused is that finding it requires knowing it exists and then working out the
exact entry. Making the correct route the easy one is what keeps people on it.

### Proposal

When the firewall refuses on PII, include in the envelope the exact
`pii_overrides` entry that would clear the named column, in both the exact and the
pattern form, ready to paste into `.dex/config.yml`.

Everything that makes the override a decision stays: it is still a committed
configuration change, still reviewed in a diff, still recorded in the cache as the
category the detector had matched, and dex still never applies it itself.

Worth considering, and worth deciding in the issue rather than assuming: whether
the suggestion should be suppressed for the highest-severity categories, where the
right answer is more often "no" than "review it".

### Acceptance

- A PII refusal includes the exact override entry for the named column.
- The pattern form is included when the column shape suggests it recurs across
  tables.
- The suggestion is text in the envelope, not an applied change.
- No value from the flagged column appears in the suggestion, asserted in the
  safety suite.
- The existing aggregate-based guidance is kept, since it is often the better
  answer.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/217
The query firewall's PII refusal offered only two routes (aggregate or drop the column); the third route, a reviewed pii_overrides entry, existed and was never mentioned, so in practice callers opened a raw SQL client instead, where neither the firewall nor the PII policy applies.
The refusal now names, for each blocking column, the exact pii_overrides entry that would clear it, in flow-mapping YAML ready to paste into .dex/config.yml: the fully-qualified exact form always, plus a scope-glob pattern form when the same column name is flagged at the same category on another dataset.
The suggested scope is never wider than the schema that evidence came from (the dataset's own identifier with the table segment wildcarded); a caller who knows it reaches further widens the glob themselves.
Widened the internal _Taint type in guards/query_firewall.py from (label, confidence) to carry the fully-qualified identifier, column name, and category too, since the short table name in the display label alone isn't enough to build a safe, unambiguous override (two datasets can share a short name).
The suggestion is text in the message only, built from column identity (dataset, column, category) alone, the same structural guarantee PIIFlag itself makes, so no value can appear in it by construction. Nothing is applied automatically.
Answered the issue's open question (suppress for high-severity categories?) as no, documented in the CHANGELOG: nothing in the engine ranks PIICategory by severity today, the block threshold is deliberately uniform, and the suggestion is inert text a human must still copy/edit/commit regardless of category.

Before the fix: 
image
After : 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,43 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **A PII refusal names the exact override entry that would clear the
+  column** ([#217]). The firewall's refusal was correct, and so is a name
+  detector flagging `account_name` on a table of companies or a team name in
+  a sports dataset; the false positive is the design working, and a human
+  reviewing it is the intended next step. But the message offered only two
+  routes, a measuring aggregate or dropping the column, both of which mean
+  not getting the answer. The third route, a reviewed `pii_overrides` entry,
+  existed and was never mentioned, and the cost of that omission was not the
+  override itself but the interval before anyone learned it existed: in
+  practice the caller opened a raw SQL client instead, where neither the
+  firewall nor the PII policy applies at all.
+
+  The refusal now says, for each blocking column, the exact entry that would
+  clear it, in flow-mapping YAML ready to paste under `pii_overrides:` in
+  `.dex/config.yml`: the fully qualified form always, and a scope-glob
+  pattern form alongside it when the same column name is flagged at the same
+  category on another dataset, which is the actual evidence a wider scope is
+  worth suggesting rather than a guess at how far a naming convention
+  reaches. The suggested scope is never wider than the schema that evidence
+  came from; a caller who knows it reaches further widens the glob
+  themselves. Both forms are text in the message, exactly like the existing
+  aggregate guidance beside them: nothing is applied, nothing changes about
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
LGTM! Thanks for this @emnarahmouni-exmergo, I actually had this issue a few days ago

PR #326, opened 2026-08-18
SCRUM-1068-explore profile flags a boolean-shaped column with >2 values

### LINKED ISSUE #218: `explore profile`: flag a boolean-shaped flag with more than two values
**Depends on:** #203

### Problem

A column whose name says boolean (`is_*`, `has_*`, `*_flag`, `*_yn`, `*_ind`) and
whose content holds more than two non-null values is a mixed-encoding defect, and
it is not reported. On the public ADE-bench `helixops_saas.duckdb`,
`raw_workspaces.primary_ws_yn` holds **five** distinct values and
`raw_workspaces.ws_stat_cd` holds six that encode three states.

### Why it matters

Every `where flag = 'Y'` written against such a column is quietly wrong for some
share of rows, and the share is invisible without asking for the domain.

### Proposal

When a column's name matches a boolean-ish pattern and its non-null distinct count
exceeds two, emit a data-quality observation. Once #203 lands, include the domain
so the observation names the encodings rather than only their number.

### Acceptance

- A `*_yn` column with five values produces the observation naming them.
- A clean two-value flag produces none.
- A three-valued flag where the third value is a genuine third state still
  produces the observation, since the tool cannot tell those apart and the caller
  can.
- A boolean-typed column produces none.

emnarahmouni-exmergo:
Closes #218.
A column named like a two-valued flag (is_*, has_*, *_flag, *_yn, *_ind) can hold more than two distinct non-null values, a mixed-encoding defect that went unreported: raw_workspaces.primary_ws_yn on the public ADE-bench helixops_saas.duckdb holds five values under a name that promises two.
explore profile now emits a data-quality observation when a boolean-ish name and a non-null distinct count above two coincide.
Where the value domain ([#203](https://github.com/exmergo/dex/issues/203)) is already known for that column, the observation names each encoding with its count; where it's not (the column failed one of that feature's own eligibility gates), it still fires on the count alone, since not being able to name the values is not evidence there are only two of them.
A genuinely BOOLEAN-typed column is excluded outright, whatever its name it cannot hold more than two values by construction.
Deliberately does not try to distinguish a data-quality defect from a genuine third state; the tool can't make that call, so it names what it found and lets the caller decide (a required acceptance criterion, not an oversight).

image

1. primary_ws_yn (5 values) → flagged, naming them: "primary_ws_yn looks boolean-shaped by name but holds 5 distinct non-null values: 1=20, N=20, No=20, Y=20, Yes=20; a two-valued read of it will misclassify some rows"  every encoding named with its exact count.
2. has_discount (clean 2-value flag) → silent: no entry for it anywhere in data_quality.
3. onboarding_stage_flag (3 values, a genuine third state) → still flagged: "...holds 3 distinct non-null values: new=34, active=33, churned=33..." fires exactly as the acceptance criteria requires, since the tool can't (and isn't asked to) tell a defect apart from a real third state.
4. is_native_bool (real BOOLEAN type) → silent: no entry, even though nothing else about its name would exclude it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **`explore profile` flags a boolean-shaped column with more than two
+  values** ([#218]). A column named like a two-valued flag (`is_*`, `has_*`,
+  `*_flag`, `*_yn`, `*_ind`) whose content holds more than two distinct
+  non-null values is a mixed-encoding defect, and it went unreported: on the
+  public ADE-bench `helixops_saas.duckdb`, `raw_workspaces.primary_ws_yn`
+  holds five distinct values under a name that promises two. Every
+  `where flag = 'Y'` written against a column like that is quietly wrong for
+  some share of rows, and the share was invisible without asking for the
+  domain directly.
+
+  A data-quality observation now fires when a boolean-ish name and a
+  non-null distinct count above two coincide, naming the encodings and their
+  counts wherever the value domain ([#203]) is already known for that
+  column, and falling back to the count alone where it is not (a column that
+  failed one of that feature's own eligibility gates is not evidence there
+  are only two values, only that the tool cannot name them). A genuinely
+  `BOOLEAN`-typed column is excluded outright, whatever its name: it cannot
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
Very nice, thank you @emnarahmouni-exmergo

PR #327, opened 2026-08-18
SCRUM-1069-explore profile: detect cumulative and snapshot measures

### LINKED ISSUE #219: `explore profile`: detect cumulative and snapshot measures
**Depends on:** #206

### Problem

A numeric column that holds a **running total** or a **point-in-time snapshot**
(a season-to-date score, a subscription's current MRR, an account balance, an
inventory level) profiles identically to one that holds a per-row increment.
Nothing warns that summing it across rows multiplies it.

### Why it matters

This is one of the most damaging analytics errors available, because the result is
plausible. Summing a cumulative measure over a period produces a number that is
large, monotonic, correctly typed, passes every uniqueness and null test, and is
wrong by a factor that grows with the number of periods.

It is also detectable from data, which is what makes it worth doing here rather
than leaving it to reviewer discipline.

### Proposal

For a numeric column in a table that has both an entity key and a temporal or
sequence column, measure the fraction of consecutive rows within entity, ordered by
time, where the value **decreases**. A fraction at or near zero over enough
observations is strong evidence of a cumulative or snapshot measure.

Report it as a data-quality observation that names the consequence: this column
does not sum across rows; take the latest value per entity, or difference
consecutive values to recover the increment.

A natural follow-on, worth its own issue once this exists: when authoring a
semantic-layer measure over such a column, propose the aggregation type that
matches the detected shape rather than defaulting to a sum.

### Cost

Higher than the rest of the profile: it needs a window function over the table
rather than a flat aggregate. It should be attempted only where the shape
prerequisites hold (an entity key, a temporal or sequence column, and a numeric
measure), it belongs on the gated side of the free-versus-gated split, and it
should be individually skippable.

### Acceptance

- A cumulative column produces the observation.
- A per-row increment column with natural ups and downs does not.
- A column that is monotonic for an unrelated reason (an auto-increment id) is
  excluded by the key check rather than reported as a measure.
- Tables lacking an entity key or a temporal column are skipped, and the skip is
  stated rather than silently producing nothing.
- Only fractions leave the engine.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/219
`explore profile --check-cumulative` (#219): flags a numeric column that
looks like a running total, an account balance, or a point-in-time
snapshot instead of a per-row increment. This shape profiles identically
to an ordinary measure (same type, null fraction, uniqueness), so summing
it downstream is a common and severely damaging misreading that nothing
else in `explore profile` catches.

## How it works

- Signal: within an entity (a repeating, id-shaped column that is not
  itself the table's key) ordered by a temporal column, a cumulative or
  snapshot measure almost never decreases from one observation to the
  next. A genuine per-row increment has natural ups and downs.
- The probe is a single aggregate-only, window-function query authored
  once in DuckDB SQL and transpiled per connector dialect via sqlglot
  (mirrors `relationships.py`'s overlap-probe pattern). No row's measure
  value ever leaves the engine, only comparison counts.
- Priced and gated like `--verify`: the base profile always completes
  and saves first; the check runs as a second, individually skippable
  phase, dry-run priced only once profiling knows what to test, and can
  come back as `pending_confirmation` on a billed connector rather than
  discarding already-paid-for work.
- A table missing an entity key or a temporal column is skipped with an
  explicit note, never silently reported as clean.

## Bugs found and fixed during manual testing

Two real bugs surfaced while exercising this against a live DuckDB file
(not caught by the initial test pass, since it always exercised a fresh
scan):

- A dataset served from the freshness cache is a deep copy, not the
  object the cache write carries forward, so a note appended only to it
  showed in that run's output but never reached disk. Fixed by resolving
  the cache's own object by identifier before appending.
- Running the check again on an already-flagged dataset duplicated the
  same sentence. Fixed by making the append idempotent (skip if the exact
  text is already present).

Both are covered by dedicated regression tests, plus manual stress
testing across multiple objects in one call, `--refresh` combined with
the flag, a first-ever call already using the flag, and a table with no
eligible shape.

image
balance correctly flagged as cumulative, purely from its shape in the data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,30 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **`explore profile --check-cumulative` detects a running total or
+  point-in-time snapshot measure** ([#219]). A numeric column holding a
+  running total, an account balance, or a subscription's current MRR
+  profiles identically to one holding a per-row increment: same type, same
+  null fraction, same uniqueness. Summing it across rows is a common and
+  severely damaging misreading, and it silently inflates every aggregate
+  built on top of it.
+
+  The signal is structural: within an entity (a repeating, id-shaped column
+  that is not itself the table's own key) ordered by a temporal column, a
+  cumulative or snapshot measure almost never decreases from one observation
+  to the next, while a genuine per-row increment has natural ups and downs.
+  Measuring that needs a window-function scan over the table, so it sits on
+  the gated side of the free-versus-gated split: opt-in via
+  `--check-cumulative`, priced and confirmed the same way `--verify` prices
+  relationship overlap probes. The base profile always completes and is
+  returned first; the check runs as a second, individually skippable phase
+  against what the base scan already proved, and never blocks it. A table
+  missing an entity key or a temporal column is skipped with a note, not
+  silently reported as clean, and only fractions and observation counts ever
+  leave the engine, never a measure's value.
+
 ## [1.6.6] - 2026-08-15
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


marcociav-exmergo (approved):
Looks good, thanks @emnarahmouni-exmergo

PR #330, opened 2026-08-19
SCRUM-1070-explore: propose joins from measured value overlap

### LINKED ISSUE #220: Propose joins from measured value overlap when names do not match
**Depends on:** #208

### Problem

Every proposed relationship starts from a name. When the naming carries no signal
at all (`acct_id_fk` to `ws_id`, or a source that names every key `id`), inference
has nothing to work with and the map is silently incomplete. The caller cannot
tell an absent edge from an absent relationship.

### Why it matters

"Point it at a warehouse you do not have a project for" is the case where naming is
least likely to help, and value overlap is the only evidence that survives bad
naming. It is also strictly stronger evidence than a name, since it is the thing a
name is a proxy for.

### Proposal

An opt-in, bounded, cost-gated sweep over key-shaped columns that no name-based
rule matched, proposing edges from measured containment alone.

Bounding is the whole design, since this is combinatorial by nature:

- candidates restricted to columns already established as keys or near-keys, and to
  pairs whose types are compatible
- a hard cap on probes per run, with the cap and the elision reported
- priced as a batch before anything runs, through the normal handshake, and never
  attempted implicitly by a command the caller did not opt into
- an edge proposed this way is labeled as overlap-derived and never presented as
  equivalent to a declared foreign key

Run it after #208, so the cheap name-based path has already been exhausted and
the expensive sweep has fewer candidates left.

### Acceptance

- Off by default, and the flag that enables it is named in the map's notes as
  something that was not run.
- The probe count is capped, reported, and priced before execution.
- An overlap-derived edge is distinguishable from a name-derived one in the cache
  and in the envelope.
- Only aggregate counts leave the engine.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/220

`explore relationships`/`explore map --infer-by-overlap` (#220, depends on #208): proposes a join from measured value overlap when no column name gives inference anything to work with (`acct_id_fk` to `ws_id`, or a source that names every key `id`). Value overlap is strictly stronger
evidence than a name, since a name is only ever a proxy for it.

## How it works 

- Off by default, and named in `explore map`'s notes even when off, so the option is discoverable without `--help`.
- Candidates: key-shaped columns (a proven single-column key, or a near-key whose distinct count clears the near-unique ratio, PII excluded) not already covered by a declared or name-inferred edge, paired across datasets, restricted to type-compatible pairs.
- Probed with the exact same aggregate-only overlap-probe SQL `--verify` already uses (authored once, transpiled per dialect); only a candidate clearing a strict containment ceiling is proposed, the rest are dropped outright.
- Bounded: the candidate pool is capped, and both the cap and how many candidates it elided are reported before anything runs, priced as a batch through the same handshake `--verify` uses.
- Runs after `--verify` on the same merged edge set, deferring entirely if verify's own checkpoint is still pending, so at most one checkpoint is ever in flight.
- A proposed edge carries a new `RelationshipKind.OVERLAP_INFERRED`, distinguishable from a declared or name-derived edge in both the cache and the envelope.

## Bug found and fixed during manual testing

An overlap-derived edge vanished on the very next plain run, since nothing else ever rediscovers it the way a declared edge (re-read from
the dbt project every run) or a name-inferred one (re-derived from cheap metadata every run) does. Fixed by carrying a prior overlap-derived edge forward unconditionally as long as both endpoints are still known objects. That fix also closes a redundancy: a later sweep now skips a pair it already confirmed instead of re-probing (and re-paying for) it.

Verified manually end to end against a real DuckDB pair with zero naming signal: discovery, persistence across a plain re-run, and idempotency on a repeated sweep all confirmed. 

image
explore relationships --infer-by-overlap discovers a real 1:1 join between billing_profiles.profile_ref and accounts.acct_num purely from measured value containment, the two column names share no naming convention at all.

image
The discovered edge survives a plain re-run with no flag. This was a real bug I found and fixed during manual testing: nothing else ever rediscovers an overlap-derived edge, so without this fix it would silently vanish the moment you ran the command again without --infer-by-overlap.


marcociav-exmergo (approved):
LGTM

Issue #137, opened 2026-07-24
[core] Injectable storage seam with in-memory default


marcociav-exmergo:
## Problem
All `.dex/` state persistence is concentrated in two concrete, filesystem-bound classes with no abstraction between them and the ~24 call sites that use them:
- `DexStore` (`packages/dex-core/src/exmergo_dex_core/cache.py:200`): `cache.json`, `snapshot.json`, `drift.json`, and the append-only ledgers `queries.jsonl` / `spend.jsonl`.
- `PlanStore` (`packages/dex-core/src/exmergo_dex_core/transform/plans.py:87`): the `.dex/plans/*.json` documents.

Every call site constructs its own store from `repo_root` (`DexStore(repo_root)` at ~19 sites, `PlanStore(repo_root)` at ~5). There is no way to substitute a non-filesystem backend, which blocks both in-process library use (#138) and a hosted deployment (#140).

## Goal
Introduce a storage `Protocol` (mirroring the existing `Adapter` protocol at `adapters/base.py:273`) and inject an implementation at the entry point instead of hardcoding filesystem access. Ship two implementations:
- `FilesystemStore`: today's behavior, byte for byte. Remains the CLI default.
- `MemoryStore`: retains all state in instance attributes for the process lifetime, writes nothing to disk.

Note `MemoryStore` must **retain**, not discard: `cmd_query` and `cmd_cluster` read the cache back (`explore/commands.py:579, 961`), so a store that drops writes would break them on the second call.

## Scope
- Define the `Protocol` covering the full `DexStore` load/save/append surface plus `PlanStore`'s plan CRUD.
- Refactor the ~24 call sites to receive an injected store rather than a `repo_root` string.
- Fix the `Path` leak: methods like `save_cache` return concrete `Path` objects that callers stringify into envelopes (for example `transform/commands.py:571`). Replace with opaque locators or drop the path from the return contract, so non-filesystem backends are representable.
- The `repo_root` resolution point (`command_args.py:52`) becomes where the CLI selects `FilesystemStore`.

## CLI impact
None. This is an internal refactor; the CLI keeps constructing a filesystem store from `repo_root` and user-visible behavior is unchanged.

## Non-goals
- No ephemeral/`--no-cache` flag on the CLI. An in-memory CLI run only makes sense for a single self-contained command and silently breaks any follow-up command that needs the cache. The in-memory store's home is the library API (#138).
- Config persistence (`load_config`/`save_config`) is out of scope here; it is handled in #138.

[...]

PR #345, opened 2026-08-26
SCRUM-991-Add opt-in SQLite cache backend

### LINKED ISSUE #139: [core] Opt-in SQLite cache backend

emnarahmouni-exmergo:
Closes: https://github.com/exmergo/dex/issues/139
- Adds `SqliteStore`, an opt-in cache backend that writes a single
  `.dex/dex.db` file instead of the loose JSON `FilesystemStore`
  writes under `.dex/`. Select it with `cache.backend: sqlite` in
  `.dex/config.yml` or `--cache-backend sqlite` for one run;
  `filesystem` stays the default.
- Implements the full storage Protocol from #137: the cache,
  snapshot, drift report, transform plans, and both ledgers.
  `spend_since` (what the cumulative session budget reads on every
  billed command) answers from an indexed query on the ledger's
  `at`/`connector` columns instead of scanning every line ever
  appended.
- The cross-process spend lock is a `BEGIN IMMEDIATE` transaction on
  the database file itself, so the cumulative ceiling still binds
  across two CLI processes sharing a repo, the same guarantee
  `FilesystemStore`'s advisory file lock provides.
- No new dependency (stdlib `sqlite3`); no changes needed to
  `config.py`'s `cache.backend` field, `cli.py`'s `--cache-backend`
  flag, or `engine.py`, since all three were already fully wired to
  the resolver before this change. The only wiring was adding
  `SqliteStore.from_context` to `resolver.py`'s `SHIPPED` registry.
- `.dex/dex.db` is a binary file, not a JSON document a reviewer can
  diff, so it belongs in `.gitignore` (documented in
  `references/storage.md`).
Before:
Screenshot 2026-08-26 141345
After: 
Screenshot 2026-08-26 141402
Screenshot 2026-08-26 141904

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,30 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **An opt-in SQLite cache backend** ([#139]). `FilesystemStore` writes loose
+  JSON under `.dex/`, which is right for the CLI (persistence is git, a reviewer
+  reads the state in a pull request) and wrong for a host that wants durable
+  local state without a directory of files to gitignore and clean up. Select
+  `cache.backend: sqlite` (or `--cache-backend sqlite` for one run) and
+  everything lands in a single `.dex/dex.db` file instead; `filesystem` stays
+  the default.
+
+  It implements the full storage `Protocol` from [#137]: the cache, the
+  snapshot, the drift report, the transform plans, and both ledgers, backed by
+  real tables rather than a directory of loose files, and it passes the same
+  conformance suite the two shipped backends already run against. The spend
+  ledger's `spend_since`, the query the cumulative session budget reads on every
+  billed command, answers from an indexed lookup on `at` and `connector` rather
+  than reading and parsing every line ever appended, which is the gap the
+  filesystem backend's JSONL scan does not close as a ledger grows. The
+  cross-process spend lock is a `BEGIN IMMEDIATE` transaction on the database
+  file itself, so the cumulative ceiling still binds across two CLI processes
+  sharing a repo, the same guarantee the filesystem backend's advisory file lock
+  provides.
+
+  `.dex/dex.db` is a binary file, not a JSON document a reviewer can diff, so it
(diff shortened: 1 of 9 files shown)


marcociav-exmergo (approved):
This LGTM!

Thank you @emnarahmouni-exmergo

PR #369, opened 2026-08-27
SCRUM-1017-Add governed drop_orphan_relations macro

### LINKED ISSUE #151: `maintain reconcile`: governed execution for `orphan_relation` cleanup via `dbt run-operation` (deferred follow-up to #113 / PR #146)
**Version:** context from dex 1.4.0; detection half shipped in 1.4.1.

Follow-up to #113 / PR #146.

## Why now

PR #146 closed #113 with, verbatim:

> Scope: detection + advisory only, two-state classification (no snapshot history) per the issue's own suggested split; **execution (macro + dbt run-operation) deferred to a follow-up issue**

That follow-up does not exist yet. This is it — and it comes with a working reference implementation, because a maintenance session on 1.4.0 had to resolve five real orphans and the cleanup had to happen regardless.

## Field evidence

Five orphans across three environments in one project: one intermediate view left behind by promoting a model into marts, and two marts tables left by an earlier `dim` → `fct` rename, the view instance replicated across all three environments.

As argued in the earlier comment on #113, layer promotion under a marts-scoped semantic-layer credential is *routine correct architecture* that mints an orphan every time, not untidiness. The frequency argument holds up.

Resolution had to leave dex. It did **not** have to leave the toolchain: the drop ran as a project-local dbt macro through `dbt run-operation`, which is exactly the shape PR #146 deferred.

## Reference implementation (shape, ~40 lines of Jinja)

- Takes an **explicit list** of `schema.identifier` or `database.schema.identifier` relations. Nothing inferred, so blast radius is visible in the invocation itself.
- **Dry-run by default**; executing requires `dry_run: false`.
- Resolves via `adapter.get_relation`, drops via `adapter.drop_relation` — no hand-written DDL string, so the agent-authored-SQL guard is never involved.
- **Refuses any identifier still present as a live model** in `graph.nodes`. Verified firing by aiming it at a real model; a typo cannot delete something real.
- Missing relations are **skipped, not errors** → idempotent, safe to re-run per target.
- Logs `would drop` / `dropped` / `skip` per relation, plus a summary naming the target.

## Why dex should own this rather than each project reinventing it

The guardrail is the argument. A hand-written macro checks `graph.nodes` because that is what a macro can reach. **dex already holds the manifest and the snapshot history**, so it can make the strictly better classification described in the earlier #113 discussion:

1. **Provable dbt orphan** — manifest-backed in a prior snapshot, absent from the current manifest, still present in the warehouse. The only class that should ever be proposed for execution.
2. **Foreign object** — in a dbt-built namespace, never manifest-backed in any snapshot. Advisory only.
3. **Unknown** — no snapshot history deep enough to classify. Advisory.

PR #146 shipped two-state classification precisely because it had no snapshot history to consult. The execution half is where that distinction starts to carry weight.

## Invariants this preserves

- dex still never executes DDL at prod: the macro runs under the project's existing deploy identity, not dex's.
- The human gate gets *stronger*, not weaker — code review of a committed diff rather than a `--confirm` prompt.
- Capturing the recreate DDL into the plan record makes a dropped view fully reconstructible, so the plan is its own undo script.
- On dev targets, where `transform build --target dev` already executes, the same macro can run directly under the existing preflight/confirm handshake. Asymmetric by environment, exactly as the safety spine already is.

## Caveat on field validation

The earlier #113 comment offered to leave the live orphans in place as a test case for whenever the fix landed. They were dropped during this session's requested cleanup, before #146 was known to have shipped. PR #146 has its own unit tests so nothing was blocked, but that environment can no longer field-validate `orphan_relation` on 1.4.1 without manufacturing a fresh orphan.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/151
- Follow-up to #113 / PR #146, which shipped `orphan_relation` detection and an advisory-only `DROP TABLE ...;` proposal, with
  execution explicitly deferred: "execution (macro + dbt run-operation) deferred to a follow-up issue."
- Adds `transform/assets/macros/drop_orphan_relations.sql`, a shipped macro (reachable via `transform macro drop_orphan_relations`, through the same plan/apply flow every other shipped macro uses).
 - It takes an explicit list of relations (nothing inferred), is dry-run by default, resolves/drops through `adapter.get_relation`/
  `adapter.drop_relation` rather than a hand-written DDL string, and refuses to run at all if any named relation is still a live model, seed, or snapshot in the current manifest, so a typo in the list cannot delete something real. A relation that no longer exists in the warehouse is skipped rather than raised on.
- `maintain reconcile`'s `orphan_relation` advisory now proposes that macro plus the exact `dbt run-operation drop_orphan_relations --args '{relations: [...], dry_run: false}'` invocation, with the finding's identifier already filled in, instead of a bare `DROP TABLE` string. Reconciling more than one orphan in a run adds one more warning with a single batched invocation covering all of them.
- dex still never executes DDL: authoring the macro and printing the invocation is as far as dex's role goes, the same boundary
  `transform macro` already draws for every other shipped macro. The human runs it under the project's own deploy identity.

## Explicitly out of scope (deferred, matching how PR #146 itself scoped down)

- **Three-state classification** (provable dbt orphan / foreign object / unknown): needs dex to retain more than the single
  most-recent `maintain snapshot`, which today's storage protocol does not keep. A separate, larger piece of work.
- **Dev-target direct execution**: letting dex run the macro itself on dev targets through the existing `transform build` preflight/
  cost-gate. Bounded, but a new dbt-invocation surface, and the original issue floats it as a "could," not a requirement.

image
maintain reconcile's orphan_relation proposal before/after: the action field now names the governed drop_orphan_relations macro and gives the exact dbt run-operation invocation with the relation already filled into --args, instead of the old bare DROP TABLE ...; string for a human to type by hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,36 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **A governed macro for dropping orphan relations** ([#151]), the execution
+  half PR #146 deferred when it shipped `orphan_relation` detection
+  ([#113]). `maintain reconcile` used to hand back a bare `DROP TABLE ...;`
+  string for a human to type by hand; it now proposes scaffolding
+  `transform macro drop_orphan_relations` and running it through `dbt
+  run-operation`, with the exact invocation (the finding's identifier
+  already filled into `--args`) spelled out in the proposal. Reconciling
+  more than one orphan in a run adds one more warning naming a single
+  batched invocation for all of them.
+
+  The macro itself takes an explicit list of relations (nothing inferred),
+  is dry-run by default, resolves and drops through `adapter.get_relation`/
+  `adapter.drop_relation` rather than a hand-written DDL string, and refuses
+  to run at all if any named relation is still a live model, seed, or
+  snapshot in the current manifest, so a typo in the list cannot delete
+  something real. A relation that no longer exists in the warehouse is
+  skipped rather than raised on, so the same list is safe to re-run per
+  target. dex still never executes it: authoring the macro and printing the
+  invocation is as far as dex's role goes, the same boundary `transform
+  macro` already draws for every other shipped macro, and the human runs it
+  under the project's own deploy identity.
+
+  Two pieces of the original issue are explicitly deferred rather than
+  bundled in here, matching the scope PR #146 itself drew: a three-state
+  classification (provable dbt orphan / foreign object / unknown) needs dex
+  to retain more than the single most-recent `maintain snapshot`, which
+  today's storage protocol does not keep; and running the macro directly on
+  a dev target through the existing build preflight is a separate, bounded
+  addition to `transform build`'s dbt-invocation surface.
+
 - **`transform rename` and `transform remove` generate the whole propagation
   plan** ([#221]). `transform references` could tell you where a name was used
   and then left you to retype the change file by file. These two make the change:
--- a/packages/dex-core/src/exmergo_dex_core/maintain/reconcile.py
+++ b/packages/dex-core/src/exmergo_dex_core/maintain/reconcile.py
@@ -106,12 +106,15 @@ def build(
 
     schema_patches: dict[str, list[DriftFinding]] = {}
     definition_churn = False
+    orphan_identifiers: list[str] = []
     for finding in findings:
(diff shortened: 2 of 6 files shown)


marcociav-exmergo (approved):
LGTM! This is awesome

PR #373, opened 2026-08-27
SCRUM-1021-maintain schema: detect model added/removed/changed

### LINKED ISSUE #164: `maintain snapshot` fingerprints `transform_layer` counts and never diffs them
**Version:** established by exhaustive grep across 1.2.1, 1.3.1 and 1.4.1; re-verified at 1.4.2. This is a claim about code that does not exist, so there is no live repro.

## Summary

`maintain snapshot` fingerprints the transformation layer in detail (models, `ref()` / `source()` edges, per-file hashes), and the envelope reports `file_count`, `model_count` and `source_count`:

```python
# maintain/commands.py:131-132
file_count=len(transform_layer.files),
model_count=len(transform_layer.models),
```

They are serialized into the baseline (`maintain/results.py:25-40`). **Nothing compares them.**

## The tightest form of the proof

Baseline `snap.transform_layer` is dereferenced in five places at HEAD, and none is a count comparison:

| site | what it does |
|---|---|
| `maintain/drift.py:161` | `_backed_tables`, feeding `orphan_relation` |
| `maintain/drift.py:167` | the same orphan path |
| `maintain/drift.py:275,282` | `dangling_source`, iterating declared sources |
| `maintain/drift.py:877` | `annotate_impacts`, which decorates findings rather than detecting |

There is no sixth reader to audit. The symbols `transform_drift`, `model_added`, `model_removed` and `model_changed` do not appear anywhere in the distribution.

So none of these raise anything today:

- a model **added** to the project
- a model **removed**, arguably the interesting one, since a deleted model is a real change to what the warehouse is supposed to be built from
- a model's **`ref()` edges rewired**, which silently changes every future blast radius
- a model's **content hash** changing, though `files` records exactly that

## Scoping correction, made against an earlier version of this claim

`orphan_relation` (#113 / PR #146) is the first baseline-to-current transform-layer comparison ever to ship, so "a project-shape change raises no drift on any axis" is now too strong. The corrected claim: **a project-shape change raises drift only where it leaves an unbacked warehouse relation behind.** Model added and model content changed remain invisible.

## The asymmetry is the argument

The semantic layer gets all of this: `definition_added` / `definition_removed` / `definition_changed`.

And it is the **same function**. `semantic_free_drift` takes `current_transform` as its first parameter and `snap` (carrying the baseline transform layer) as its last, so current and baseline transform layers are both in scope, together, in the one function whose job is diffing. It touches the transform halves exactly once, as a membership test for `dangling_reference`. Everything a `transform_drift` axis would need is already sitting in that signature.

## Why it matters to a consumer

A downstream operator documented an expectation, and held it for months, that a project regeneration changing the model count would raise drift with an acknowledgement step. That reads like a reasonable inference from a tool that fingerprints the model count and reports it back. It was never true.

**The fingerprint's existence implies a comparison that is not there.** The envelope reports three counts that look like a tracked baseline and are not one.

## Question rather than a defect claim

Is a `transform_drift` axis over the stored file hashes deliberately out of scope, on the reasoning that a project edit is the user's own change and not "drift"? That is a defensible position and this can be closed on it.

If so, the cheap fix is to stop reporting the three counts as if they were tracked, or to say in the docs that the transform fingerprint drives blast-radius attribution and dangling-reference detection rather than a drift axis of its own.

If not, a `transform_drift` axis symmetric with `semantic_free_drift` is the obvious shape, and the fingerprint already carries everything it would need.

## Related

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/164

- Follow-up correction to `maintain snapshot`/`maintain check`: the transform layer's fingerprint (model names, per-file content hashes) was captured on every snapshot and reported back as `file_count`/`model_count`/`source_count`, but nothing ever diffed it. A model added, removed, rewired to a different `ref()`/`source()`, or edited in place all raised zero drift; only a warehouse table left orphaned by the change was ever caught (#113 / PR #146).
- Adds `transform_drift` in `drift.py`, symmetric with the semantic axis's existing `definition_added`/`_removed`/`_changed`: new `model_added`, `model_removed`, `model_changed` findings. Folded into the existing `schema` axis rather than a new CLI verb, since `schema_drift` already loads the project's transform layer for `orphan_relation`.
- `model_changed` diffs content hashes through a new `TransformLayer.model_paths` field (model name to the one file that builds it), so it also catches a rewired `ref()`/`source()` call with no second comparison needed: rewiring one changes the file's text, and so its hash. A baseline pinned before this field existed has an empty `model_paths`, and a model missing from it is skipped for the content comparison rather than reported changed.
- `reconcile.py` treats `model_*` findings the same way it already treats a semantic `definition_*` finding: the project's own edit, not a warehouse-side problem to fix, so no proposal is generated, just a warning nudging toward `maintain snapshot` if the change is intended.
- Docs: `AGENTS.md`, `references/command-contract.md`, and `results.py`'s `LayerFingerprint` docstring updated to describe the real behavior instead of the gap.

## Note on scope

This deliberately reverses an existing, tested design decision, not an oversight: `TransformLayer`'s own docstring stated "nothing compares [the file fingerprint] across snapshots to raise a finding," and `test_transform_layer_scope.py` had a regression test locking in "no detector diffs the file set or the model list" as intended behavior, with its own stated reasoning. That reasoning turned out to be protecting against a different, narrower risk (a node-family-widening causing false drift against an old baseline) rather than a considered stance that project edits should never be drift, and it left an asymmetry with the semantic axis that made the fingerprint's existence misleading (the issue's core complaint: three reported counts that looked tracked and were not). The regression test is updated, not removed, and now pins the opposite with the same rigor.

## Test plan

- [x] `tests/maintain/test_schema.py`: `model_added` on a new staging model, `model_removed` in isolation (never built) vs. alongside `orphan_relation` (was built), `model_changed` on a content edit, `model_changed` catching a rewired `ref()` with no separate check, silence on an unchanged project
- [x] `tests/maintain/test_reconcile.py`: `model_*` findings produce no proposal, just the re-snapshot warning
- [x] `tests/maintain/test_transform_layer_scope.py`: the existing regression test rewritten to assert the new real findings (`model_added` x2, `model_removed` x1) instead of zero
- [x] `uv run pytest tests/maintain -q`: 173 passed
- [x] Full suite: 2730 passed, 84 skipped, 5 pre-existing failures (Windows path-separator bugs in `demo`, unrelated to this change)
image


marcociav-exmergo (approved):
LGTM

PR #379, opened 2026-08-28
SCRUM-1123-explore inventory: cap --rank's payload by default

### LINKED ISSUE #289: `explore inventory --rank`: a 464.7 KB payload containing every object in the catalog
## Problem

`explore inventory --rank` returns the entire catalog. Against a 2,356-model dbt
project, a single invocation returns **3,493 objects in a 464.7 KB payload**. Every
such payload captured during a 103-trial agent run was byte-identical in size, which
is what a fixed full-catalog dump looks like rather than anything responsive to
`--rank`.

The ranking itself works: all 3,493 objects carry a populated `rank_score`, sorted
descending. The problem is purely that the command hands back everything it ranked.

Every invocation in that run was piped through a truncator by the caller before it
was read: `| head -100`, `| head -300`, `| head -c 4000`, and in one case redirected
to a file and post-processed with a Python one-liner.

Measured on 1.6.3, DuckDB connector.

## Why it matters

`inventory --rank` is the command whose entire purpose is "tell me what matters
here, in order". A half-megabyte answer is not a ranked shortlist, it is the catalog
with a sort applied, and at that size the ranking is the one part the caller never
reads.

This is also the first command a new caller runs against an unfamiliar warehouse. It
is the worst possible place to spend the context budget, and a caller who gets
464.7 KB back on their first command reasonably concludes the tool is not for them.

## Proposal

- Default `--rank` to a bounded head of the ranking, with `--limit` to widen it and
  an explicit `--all` to get today's behaviour. A default in the range of 20 to 50
  objects, each with its score and one line of justification, is the shape the
  command's name promises.
- State the ranking basis in the envelope. A rank the caller cannot inspect is a
  rank they cannot trust, and at 3,493 rows nobody is inferring it from the output.

#202 is the same principle from the opposite side: `map` returns too little where
`inventory --rank` returns too much. #288 is the same failure on `explore profile`,
and a shared `detail` level would settle both.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/289
- `explore inventory --rank` returned the entire ranked catalog unbounded. Measured on a 2,356-model project: 3,493 objects in a 464.7 KB payload, every real invocation in an agent benchmark run piped through `head` by the caller before being read.
- `--rank` now defaults to the top 30 objects by score. `--limit N` widens it, `--all` lifts the cap entirely (today's behavior). Both are no-ops without `--rank`, since the unranked list carries no order to cut a shortlist from.
- Elided objects are counted in `elided_object_count` and named in a note, the same cap-and-count convention `explore map` already uses.
- A ranked call now states its basis in `notes`: size, naming convention, and shape feed the score, and connectivity does not, because inventory runs no relationship pass (`explore map`'s rank does include it).

## Explicitly out of scope

Per-object one-line rank justifications were part of the original proposal but are deferred: they need `rank()` to expose per-signal contributions rather than only the final score, which is new design work the capping fix doesn't need.

image
explore inventory --rank against a 40-table warehouse: the default cap kicks in at 30 objects (elided_object_count: 10, with a note naming the cap and how to widen it), --limit 5 narrows further, and --all returns the complete, unbounded ranking (40/0). Before this fix, all three calls would have returned all 40 with no elided_object_count field at all.


marcociav-exmergo (approved):
LGTM

PR #381, opened 2026-08-28
SCRUM-1122-explore profile: lead with the verdict, summarize wide columns

### LINKED ISSUE #288: `explore profile`: the grain verdict is the last 2% of the payload, and the first thing truncated
## Problem

A single `explore profile` of one staging table returns 38,651 bytes. The table has
107 columns. The payload is a `columns` array of 107 entries followed, at the very
end, by the fields that answer the question the caller asked:

| Field | Byte offset | Position in payload |
|---|---|---|
| `columns` | 150 | 0.4% |
| `candidate_keys` | 37,575 | 97.2% |
| `grain` | 37,735 | 97.6% |
| `data_quality` | 37,953 | 98.2% |

Measured on 1.6.3, DuckDB connector, profiling a 2,037-row staging table with 107
columns.

The consequence is mechanical. Claude Code truncates a large tool result to a short
preview and spills the rest to a file, so the agent sees 107 columns worth of
`distinct_count` and never sees `grain`. An agent that defends itself with `| head`
gets the same outcome.

Both behaviours turn up in practice. Across a 103-trial agent run against an
unfamiliar warehouse, `explore profile` output was routinely piped into `head`,
`tail`, `python3` or `jq` before the agent had seen any of it, and the harness
truncated it anyway: the spilled payloads ran to a median of 38.6 KB and a maximum
of 124.7 KB.

The pipes into `tail` are the tell. The agent had worked out that the useful part
was at the end, and was discarding the column list on purpose to reach it. It should
not have to choose.

## Why it matters

Any caller with a context window pays this, and the cost is worst exactly where dex
is most useful: a wide table in an unfamiliar warehouse is both the case with the
most columns and the case where the grain matters most.

It also silently inverts the tool's value proposition. `explore` exists so the
caller does not have to dump a schema into context. At 107 columns, `profile` **is**
the schema dump, with the sense-making stapled to the end where it gets cut.

## Proposal

Lead the envelope with the verdict and let the caller ask for the rest.

- Emit `grain`, `candidate_keys`, `data_quality`, `row_count` and a short findings
  block **before** `columns` in the serialised payload. Key order in JSON is not
  semantically meaningful, but it is operationally decisive here and it costs
  nothing.
- Add a `--columns` control with a summarising default: return the columns that
  carry a finding (non-zero null fraction, PII, a data-quality note, membership in a
  candidate key) and a count of the rest. `--columns all` restores today's
  behaviour.
- Consider a `detail` level shared across `explore` subcommands rather than a
  per-command flag, since #289 and #290 want the same control.

The full column detail should stay available and unchanged under an explicit flag.
The problem is the default, not the data.

#202 is the same principle from the opposite side: `map` returns too little where
`profile` returns too much. Same rule, "sense-making, not enumeration", violated in
both directions.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/288

- `explore profile` on a wide table returned the verdict a caller actually asked for (`grain`, `candidate_keys`, `data_quality`, `row_count`) at 97-98% of the payload's byte offset, behind a `columns` array of every profiled column. Measured on a 107-column staging table: 38,651 bytes, with the useful fields at the very end. An agent harness that truncates a large tool result to a preview never saw them; the same invocations were observed piped through `head`/`tail`/`jq` before the agent had read any of it.
- Verdict fields now lead each dataset's serialized dict, `columns` trails. Key order carries no semantic meaning to a parser but is operationally decisive for truncation, the same fix already used elsewhere in this module (`SemanticValuesResult`).
- Each dataset's `columns` is summarized by default to the ones carrying a finding: PII, a non-zero null fraction, membership in a candidate/composite key (or being proven unique), a reported value domain, or a mention in one of the dataset's own `data_quality` sentences. The note check is a word-boundary match, not a substring, so a column named `am` can't match a note about `amount`.
- The rest are counted in `elided_column_count` rather than silently dropped. `--columns all` restores every column.
- `ProfileResult.datasets` itself stays the full, unreduced set; only `.data()`'s serialization is affected, so a library caller reading `datasets` directly is unaffected by the CLI's `--columns` default.

## Design note

This is a new predicate (`Dataset.columns_with_findings`), deliberately not a widening of the existing `notable_columns` that `explore map`/`diagram` already share. That method's notion of "notable" is grain/key/join/PII role only, with no concept of null fraction or data-quality mentions; widening it in place would have changed what those two commands consider notable too.

image
explore profile's default output on a table with a clean finding/no-finding split: the payload key order now leads with candidate_keys/grain/data_quality before columns, the columns list is summarized to just id (key) and email (PII), and elided_column_count: 2 accounts for the two columns left out. --columns all restores the full four-column list with elided_column_count: 0.


marcociav-exmergo (approved):
LGTM!

PR #382, opened 2026-08-28
SCRUM-1130-Turn dex --help into onboarding, not a flag dump

### LINKED ISSUE #296: `dex --help` is doing the onboarding, and it is not built for it
## Problem

`dex --help` is doing the onboarding, and it is not built for that job.

In a 103-trial agent run, `which dex` followed by `dex --help` was among the first
things a session did, in a container where the tool was already on PATH and a
project `CLAUDE.md` listed four of its commands with one-line descriptions. Several
sessions ran it more than once.

The common form is the interesting part:

```
which dex; dex --help 2>&1 | head -30
```

That is a caller who expects the help output to be long enough to need truncating,
truncating it before reading a word of it.

## Why it matters

This is lower value than most of what is open, and worth filing anyway, because it
measures the onboarding surface in a way that is hard to get otherwise. If a large
share of first contacts go through `--help`, then `--help` **is** the onboarding
documentation, whatever the README says. The `| head -30` says callers do not expect
it to be short.

## Proposal

- Ensure top-level `dex --help` answers, inside roughly 30 lines: what the three
  verbs do, how to point it at data, and what single command to run first against an
  unknown warehouse.
- Consider what a bare `dex` with no arguments returns. For a caller probing the
  surface it is the most likely first keystroke, and it is a good place for the
  one-screen orientation.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/296

- `dex --help` used to be a bare one-line description plus an undocumented
  flag/subcommand dump. It answered none of "what do Explore/Transform/
  Maintain do", "how do I point this at data", or "what do I run first
  against a warehouse I've never seen" -- exactly what a first-time caller
  needs.
- The description now names the three verbs, every group carries its own
  one-line help text in the subcommand listing (previously only `demo`
  did), and an epilog covers pointing dex at data and the first command to
  run (`dex demo` with no warehouse, `dex explore map` with one).
- A bare `dex` used to fail with argparse's "the following arguments are
  required: group" and exit 2. It now prints the same orientation and
  exits 0, since a stranger's first keystroke should not spend itself on
  an error about an argument they don't know exists yet.

Screenshot 2026-08-28 145623
Screenshot 2026-08-28 145652
Before: `dex --help` was a bare description plus an undocumented flag/subcommand dump, and a bare `dex` just errored with "the following arguments are required: group"

After (screenshots): the description names the three verbs, every group has its own help line, and an epilog covers pointing dex at data plus what to run first. A bare `dex` now shows the same orientation and exits 0 instead of erroring.


marcociav-exmergo (approved):
Nice, this one's nice and helpful

PR #384, opened 2026-08-29
Issue-307: `explore map` and `explore profile` failed on the seeded schema

spascarella:
## Redshift: `explore map` and `explore profile` failed on the seeded schema

Three bugs on one code path, each hidden behind the one before it. All three are
fixed and verified live against the seeded Serverless workgroup.

### 1. `DATEDIFF` has no `TIMESTAMPTZ` overload

Redshift resolves `DATEDIFF` to `pg_catalog.date_diff`, declared over
`DATE`/`TIME`/`TIMETZ`/`TIMESTAMP`, while `DATE_TRUNC` over a `TIMESTAMPTZ`
column returns `TIMESTAMPTZ`. The periods the temporal-continuity probe diffs
were exactly the shape it rejects, so any table with a `TIMESTAMPTZ` column
failed the whole profiling statement. Both operands are now cast to `TIMESTAMP`.

### 2. `SUBSTR` is refused by name

`SUBSTR() function is not supported (Hint: use SUBSTRING instead)`, raised at
execution over a real table. The shared slash-date extraction now goes through a
per-connector callable defaulting to `SUBSTR`, with Redshift supplying
`SUBSTRING`. The swap is not universal in reverse: BigQuery has `SUBSTR` and no
`SUBSTRING`.

### 3. A progress line could fail a run that had already billed

`ProgressReporter`'s `stream` was a `TextIO = sys.stderr` default argument,
evaluated once at import. Under pytest that buffer is closed by the time a run
crosses the five-second progress gate, so `explore map` profiled the schema,
spent its compute seconds, and returned `errors: ['I/O operation on closed
file.']` instead of the map. The stream now resolves per line and the write is
best-effort. The same binding also broke `contextlib.redirect_stderr` for any
embedder.

### Tests

- Offline: both SQL spellings pinned against the Redshift fake, with a
  `TIMESTAMPTZ` column added to the fixture. The old one profiled a bare
  `TIMESTAMP`, which is why an offline test asserting this statement passed while
  the live one could not run. Two more tests cover the progress stream.
- Live: `tests/integration -m redshift`, 8 passed, 3 skipped.
- Full offline suite: 2870 passed.

Also documents both dialect refusals in `references/redshift.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,50 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Redshift: `explore map` and `explore profile` died on any table with a
+  `TIMESTAMPTZ` column, and on any slash-date string column.** Two spellings in
+  the generated profiling statement are refused by the Redshift server, and
+  because both ride the single batched aggregate pass, either one failed the
+  whole profile rather than degrading one statistic.
+
+  `DATEDIFF` resolves to `pg_catalog.date_diff`, declared over
+  `DATE`/`TIME`/`TIMETZ`/`TIMESTAMP` with no `TIMESTAMPTZ` overload, while
+  `DATE_TRUNC` over a `TIMESTAMPTZ` column returns `TIMESTAMPTZ`. So the periods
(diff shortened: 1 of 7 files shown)


marcociav-exmergo (approved):
Let's let it rip

PR #385, opened 2026-08-31
SCRUM-1201-Close the DexEngine/CLI parity gap and enforce it with a test

### LINKED ISSUE #344: CLI and `DexEngine` can drift: no test asserts the two surfaces are at parity
**Depends on:** nothing.

### Problem

`DexEngine` and the CLI are two independent surfaces over the same
implementation, and nothing keeps them in step.

The shape is easy to misread. The implementation functions take the engine as
their *first argument* (`transform.commands.plan(engine, intent, ...)`), so a
`cmd_*` handler and the matching `DexEngine` method are **siblings** over a
shared layer, not a stack. On `main`, 27 of 27 `cmd_*` handlers call the
implementation function directly; none of them goes through a `DexEngine`
method.

The consequence is that a capability can exist on one surface and not the
other, and no test notices, because each surface is tested through its own
entry point. `tests/test_cli_contract.py` walks `COMMAND_SURFACE` and asserts
the CLI's envelope contract; `tests/test_engine.py` exercises the engine.
Both stay green while the two drift apart.

There is a live instance on `main`:

```
DexEngine.check            []          <- no object scope
DexEngine.schema_drift     ['objects']
DexEngine.volume_drift     ['objects']
DexEngine.grain_drift      ['objects']
DexEngine.semantic_drift   ['objects']
```

`maintain.check(engine, objects=None)` accepts an object scope and
`maintain check <objects>` exposes it, but `DexEngine.check()` hard-codes the
call and drops it. All four sibling detectors accept `objects`, which is what
makes this an oversight rather than a decision: a library caller cannot narrow
a drift sweep, and a CLI caller can.

The intent is already written down in `engine.py`'s module docstring, and only
there:

> The CLI is the first consumer of this API rather than a parallel
> implementation of it, which is what keeps the two from drifting.

Nothing enforces it, and the docstring is not where a contributor adding a
subcommand is looking.

### Proposal

**State the rule** somewhere a contributor reads before adding a command: the
CLI is a wrapper over `DexEngine`. Every subcommand is reachable from the
Python API, with the same capability.

**Enforce it with a test** over `COMMAND_SURFACE` and `DexEngine`, failing when
a subcommand has no method, or when a method cannot express something the CLI
can. An explicit allowlist carries each CLI-only subcommand with its reason
(`demo` creates a warehouse file and is not a command a library caller drives;
`transform test` is scaffold-only and spelled `test_scaffold`; `semantic *` is
spelled `semantic_*`).

**Close the `check` gap** so it matches its four siblings.

**The test has to compare capability, not flag spelling.** A naive comparison
of argparse `dest` names against method parameters reports mostly false
positives, because the translation between the two surfaces is deliberate:

- argparse reuses one generic `argument` positional across several `transform`
  subcommands; the handler maps it to `plan_id` / `name` / `intent`
- a negating flag pair (`--no-auto-profile`) collapses into one tri-state
  parameter (`auto_profile`)
- `--edits-file` and `--sql-file` become parsed *content* on the API, because a
  library caller holds data rather than a path
- one subcommand can fan out to two methods: `explore semantic` dispatches to
  `semantic_list` or `semantic_query` on its `mode` positional

Two asymmetries are correct and must survive any tidying, because renaming
either breaks every keyword caller for no gain: `obj` rather than `object`
(which is a builtin), and `expression` rather than `expr`.

Optionally, and separately, the handlers could be routed through the engine
methods so parity becomes structural rather than asserted: a subcommand with no
method would then fail to work rather than fail a test. That touches every
handler, so it is worth deciding on its own.

### Acceptance

- A test fails when a subcommand has no `DexEngine` method.
- A test fails when a subcommand exposes a capability its method cannot express.
- The allowlist names every CLI-only subcommand and why it is CLI-only.
- `DexEngine.check` accepts an object scope, matching the four detectors.
- The rule is written where someone adding a subcommand will encounter it.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/344
- `DexEngine` and the CLI are two independent surfaces over the same
  implementation, and nothing kept them in step: each is tested through its
  own entry point, so a capability could exist on one and not the other and
  both suites stayed green.
- `DexEngine.check()` hard-coded no object scope while its four sibling
  detectors (`schema_drift`/`volume_drift`/`grain_drift`/`semantic_drift`)
  all accept one. `maintain.commands.check()` already threaded `objects`
  through; only the engine method dropped it. Fixed: it now accepts and
  passes `objects` like its siblings.
- Building that fix surfaced a second, matching bug: `transform place`
  inherited `--edits-file` from the block meant for its two neighboring
  propagation verbs (`rename`, `remove`), but `cmd_place` never read it, so
  the flag parsed and silently did nothing. Removed.
- Added a parity test over `COMMAND_SURFACE` and `DexEngine` that compares
  capability rather than flag spelling (the translation between the two
  surfaces is often deliberate: a shared `argument` positional means a
  different keyword per subcommand, a negating flag pair collapses into one
  tri-state parameter, a file path becomes parsed content), with an explicit
  allowlist naming every CLI-only subcommand and why (`demo` writes a file
  outside `DexEngine`; `transform test` is scaffold-only and reachable as
  `test_scaffold`, not a method; `viz preview` is not yet implemented). It
  caught the `transform place` bug on its first run.
- The rule ("adding a subcommand means adding it to `DexEngine`, or
  allowlisting it with a reason") is now written next to `COMMAND_SURFACE` in
  `cli.py`, where a contributor adding one will see it.
Before : 
image
After fix : 
image
The tests: 
image


marcociav-exmergo (approved):
LGTM!

PR #386, opened 2026-08-31
BigQuery DATETIME hour-grain continuity now has live coverage

### LINKED ISSUE #315: BigQuery `DATETIME` hour-grain continuity has no live coverage
`is_date_only_type` previously claimed any type whose name contained `DATE` and
not `TIMESTAMP`. That is true of BigQuery `DATETIME` as well as of ClickHouse
`DateTime`, so those columns were read as bare calendar dates and the hour-grain
half of temporal continuity was skipped for them. The column reported day and
month gaps and simply never reported an hour gap, which reads as a clean result
rather than a missing one.

The predicate is fixed and unit-tested, and the behavior is verified live on
ClickHouse against a column with a known gap. **BigQuery has no equivalent live
assertion**, so the connector that carried the bug longest is the one with no
end-to-end proof that it is gone.

Small and worth closing: add a `DATETIME` column with a known hole to the
BigQuery integration fixture and assert the hour-grain span, distinct periods,
and largest gap, matching the shape the ClickHouse suite already asserts.

Worth checking the same predicate against the other connectors' spellings while
in there, since the fix was a substring rule and substring rules are exactly
what produced the original defect.

spascarella:
## Problem

`is_date_only_type` used to claim any type containing `DATE` and not `TIMESTAMP`, which is true of BigQuery `DATETIME` as well as ClickHouse `DateTime`. Those columns were read as bare calendar dates, so the hour-grain half of temporal continuity was skipped: the column reported day and month gaps and never an hour gap, which reads as a clean result rather than a missing one. The predicate is fixed and unit-tested, and ClickHouse proves it live, but BigQuery, the connector that carried the bug longest, had no end-to-end assertion that it is gone.

## Fix

Add BigQuery's own live proof, plus an offline audit of the substring rule across every connector.

- A `bq_continuity_table` fixture writes one 45-row table into the TTL'd scratch dataset and drops it afterward (public data offers no `DATETIME` column with a hole whose size the suite can state). `occurred_at DATE` advances a day per row; `recorded_at DATETIME` advances an hour per row from 00:30, so it is never day-aligned. Three dropped rows give a 3-day hole in one column and a 3-hour hole in the other; four later rows keep their date but carry a NULL `recorded_at`, giving the hour grain a second, wider hole of 4. The two grains therefore report different numbers, so hour statistics that were really the day column's would not pass.
- The test asserts the hour grain end to end (`DATETIME`, hour, span 48, distinct 41, missing 7, largest gap 4) in the shape the ClickHouse suite already asserts, and pins the other direction on the same table (`DATE`, day, 48/45/3/3), since `DATE_TRUNC` is the one BigQuery truncation function with no HOUR unit.
- A new table test holds 44 spellings, every temporal type the seven connectors actually report, against `is_temporal_type`, `is_date_only_type`, and `temporal_units_for`, each row labeled with the metadata source it comes from so a connector that changes how it reports a type has to face the table.
- 
## Related issues

Closes #315 

## Files changed:

- `packages/dex-core/tests/integration/test_bigquery_explore.py:` the fixture, the continuity test, and an updated module docstring (the suite now writes one tiny table).
- `packages/dex-core/tests/adapters/test_temporal_type_predicates.py`: new, the per-connector spelling audit.

### DIFF
--- a/packages/dex-core/tests/adapters/test_temporal_type_predicates.py
+++ b/packages/dex-core/tests/adapters/test_temporal_type_predicates.py
@@ -1,0 +1,141 @@
+"""Every temporal spelling the connectors actually report, held against the
+shared type predicates in one table.
+
+``is_date_only_type`` is a substring rule, and a substring rule is what
+produced the defect it now guards (#188): ``DATETIME`` contains ``DATE`` and
+not ``TIMESTAMP``, so BigQuery ``DATETIME`` and ClickHouse ``DateTime`` were
+read as bare calendar dates and their hour-grain continuity was never
+computed. Nothing about that was loud -- the column kept reporting day and
+month gaps and simply never reported an hour one, which reads as a clean
+result rather than a skipped one.
+
+So the rule is pinned against spellings, not against intent, and each row
+names the metadata source it comes from: that is what has to be re-read when a
+connector changes how it reports a type, and a connector added later owes this
+file its own rows. The live proof that the hour grain is really computed lives
+per connector (tests/integration/test_bigquery_explore.py and
+test_clickhouse_explore.py both profile a column with a known hole); what is
+here is the offline classification every one of them depends on.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from exmergo_dex_core.adapters.base import (
+    TEMPORAL_UNITS_DATE_ONLY,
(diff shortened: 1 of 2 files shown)


marcociav-exmergo (approved):
LGTM

PR #388, opened 2026-08-31
Self-describing query envelope

### LINKED ISSUE #282: `explore query`: result payload shape is not discoverable from a successful envelope

### Problem

`explore query` returns results as `data.columns`, `data.types`, and
`data.cells`. A caller expecting a row-oriented shape (`data.rows`, or
`data.results[].rows`) reads `status: ok` alongside what appears to be an empty
payload, with nothing in the envelope pointing at the right key.

### Why it matters

Minor in isolation, and it cost one wasted round trip on first use.

It compounds with two neighbouring reports from the same session, where three
different envelope keys were guessed wrongly and in each case a **successful**
command looked like an empty one. For a tool whose contract is "read the
envelope and decide the next step", guessable payload shape is part of the
contract rather than a documentation nicety.

The columnar layout itself is the right choice (it is compact and avoids
repeating keys per row) and should stay. Only its discoverability is at issue.

### Proposal

Either of:

- a one-line `notes` entry on the first `explore query` of a session naming the
  payload keys, or
- an explicit `"shape": "columnar"` discriminator in `data`.

### Acceptance

- A caller can determine the payload layout from the envelope alone, without
  consulting documentation or reading the source.

spascarella:
Implemented the self-describing query envelope.
  - Single queries now include data.shape: "columnar".
  - Batch queries include data.shape: "columnar" and data.results[].shape: "columnar".
  - Existing columns, types, and cells layout is unchanged.
  - Updated contract documentation and changelog.
  - Added regression coverage for single and batched responses.
Validation:
- Query tests: 43 passed
- Safety and billed-query tests: 225 passed
- Ruff lint and formatting checks passed

## Related issues
Closes #282.


marcociav-exmergo (approved):
LGTM

PR #389, opened 2026-08-31
SCRUM-1153-Escalate an unrecognized dialect's silent full scan to a warning

### LINKED ISSUE #313: `explore cluster`: an unrecognized dialect silently scans the whole table
`_sample_parts` in `explore/cluster.py` dispatches per dialect and falls through
to `return "", "", "no sample clause (unrecognized dialect)"`. Every shipped
connector has an entry, so nothing is broken today. The problem is what happens
to the next one: a connector added without an entry here does not fail, it
**silently reads the whole table**, bounded only by the cost gate.

That is the failure shape this codebase treats as worse than an error. The note
says "unrecognized dialect", but it is a note on a result that otherwise looks
completely normal, and the cost shows up as a larger bill rather than as a
refusal.

Two candidate fixes, either is fine:

- Make the fallthrough loud: raise, or return a note the caller escalates to a
  warning on the envelope rather than an informational line.
- Assert coverage in the suite, the way other per-connector tables are covered:
  a test that every name in the dialect registry has a sampling entry, so
  omitting one fails at merge rather than at someone's month-end bill.

The second is cheaper and catches it earlier. Worth doing both.

emnarahmouni-exmergo:
Closes : #313
- `_sample_parts` (`explore/cluster.py`) falls through to an informational
  note, `"no sample clause (unrecognized dialect)"`, for any dialect it has
  no case for. Every shipped connector has an entry today, so nothing is
  broken; the risk is the next one: a connector added without an entry does
  not fail, it silently reads the whole table, bounded only by the cost gate.
- The note alone lives inside `notes` and is easy to miss. `explore cluster`
  now escalates it to `warnings` too, so a caller checking cost cannot miss a
  full, unbounded table scan.
- Added a coverage test that walks every dialect in the adapter registry
  against `_sample_parts` directly, so a connector added without a sampling
  entry now fails at merge rather than at someone's month-end bill (the
  cheaper of the issue's two proposed fixes, done alongside the warning).

image

`explore cluster` end to end, with `build_sample_sql` faked to return the
unrecognized-dialect note (no shipped connector triggers it for real).
`warnings` now carries the escalated message; before this fix it was empty
(`[]`) for the exact same run, so the full-table scan showed up only inside
`notes`, easy to miss. Don't forget `git stash pop` to restore the fix if
you reproduce the before state yourself.


marcociav-exmergo (approved):
LGTM

PR #391, opened 2026-08-31
Implemented connection provenance across dex envelopes

### LINKED ISSUE #295: The envelope never states which connection answered
## Problem

A successful dex envelope never states which connection answered it.

The observable consequence, from a 103-trial agent run: callers defensively re-set
connection environment variables on call after call, within a single session, in at
least six syntactic variations.

```
DB_TYPE=duckdb DUCKDB_PATH=/app/database/retail.duckdb dex explore query ...
export DBT_PROFILES_DIR=/app/dbt_models_duckdb && dex explore query ...
export DBT_PROFILES_DIR=... DB_TYPE=... DUCKDB_PATH=... && dex explore query ...
```

That includes `DBT_PROFILES_DIR`, which dex does not read at all, re-exported ahead
of nearly every command in some sessions.

dex was not refusing any of these. The caller was defending against a failure that
was not occurring, because a successful envelope gives them no way to confirm which
database answered.

## Why it matters

#199 argues that dex should not invent a connection target, and that rule is right
and should stay. This is the other half: having correctly resolved a target, dex
should say so. Silence on success is what makes a caller hedge, and the hedging is
not free, in tokens or in the possibility of pointing a later command somewhere
different by accident.

The failure this prevents is worse than verbosity. A caller who cannot tell which
database answered cannot tell when it answered about the wrong one, which is the
exact hazard #199's refusals exist to prevent.

## Proposal

- Include a small `connection` block in every successful envelope: connector,
  resolved target (path, or project and dataset), and where the resolution came
  from, whether that is `.dex/config.yml`, an environment variable, a flag, or a
  directory-local inference.
- Keep it compact enough to be free and constant enough to be ignorable.
- If `DBT_PROFILES_DIR` is genuinely not read by dex, say so in the error and help
  text. Several callers assumed it was.

Adjacent to #199, opposite failure mode: #199 is about refusing too readily, this is
about succeeding without saying what it connected to.

spascarella:
Implemented connection provenance across dex envelopes.
- Added top-level connection with connector, non-secret target, and resolution source.
- Supports flags, .dex/config.yml, environment variables, dbt profiles, host-supplied connections, and directory-local DuckDB inference.
- Added connection-target secret scanning.
- Clarified in help and errors that DBT_PROFILES_DIR locates profiles.yml but does not select or override dex’s connector.

## Related issues
Closes #295


marcociav-exmergo (approved):
Nice, thanks @spascarella

PR #392, opened 2026-08-31
Exhaustion refusal now inside CostGate

### LINKED ISSUE #316: `remaining_for_statement` uses 0 for both "exhausted" and "no limit"
`CostGate.remaining_for_statement` returns an integer because every connector's
server-side cap setting takes one. On the time-paradigm connectors a cap of `0`
does not mean "spend nothing", it means **no limit** (Postgres
`statement_timeout`, ClickHouse `max_execution_time`). So a 0 that reached the
server would remove the backstop at exactly the moment the budget is nearly
spent, and each adapter independently avoids that by checking for a sub-unit
remainder and refusing first.

That works, and it is a convention rather than a contract: it lives in each
adapter's billed-statement path, it is easy to omit in a new one, and omitting it
fails in the most expensive possible direction while looking like a normal run.
A recent defect in this area came from the same ambiguity on the other side, a
legitimate fractional remainder truncating to 0 and reading as exhaustion.

Suggested change: make the two states distinguishable at the boundary rather than
by an adapter-side magnitude check. Either return `None` for "no cap applies" and
reserve 0 for "exhausted", or expose the refusal as its own call so an adapter
cannot forget it. The property to preserve, and to test once centrally rather
than per connector: **no code path can hand a server a cap value that the server
reads as unlimited.**

spascarella:
## What this changes
`CostGate.remaining_for_statement` becomes the private `_remaining_for_statement`, reachable only through a new `CostGate.statement_cap(*, unit, minimum=1) `that raises `OverCeilingError` on exhaustion. Its public result is now either None or a strictly positive, server-safe integer. The six per-adapter exhaustion checks are deleted, and the property they enforced between them is tested once, centrally.

## Problem
The server-side cap is an integer because every connector's cap setting takes one, and on the time-paradigm connectors a cap of 0 does not mean "spend nothing" but no limit: Postgres `statement_timeout`, ClickHouse `max_execution_time`, Databricks `STATEMENT_TIMEOUT`.

Exhaustion and "no cap applies" were already distinguishable at the boundary, 0 against None, but telling them apart was still the caller's job, and all six billed adapters did it the same way in their own billed-statement path. That is a convention, not a contract. One forgotten if in a new connector hands the server a 0, which removes the backstop at exactly the moment the budget is nearly spent, and it fails in the most expensive possible direction while looking like an ordinary run.

## Fix
`statement_cap` owns the refusal, so there is no intermediate value left for a caller to misread. Two parameters carry what the adapters used to hardcode: `unit` names the unit in the refusal in the connector's own vocabulary (a "database-second", a "byte"), since that is what the caller has to raise `--budget` in, and `minimum` is the smallest cap that server can usefully be given. `minimum` is how BigQuery's 10 MB per-query billing minimum is now expressed, rather than as a seventh hand-written check.

The distinction the earlier defect turned on is preserved. Only the ceiling minus actual billing may produce the 0; a per-command booking under one unit still tightens the cap without ever becoming a refusal.

Two behavior changes worth review: the BigQuery refusal now carries its `Cost` (it previously reported an empty cost block on a spend refusal), and its message names the per-query minimum rather than the connector.

Checked repo-wide, not just in the adapters: `transform/build.py` and `transform/init.py` build caps on a separate path that never touches a CostGate, and both already floor at 1, so no change was needed there.

Tests: 807 passing in the adapter and guard suites, 3052 across the full non-integration suite. Three test_packaging.py failures are pre-existing (they reproduce on a clean stash) and stem from a dbt-core-experimental-parser prerelease resolution issue.

## Related issue
Closes #316

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,39 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Changed
+
+- **An exhausted budget is now refused by the cost gate itself, rather than by
+  a check each adapter had to remember to write** ([#316]). The server-side cap
+  is an integer because every connector's cap setting takes one, and on the
+  time-paradigm connectors a cap of 0 does not mean "spend nothing" but *no
+  limit* (Postgres `statement_timeout`, ClickHouse `max_execution_time`,
+  Databricks `STATEMENT_TIMEOUT`). Exhaustion and "no cap applies" were already
+  distinguishable at the boundary, 0 against `None`, but telling them apart was
+  still left to the caller, and all six billed adapters did it the same way in
+  their own billed-statement path. That is a convention, not a contract: one
+  forgotten `if` in a new connector hands the server a 0, which removes the
+  backstop at exactly the moment the budget is nearly spent, and fails in the
+  most expensive possible direction while looking like an ordinary run.
+
+  `CostGate.remaining_for_statement` is now the private
+  `_remaining_for_statement`, reached only through `CostGate.statement_cap`,
+  which raises `OverCeilingError` on the shortfall. Its public result is
+  therefore either `None` or a strictly positive cap the server will honour,
+  with nothing in between for a caller to misread. `statement_cap` takes the
+  `unit` the refusal should name in the connector's own vocabulary (a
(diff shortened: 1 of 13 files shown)


marcociav-exmergo (approved):
LGTM

PR #395, opened 2026-08-31
A budget refusal should cite the historical estimate-to-billed ratio it already has

### LINKED ISSUE #278: A budget refusal should cite the historical estimate-to-billed ratio it already has

### Problem

A `transform build` was refused:

```
estimated cost 6905293058.0 exceeds the ceiling 5000000000.0 (bytes_scanned);
raise the budget or narrow the work
```

Re-run at a raised ceiling, the same build **billed 4.75 GB**. The estimate was
**45% high**, and the refused build would have fit comfortably inside the
original ceiling.

This is systematic rather than unlucky. Dry-run estimates on partitioned and
clustered tables are upper bounds by construction, and dex documents as much:

> BigQuery dry-run estimates on clustered tables are upper bounds that partially
> reflect block pruning; actual billed bytes can differ.

Every large build on such a warehouse will over-estimate, and the ratio is
reasonably stable per project.

### Why it matters

The refusal itself is correct and should stay. A ceiling that can be overridden
by confirmation is not a ceiling.

What the message lacks is any basis for the operator's next decision. "Raise the
budget or narrow the work" invites a guess, and the guess is made under the
natural but wrong impression that the estimate approximates the cost. Guess too
low and the command is refused again; guess too high and the ceiling stops
meaning anything.

dex already holds the answer. `.dex/spend.jsonl` records both the estimate and
the settled figure for every prior command on this connector.

### Proposal

When an estimate exceeds the ceiling, append a calibration line drawn from the
ledger:

```
your last 8 BigQuery builds billed a median 69% of estimate (range 61-88%)
```

Emit nothing when there is insufficient history, rather than inventing a ratio
from two data points.

### Acceptance

- A refusal on a connector with prior settled history carries the observed
  ratio.
- A refusal with no history carries no ratio, and says that rather than
  implying one.
- The ratio is computed per connector, never pooled across connectors (a DuckDB
  history must not calibrate a BigQuery refusal).

spascarella:
## What changes
Over-ceiling refusals now end with what this connector's recent commands actually billed against what they were estimated at. Settlements record the estimate they were admitted on, and SpendHistory (spend_entries) is added as an optional store capability to read them back.

## Problem
A build was refused at an estimated 6.9 GB against a 5 GB ceiling. Re-run at a raised ceiling, it billed 4.75 GB. The estimate was 45% high, and the refused build would have fit comfortably inside the original ceiling.

This is systematic rather than unlucky: dry-run estimates on partitioned and clustered tables are upper bounds by construction. "Raise the budget or narrow the work" gives the operator no basis for the next decision, so they guess, and the guess is made under the natural but wrong impression that the estimate approximates the cost.

## Fix
The ledger had to gain the estimate first. It previously held one only as a reservation amount, and reservations are written only when budget.session_ceiling is set; transform build recorded none either way. Settlements now carry estimate beside the billed figure.

On refusal the guard reads that history back and appends the observed ratio:

The last 8 settled bigquery commands in this project's spend ledger billed a median 70% of estimate (range 61%-88%) ... admitting this command takes a budget above 6,905,293,058, at which it would be expected to bill around 4,868,231,606 bytes_scanned.

Scoped per connector and never pooled, matched on every entry, with a paradigm filter one level down so a seconds history cannot divide a bytes estimate. One command is one data point, not one statement. Below three settled commands it says it has no ratio instead of inventing one, and distinguishes "none yet" from "only 2, too few".

## What still stands

- The refusal. Unchanged, and still cannot be confirmed through.
- The ceiling binds on the estimate, not on what settles. The clause says so and names the budget that actually admits the command, because a caller reading only the ratio would set one at 70% of the estimate and be refused again by arithmetic.
- A known, documented bias. Killed commands (a reservation with no release) and zero-billed cache hits are excluded. Errored commands settle and release like any other and are not detectable, so they are counted. The number can read low, never high.
- Backends without the capability. They lose the sentence and nothing else, and nothing warns, since no guard is narrower than it looks. An unreadable ledger is swallowed rather than converting a correct refusal into a different error.
- Existing projects. Entries written before this carry no estimate, so a project sees the no-history sentence until it has settled enough new commands.

## Related issue
Closes #278 

## Note
Close #394 first


marcociav-exmergo (approved):
LGTM

PR #396, opened 2026-08-31
SCRUM-1116-maintain grain: damp lost-uniqueness severity on small relations

### LINKED ISSUE #280: `maintain grain`: high severity on relations too small for uniqueness to mean anything

**Depends on:** #218 (for the boolean-shape detection reused below)

### Problem

One `maintain check` produced six **high**-severity grain findings. All six were
artifacts of the data's shape. None was a defect:

- **Three on a 4-row table**, against boolean and status columns: "2 distinct
  over 4 rows (~2 duplicate rows); joins on it will fan out". Those columns were
  only ever unique because the table previously held 2 rows.
- **Two on a change-data-capture changelog's document identifier**: 4 distinct
  over 13 rows. Repeated identifiers are the entire design of a changelog, and
  the staging model deduplicates them.
- **One on a changelog's `(id, operation)` composite** at 2,004 distinct over
  2,048 rows, where no model in the project keys on that composite.

### Why it matters

High severity is the signal a triager reads first. Six false highs per run on a
project of this size trains people to skim the axis most likely to carry a real
defect, and the failure mode of drift detection is being ignored rather than
being wrong.

The three cases have different causes and each is separately addressable:

1. A **small-sample artifact**. At 4 rows, any column with 2 distinct values
   "loses" a uniqueness it never meaningfully had.
2. A **category error**. A relation whose purpose is one row per mutation event
   should not be judged against a per-entity uniqueness expectation.
3. A **composite nobody declared**, where the finding is true but inert.

### Proposal

Three independent damping rules. Any one helps on its own.

1. **Row-count floor.** Below a threshold (100 rows is a reasonable default,
   configurable), do not emit a uniqueness regression above `low`. The envelope
   already carries the row count used in the verdict, so this needs no new
   measurement.
2. **Column shape.** Never emit `high` for a lost-uniqueness finding on a column
   whose profiled distinct count is 2 and whose name or type reads boolean.
   #218 builds exactly the boolean-shape detection this would reuse.
3. **Append-only relations.** Where the cache or the project marks a source as a
   changelog or event stream, judge grain against the declared deduplication key
   rather than against every candidate key.

### Acceptance

- A 4-row relation whose boolean column holds 2 distinct values produces no
  `high` finding.
- A finding suppressed by the floor is still visible at `low`, carrying the
  reason it was damped, so nothing is silently dropped.
- The threshold is configurable in `.dex/config.yml`.

emnarahmouni-exmergo:
Closes : https://github.com/exmergo/dex/issues/280
- One `maintain check` produced six high-severity grain findings on a real
  project, all six artifacts of the data's shape and none a defect: three on
  a 4-row table's boolean and status columns (only ever unique because the
  table once held 2 rows), two on a change-data-capture changelog whose
  repeated identifiers are the entire design, and one on a composite nobody
  declared. High severity is the signal a triager reads first; six false
  highs per run trains people to skim the axis most likely to carry a real
  defect.
- A lost-uniqueness finding (`key_lost_uniqueness` / `declared_grain_not_unique`)
  is now damped to `low` rather than `high` below `maintain.grain_min_rows`
  rows (default 100, new in `.dex/config.yml`): on a handful of rows, losing
  uniqueness means the least, and a 4-row table's boolean column "loses" a
  uniqueness it never meaningfully had once a fifth row repeats a value.
- Damped, never dropped: the finding still reports, just not at the severity
  that trains a triager to stop reading. The damping is named in the
  finding's own `data` (`severity_floor_applied`, `grain_min_rows`) and in
  its prose, so nothing about the run is silent.
- Applies uniformly to all three sites that emit a uniqueness-regression
  finding (single-key, composite, declared-composite); `join_orphans_increased`
  is untouched since it already grades severity from the measured orphan
  fraction rather than asserting `high` unconditionally.
- The issue's other two proposed damping rules (never `high` on a
  boolean-shaped 2-value column regardless of row count; judge an
  append-only/changelog relation against its declared dedup key) are
  deferred: the issue frames all three rules as independent, and the
  row-count floor alone satisfies every acceptance criterion.

image
`active` on a 4-row table would previously report `severity='high'`; it now
reports `low`, with the damping named right in the data
(`severity_floor_applied`, `grain_min_rows`) rather than silently applied.


marcociav-exmergo (approved):
LGTM!

PR #397, opened 2026-08-31
Fix data.bytes_billed present on transform build, absent on maintain check 

### LINKED ISSUE #276: `data.bytes_billed` is present on `transform build` and absent on `maintain check`

spascarella:
## What changes
`data.spend` becomes the single canonical location for settled spend on every billed command. A contract test enforces key parity across transform `build, maintain check, explore query, explore map, and explore profile`.

## Problem
`transform` build reported its billed magnitude twice, once under data.spend like every other billed command and once as a top-level `data.bytes_billed` that no other command carried. A caller reading the top-level key saw a build's spend and read maintain check's missing key as zero, so a command that had just scanned 0.89 GB was reported as free. That happened in a real session and survived into the first draft of a written cost report.

A build that settled at zero also omitted `data.spend` entirely, which is the same ambiguity one level down.

## Fix

- `_shape_build_result` pops dbt's per-node billing figure instead of leaving it in the payload, so `data.spend` is the only place spend appears.
- A billed paradigm always reports the unit key, whatever it settled at. Zero is reported as zero.
- Where dbt executed statements and reported no billing figure for any of them, the key reports `null` with a note and nothing reaches the ledger. Rounding an unknown down to zero would be the same under-report by another route.
- ClickHouse Cloud's translated keys (compute-unit-hours, USD) are emitted at zero seconds too, matching what explore always reports.
- The build's ledger read-back at settlement is best-effort, reporting `session_spent_today`: null as gate settlement already does, rather than turning a build that already spent into a failure that reports nothing.
- Also updated: the four integration tests that read the removed key, the `data.spend` bullet in `references/command-contract.md`, the spend paragraph in `AGENTS.md`, and a `Fixed changelog entry`.

Full unit suite green (3106 passed). The new parity test was verified to fail against the pre-fix code with transform build reports spend outside data.spend: ['bytes_billed'].

## What still stands

- The removal is a breaking envelope change for any caller reading `data.bytes_billed` or `data.seconds_billed` on transform build. That read is the defect, so it is not preserved.
- `connect` test still reports `session_spent_today` nested under `data.budget`. That command bills nothing and the contract documents the placement, so it is out of scope here.
- `explore`'s preflight estimate keys (for example ClickHouse's `estimated_compute_unit_hours`) are unchanged. They are estimates on the confirmation ask, not settled spend.

## Related issue
Closes #276 

## Note
Close #395 first

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -192,7 +192,11 @@ still binds exactly via a server-side statement timeout, except on ClickHouse,
 where it binds via `max_execution_time` **and** `max_bytes_to_read`, because
 time alone is checked only at block boundaries there. Actual spend comes
 back under `data.spend` (`bytes_billed` or `seconds_billed`) and accumulates
-in the `.dex/spend.jsonl` ledger per connector. That ledger gates billing and
+in the `.dex/spend.jsonl` ledger per connector. That is the only place spend is
+reported: every command that can bill carries the unit key whatever it settled
+at, zero included, and no command puts a billed magnitude anywhere else in
+`data`, because a key present on one command and absent on another reads as a
+spend of zero rather than as a key to look for elsewhere. That ledger gates billing and
 nothing else: it is read where work is admitted, not where a connection is
 assembled, so a command that cannot spend does not depend on it, and a ledger
 that cannot be read refuses billed work by name while reporting the day's total
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -110,6 +110,56 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   as independent, and the row-count floor alone satisfies every acceptance
   criterion.
 
+- **`data.spend` is now the only place any command reports what it billed**
+  ([#276]). `transform build` stamped its billed magnitude twice, once under
+  `data.spend` like every other billed command and once as a top-level
+  `data.bytes_billed` that no other command carried:
+
+  | Command | `data.bytes_billed` | `data.spend.bytes_billed` |
+  |---|---|---|
+  | `transform build` | 4750049280 | 4750049280 |
+  | `maintain check` | absent | 891289600 |
+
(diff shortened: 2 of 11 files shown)


marcociav-exmergo (approved):
LGTM

PR #399, opened 2026-09-01
Cache-backed explore query annotations

### LINKED ISSUE #293: `explore query` returns rows with none of the column intelligence dex already holds
## Problem

`explore query` returns cells. It says nothing about the columns it just returned,
even when those columns belong to a relation dex has already profiled and cached in
the same session.

That matters more than it sounds, because `query` is where callers actually live.
In a 103-trial agent run against an unfamiliar warehouse, `explore query` was the
most-reached-for subcommand, ahead of `explore profile`, despite project
instructions that named `profile` explicitly and did not mention `query` at all.

## Why it matters

This is the "meet the caller where they are" issue. Whatever the intended entry
point, the observed entry point is `query`, and it is the one surface that carries
no dex judgment whatsoever. A caller who lives in `query` gets a SQL runner and
never encounters the part of the product that is not a SQL runner.

The fix is close to free because the facts are already computed and cached. A caller
who selects a column that is 40% NULL, or groups by a column that is not the grain,
can be told so at the moment they do it rather than in a separate command they did
not run.

## Proposal

- When a result's columns resolve to columns of a profiled relation in the cache,
  attach a compact `column_notes` block: null fraction where non-zero, PII flags,
  and whether the selected columns cover a known grain. Cache-only, no extra
  warehouse work, silent when nothing is known.
- When the query groups or joins, note whether the grouping keys match a known grain
  and whether a join key pair has verified overlap, reusing what
  `relationships --verify` already computes.
- Keep this strictly additive and off the hot path. `query` returning fast and plain
  is the right default; the notes should be a small block, not a second profile.

Depends on #282, which settles the shape of this envelope. #282 is about the shape
of the result payload; this is about what is missing from it. Best filed after #282
lands.

spascarella:
Implemented the cache-backed explore query annotations.

## What changes:
- [commands.py](/Users/saasyp/Desktop/exmergo/dex/packages/dex-core/src/exmergo_dex_core/explore/commands.py): adds cache-only column_notes and query_notes after result shaping.
- [results.py](/Users/saasyp/Desktop/exmergo/dex/packages/dex-core/src/exmergo_dex_core/explore/results.py): adds optional result fields, omitted when empty for both single and batch query payloads.
- [test_query.py](/Users/saasyp/Desktop/exmergo/dex/packages/dex-core/tests/explore/test_query.py): covers null-fraction notes, selected-grain coverage, grouping-vs-grain notes, verified join overlap notes, and silent derived/batch outputs.

## Related issue
Closes #293


marcociav-exmergo (approved):
LGTM!

Issue #233, opened 2026-08-05
`maintain`'s description does not trigger on the diagnostic intent


marcociav-exmergo:
**Depends on:** #224, #231

### Problem

`maintain`'s trigger examples are all drift-shaped: "what changed in the
warehouse", "did anything drift", "the row count dropped", "which models are
stale". Each presupposes a baseline and a change.

The request that actually arrives most often presupposes neither:

- "this model is wrong, can you work out why"
- "the numbers look too high"
- "some rows are missing from this table"
- "the project is broken"
- "this column is empty and it should not be"
- "check my models against the warehouse"

None of the three skill descriptions contains *fix*, *broken*, *wrong*, *missing
rows*, *duplicates*, *why*, or *diagnose*. So the request lands nowhere: it is not
exploration (a project exists), not authoring (the model exists), and not drift
(there is no baseline).

### Proposal

Once #224 and #231 exist, update `maintain`'s description to claim the
diagnostic intent in the words the request actually uses, and restate the skill's
scope as "is this project correct", with drift as one of its two modes.

The negative boundaries need restating too, since the widened claim now overlaps
`transform` more than it did: authoring belongs to `transform`, sense-making with
no project belongs to `explore`, and correctness of an existing project belongs
here.

### Ordering, and why it matters

**This must not land before the capability.** A description that triggers a skill
which then refuses for want of a baseline is worse than one that does not trigger:
it costs the caller a round trip and teaches them the tool does not work. File it,
block it on #224 and #231, and land it in the same release.

### Acceptance

- The description names the diagnostic intent explicitly.
- The triggering eval suite covers the phrasings above as positives.
- The negative sets keep the three skills mutually exclusive.
- The `npx skills install` frontmatter parse test still passes, since the
  description is parsed by tooling outside this repository.

Issue #231, opened 2026-08-05
`transform build --verify`: run the sweep over the nodes just built


marcociav-exmergo:
**Depends on:** #225, #226, #227

### Problem

`transform build` runs dbt against a dev target, parses `run_results.json`, and
reports node status. A caller who reads `"success": true` learns that dbt executed
without erroring. They do not learn whether the relations it produced are right,
and that is the question they asked.

### Proposal

Add `--verify`, which runs the #224 sweep scoped to the nodes this build touched,
and folds the findings into the same envelope as the build result.

Design points that matter:

- **Scoped to what was built.** The whole project is a different and more expensive
  question, and `maintain verify` is where that lives.
- **Findings do not fail the build.** A build that dbt completed is a build that
  completed. Findings are reported alongside, and whether they should gate anything
  is a policy decision for a caller, not a default.
- **Priced with the build.** On a metered connector the verification scans are part
  of the same estimate and the same handshake, so the caller confirms one number
  rather than two.
- **Default.** On `free_local` this should probably be on by default, since it
  costs nothing and the whole point is that verification should not be something
  you remember to ask for. On metered connectors it must be opt-in. Worth deciding
  explicitly.

### Why this entry point matters more than the standalone command

Verification that requires a caller to know it exists, and to choose it, is
verification most callers will not run. Attaching it to the command they already
run to validate their work is what makes it routine, and it is what makes
`transform build` strictly more useful than invoking dbt directly, which is the
comparison it is actually up against.

### Acceptance

- `transform build --verify` returns build results and findings in one envelope.
- Findings are scoped to the built nodes.
- A build that succeeds with findings is still `status: ok`, with the findings
  surfaced in `warnings` or a dedicated field, not in `errors`.
- On a metered connector, one estimate covers build and verification.
- Cost paradigm and default behavior are documented per connector.

marcociav-exmergo:
Field evidence for the reporting half of this, from a dex 1.6.3 / BigQuery
session.

A full `dbt build` came back green:

```json
"counts": {"success": 38, "pass": 246, "warn": 17}
```

Correctly green. But the envelope gives no way to learn **which** 17 tests
warned, so it cannot answer the only question a green-with-warnings build
actually raises: *are these the same 17 as last time, or is one of them new?*
Reading `target/run_results.json` directly was the only route.

#225 covers nodes that failed or were skipped. Warn status is neither, and on a
project that deliberately runs relationship tests at `severity: warn` (documented
[...]

Issue #172, opened 2026-07-29
spec: declare when an axis is structurally inert instead of returning a clean result


marcociav-exmergo:
**Framing issue.** It asks for nothing to be built immediately, and every instance it names is separately filed and separately fixable. What is worth writing down is that they share a shape, and that the shape has a single contract-level fix.

## The class

> **A caller cannot tell the difference between "ran and found nothing" and "could not have found anything."**

Each instance returns a successful envelope. None warns. In several the honest answer is *this axis is structurally inert for your project*, and the observable output is indistinguishable from a clean bill of health.

## The instances, all filed separately

- **A declared relationship can never be verified** (#163), so `maintain grain`'s join-fanout half cannot run for a project that declares its joins. The route returns `ok` with uniqueness findings only, which reads as a clean join graph.
- **`transform_layer` counts are fingerprinted and never diffed** (#164), so a snapshot reports three counts that look like a tracked baseline and are not one.
- **A two-column grain can be unreachable by measurement** (crowded out of the candidate set by the pair cap, #168) **and inexpressible by declaration** (single-column `DeclaredKey`, #169), producing a confident wrong election rather than an abstention.
- **`explore cluster`'s `dropped_null_rows` is structurally near-always zero** (#160), because the null filter runs in SQL one layer below the counter. A completeness field reports a confident `0` while rows were dropped.

## Plus one the reporter owned and fixed, which is what prices the class

Blast-radius attribution matches a warehouse table to a model by name. An externally managed vendor export in their warehouse had no same-named model, so its drift findings reported `impacted_models: []`, **the identical value a genuinely isolated table returns**, while the table one hop downstream resolved a full radius. It was the root of a whole cost pipeline and the only table in the project they do not write themselves, which is precisely the case where a blast radius earns its keep.

Silent for months. Found by accident while proving something else. The fix, once they knew, was two hours.

**The cost of this class is not the fix. It is the interval before anyone suspects there is something to fix.**

## The structural point underneath that one

`_models_by_table` seeds `by_table` from two loops that union **unconditionally**, with no `if` / `else` between them. So there is no point in the function where "nothing matched this table" is a distinguishable state.

**The information needed to warn is discarded by construction**, which is why this is raised as a contract question rather than a request for a warning. You cannot warn about a miss the function cannot observe.

## The ask

A caller-visible signal that an axis could not run for this project. Two shapes, either is useful:

[...]

PR #402, opened 2026-09-01
SCRUM-1074-maintain verify: baseline-free build-status findings (#224, #225)

### LINKED ISSUE #224: Epic: a baseline-free `maintain verify` that answers "is this project correct now"
**Depends on:** nothing. **Blocks:** #225 through #231, #233.

### Problem

`maintain` detects **drift**: it compares the warehouse and the project against a
`.dex/snapshot.json` baseline and reports what changed. Without a baseline, every
axis refuses:

```
$ dex maintain check
no drift baseline yet; run `maintain snapshot` first
  (ideally right after a known-good build)

$ dex maintain schema
no drift baseline yet; run `maintain snapshot` first
  (ideally right after a known-good build)
```

Verified on 1.5.1, and correct: you cannot report a change against nothing.

But it means dex has no answer to the most common question asked of an analytics
project, which is not "what changed" but **"is this right"**. That question arrives
with no baseline, because the project was often never correct: a model was written
wrong, a join was written against the wrong key, a filter was never applied. There
is no earlier good state to diff against.

The three skills between them cover an unfamiliar warehouse with no project
(explore), authoring a project (transform), and a project that has drifted from a
known-good baseline (maintain). A project that exists, builds cleanly, and is
wrong falls between all three.

### Evidence that callers look for this and do not find it

The refusal `no drift baseline yet; run maintain snapshot first` is the most
frequent refusal in the transcripts archived under
`benchmarks/ade_bench/experiments/`, at thirteen occurrences across the runs. In
each, a caller reached for `maintain check` as "tell me what is wrong with this
project" and got a correct answer to a different question.

### What is missing, concretely

dbt already answers "did it run" and "do the declared tests pass". Neither answers
"is the result right", and the gap between them is where the expensive errors live:
a join on a key with no overlap, a model that lost a tenth of its rows to an inner
join, an output column that is entirely NULL, a filter applied one layer too late.
Each is measurable from the warehouse, and none is measurable from
`run_results.json`.

Today `transform build` reports node status parsed from `run_results.json` and
stops. **dex verifies that dbt ran, not that the model is right.**

### Proposal

A read-only sweep with no baseline requirement, over the current project and the
current warehouse, emitting findings in the shape `maintain` already uses (ranked,
typed, `exact` flagged when a verdict rests on an approximation).

Two entry points, deliberately:

1. `maintain verify [<selector>]`, which works on a project as it stands,
   including after a `dbt run` that dex did not perform.
2. `transform build --verify` (#231), so verification is part of authoring
   rather than a separate thing to remember.

The finding set is specified in #225 through #230, each independently
shippable. This epic owns three cross-cutting decisions:

**The free versus gated split**, which must be declared before any of it ships:

| Half | Cost |
|---|---|
| project parse, compiled manifest, `run_results.json`, declared contracts | free everywhere |
| relation row counts | free metadata on most connectors |
| join overlap probes, NULL fractions, grain probes | scans, gated, through the normal handshake |

So `maintain verify` on a metered connector returns its free findings immediately
and asks once for the scanning half, and on DuckDB all of it is free.

**Inertness.** Where a finding class cannot run (no compiled manifest, no built
relations, a selector matching nothing), it must say so rather than returning
clean. This is the same requirement as **#172** and should share its mechanism.

**Scope of the change.** This widens `maintain` from "what drifted" to "is this
correct", with drift as one of its two modes. That is a deliberate change to what
the skill claims, and #233 is where the claim gets updated. It should not become
a fourth skill: the three skills share one description budget, and a fourth
competes with the three for the attention that decides whether any of them trigger
at all.

### Acceptance for the epic

- `maintain verify` runs against a project with no `.dex/snapshot.json` and
  returns findings.
- Every finding class declares which side of the free-versus-gated split it is on.
- A class that cannot run says so, per #172's mechanism.
- Detection stays read-only. Nothing here proposes edits; `maintain reconcile`
  remains the only thing that does.

### LINKED ISSUE #225: `maintain verify`: report nodes that failed, were skipped, or never built
**Depends on:** #224

### Problem

A build that partially fails leaves some relations stale and some absent, and
downstream work proceeds against whichever survived. Nothing surfaces the gap after
the fact: the run output has scrolled away, and the project looks like a project.

### Proposal

From the compiled manifest and `run_results.json`, report:

- nodes with a failure status
- nodes skipped because a parent failed, with the parent named
- nodes present in the manifest with **no corresponding relation** in the warehouse
- nodes whose relation is older than a parent's, so the model is stale relative to
  its own input
- the project failing to parse or compile at all, named to the offending node

The last is the cheapest and most valuable of the five: a project that does not
compile should be the first thing said, before any other finding, because every
other finding computed against it is unreliable.

### Acceptance

- A project with one failed model and two skipped children reports all three and
  the causal link.
- A model in the manifest with no relation is reported.
- A project that does not compile reports that first, and suppresses findings that
  would be computed from a stale manifest.
- All of it is free on every connector.

emnarahmouni-exmergo:
## This PR implements #224 and #225 together, deliberately

Closes #224. Closes #225.

#224 is an umbrella issue: it has no code of its own, only architectural
requirements (a baseline-free `maintain verify` command, a declared
free-versus-gated cost split, an inertness mechanism so a check that could
not run says so). Its own acceptance criteria include "runs against a
project with no `.dex/snapshot.json` and returns findings" -- which is
impossible to satisfy with zero finding classes. Shipping #224 alone would
mean a command that runs but can never find anything: a scaffold, not a
feature. #225 is the smallest finding class that could go first (it needs no
join analysis, no grain heuristics, no gated-scan pricing, unlike #226-230),
so building it alongside #224 is what makes the epic's acceptance criteria
actually true rather than technically true.

## Summary
- Every existing `maintain` subcommand answers "what changed", and refuses
  outright with no `.dex/snapshot.json` baseline to compare against. That
  has no answer for a project that was never correct in the first place (a
  model written wrong, a join against the wrong key). The refusal `no drift
  baseline yet; run maintain snapshot first` was the most frequent refusal
  across a 103-trial agent benchmark's archived transcripts (#224).
- New `maintain verify [<selector>]`: a read-only sweep with no baseline
  requirement. This PR ships its first finding class (#225):
  - nodes that failed to build
  - nodes skipped because a parent failed, naming it, and walking back
    through however many transitively-skipped parents it takes when the
    immediate parent was itself only skipped
  - models the project declares that have no relation in the warehouse
  - a project that fails to compile, reported first and suppressing every
    other check, since a finding computed from a manifest a broken project
    could not have produced honestly is not a finding at all
- All of it is free: the manifest and the last run's `run_results.json` are
  read off disk, and the relation check reads only cheap object-metadata
  listing, never a scan.
- `data.suppressed` names every finding class that did not run and why (no
  run results yet, no project, an unreachable warehouse, a compile
  failure) -- the "inertness" mechanism #224 asked for and #172 never
  actually built (#172 was closed as framing-only), so an empty
  `data.findings` from a degraded run is never mistaken for a clean project.
- New `DexEngine.verify()`, full CLI wiring (`COMMAND_SURFACE`, subparser,
  dispatch table).
- Deferred to their own issues, each independently shippable: row-loss/fanout,
  NULL columns, join-overlap, and grain findings (#226-230), folding
  verification into `transform build --verify` (#231), and updating the
  `maintain` skill's own description for this diagnostic intent (#233).

image
Three scenarios run against the real CLI, one per #225 acceptance bullet:
a failed model with two skipped children (both independently reported,
both naming the failed parent as the cause), a model declared with no
matching warehouse relation, and a project that fails to compile
(suppressing the other two checks and naming why in `data.suppressed`,
the inertness mechanism #224 required).


emnarahmouni-exmergo:
@marcociav-exmergo #224 has no code of its own  it just needed #225's finding class to make its acceptance criteria real, so I did both in one PR

marcociav-exmergo (approved):
LGTM, thank you @emnarahmouni-exmergo

PR #403, opened 2026-09-01
Fix BigQuery integration test for #288's column elision

emnarahmouni-exmergo:
`explore profile`'s column summarization (#288) elides columns with no
finding by default. The Shakespeare integration test asserted the full
schema without `--columns all`, so `word_count` (no PII/null/key/data-quality
finding) dropped out and the live BigQuery suite started failing.

## Fix
Added `--columns all` to the confirmed-run call in
`test_bigquery_explore.py`, restoring the full schema this test actually
checks.


marcociav-exmergo (approved):
LGTM!

Issue #283, opened 2026-08-11
Prompt rather than warn on the first billed command when `budget.session_ceiling` is unset


marcociav-exmergo:
**Environment:** dex 1.6.3, BigQuery connector.

### Problem

With `budget.session_ceiling` unset, every billed command emits:

> no cumulative spend ceiling: `budget.session_ceiling` is unset in
> `.dex/config.yml`, so this command was bound by its own budget alone and
> nothing bounds the day's total across commands.

The warning is accurate and well worded. It is also the **default state**, and
it repeats on every billed command, which is the condition under which warnings
stop being read.

In one observed session, five billed commands ran to 6.60 GB bound only by
per-command caps, each one carrying the warning, and the aggregate was never
bounded by anything.

### Why it matters

`budget.ceiling` is *refused* when missing, on the stated grounds that nothing
runs unbudgeted. `budget.session_ceiling` is only *warned* about, on the stated
grounds that refusing would break every project that never set one.

Both positions are defensible on their own. The gap between them is where an
unbounded day lives, and the current default puts every new project in that gap
silently.

### Proposal

On the **first** billed command in a repo root with no `session_ceiling`, return
`needs_confirmation` naming a suggested ceiling (a small multiple of the current
command's estimate is a reasonable starting point), and offer to write it into
`.dex/config.yml`.

Once the ceiling is set, or the prompt is explicitly declined, never ask again
in that repo. That preserves today's behaviour for anyone who wants it while
making the unbounded case a decision rather than a default.

### Acceptance

- The first billed command in a repo with no cumulative ceiling prompts once.
- An explicit decline is recorded and not re-prompted.
- No behaviour change for a repo that already has a ceiling set.

PR #404, opened 2026-09-01
Fix DexEngine/CLI parity gap for --columns and session-ceiling flags

emnarahmouni-exmergo:
## Summary
Two gaps the #344 parity test (`test_cli_contract.py`) itself caught once
this branch met current `main`:
- `explore profile --columns all` reached the CLI (#288) but not
  `DexEngine.profile()`, which was missing `show_all_columns`.
- `--session-ceiling`/`--no-session-ceiling` (#283) are connection-level
  flags on every subcommand, and the parity test's own exclusion list
  predated them, so 30 subcommands failed the check.

## Fix
Added `show_all_columns` to `DexEngine.profile()`; added both flags to
`_CONNECTION_DESTS`.

### DIFF
--- a/packages/dex-core/src/exmergo_dex_core/engine.py
+++ b/packages/dex-core/src/exmergo_dex_core/engine.py
@@ -847,6 +847,7 @@ class DexEngine:
         refresh: bool = False,
         use_project: bool = False,
         check_cumulative: bool = False,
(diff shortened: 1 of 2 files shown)


marcociav-exmergo (approved):
LGTM

PR #418, opened 2026-09-02
explore profile: drop fields null on every column, cap each value domain

### LINKED ISSUE #290: `explore profile`: fields that are null for every column still cost a slot per column
## Problem

On a reference `explore profile` payload of a 107-column staging table, each column
carries a 12-field object. Several of those fields are null on most or all columns:

| Field | Null on |
|---|---|
| `pii_overridden` | **107 of 107** |
| `pii` | 100 of 107 |
| `value_domain` | 62 of 107 |
| `min_value` | 61 of 107 |
| `max_value` | 61 of 107 |

`pii_overridden` is emitted 107 times and carries no information in any of them. The
populated `value_domain` blocks account for 7,384 bytes, 19% of the 38,651-byte
payload on their own.

`min_value` and `max_value` are null on 61 of 107 columns because they are not
computed for VARCHAR. That is a defensible choice, but range is precisely what a
caller needs when deciding a population boundary.

Measured on 1.6.3, DuckDB connector.

## Why it matters

This is the cheapest available reduction in payload size, and it compounds with #288
rather than competing with it. A caller paying 38 KB should not be paying any of it
for a field that is null everywhere.

## Proposal

- Omit any per-column field that is null for every column in the dataset, and note
  the omission once at dataset level (`"suppressed_fields": ["pii_overridden"]`) so
  the absence is legible rather than mysterious.
- Cap `value_domain` at a configurable number of values with a total, rather than
  enumerating every distinct value. A 4-value enum is useful; a 200-value one is a
  column dump.
- Separately, consider computing `min_value` and `max_value` for VARCHAR and
  date-typed columns, where lexical or temporal range is cheap and directly answers
  "what period does this table cover".

Nothing here changes what dex computes, only what it serialises.

Parogar:
## Summary

Closes #290 (Jira SCRUM-1124).

`explore profile` serialized every one of a column's 17 fields whether or not it held anything, so a field like `pii_overridden` was emitted as `null` once per column on every table. Two payload-only changes:

- **Fields null on every column are dropped and named once.** Any per-column field that is null on every column the payload serializes is removed from each of them and listed in a new dataset-level `suppressed_fields`, placed ahead of `columns` so a truncated read still explains the absence. The judgment is over the columns actually shown, so the payload is consistent with itself under both the default summary and `--columns all`. A field null on some columns but not all stays on every column. `suppressed_fields` is always present; an empty list means the columns shown carry their full shape.
- **Each `value_domain` is capped.** A new `profile_value_domain_cap` in `.dex/config.yml` (default 25, the probe's own cap, so nothing changes until a repo lowers it) keeps the most frequent values and folds the rest into the domain's existing `elided` count, so `values` plus `elided` is still the exact distinct total.

Neither touches the cached profile or `ProfileResult.datasets`; only `data()` changes.

**Not taken up:** the issue's separate suggestion to compute `min_value`/`max_value` for VARCHAR columns. That suppression is a guardrail (a string extreme is a raw value, and no raw value crosses the envelope), and date/timestamp columns already report their range. Noted in the changelog entry.

## Measured

On the seeded demo warehouse (7 tables, 36 columns), clean `main` spent 6,132 of 18,165 payload bytes (34%) on fields that were null on every column of their table; `pii_overridden` was null on 36 of 36 and the five `temporal_*` fields on 35 of 36. With this change the profile payload is 26% smaller with `--columns all` and 23% smaller by default, and no remaining field is null on every shown column.

## Tests

- Four new tests in `tests/explore/test_explore.py`: the suppression contract end to end, the judgment over serialized columns (with the cache untouched), the value-domain cap with its `elided` arithmetic, and the config key read from `.dex/config.yml`.
- Three existing tests updated to the new contract: the empty-table profile (no rows leaves every statistic null) and the two `pii_overrides` tests (the override clears the only flag, so `pii` is suppressed and named rather than repeated).
- `ruff check` and `ruff format --check` clean across the repo; Markdown passes `scripts/check_no_em_dashes.py`.
- Full suite on Windows: 71 failures that reproduce identically on clean `main` (dbt build and path-separator tests), zero attributable to this branch.

## Docs

`references/command-contract.md`, `AGENTS.md`, and a `CHANGELOG.md` entry under Unreleased.


marcociav-exmergo (approved):
LGTM, thanks @Parogar

PR #420, opened 2026-09-03
Implemented maintain snapshot --project-only

### LINKED ISSUE #281: `maintain snapshot --project-only`: re-pin the project layers without re-profiling the warehouse

**Relates:** #164

### Problem

A repo-wide refactor moved every model file into a new directory and renamed the
dbt project. **No warehouse object changed**: same tables, same views, same
columns, same rows. Only file paths and the project name moved.

Re-pinning the baseline after that change requires the documented discipline:
`rm .dex/cache.json`, then `explore map --full`, then `maintain snapshot`,
because a partial map or carried-forward profiles poison the baseline.

Priced on this project, that `explore map --full` estimated **8.69 GB**.

8.69 GB of warehouse profiling to re-record a list of file paths.

### Why it matters

The cost is disproportionate enough that the rational choice is to skip it,
which is what happened: the session ended green with a baseline 363 hours old.

A discipline too expensive to follow after a routine refactor is a discipline
that decays, and a drift baseline is worth exactly what its freshness is worth.

Project-side refactors (renames, moves, layer reorganisations) are common, and
they are precisely the case where the warehouse half of the baseline is provably
unchanged.

### Proposal

`maintain snapshot --project-only`:

- re-reads the transform and semantic layers from the project files,
- carries the existing `warehouse` block forward untouched,
- preserves its original `cache_updated_at` and `warehouse_from` so staleness
  stays visible and honest rather than being laundered into looking fresh,
- opens no connection, and therefore bills nothing.

Refuse to combine it with any flag implying warehouse state, and say plainly in
the envelope that the warehouse half was carried forward rather than
re-measured.

This pairs with #164 from the opposite side: that issue observes the transform
layer is fingerprinted and never diffed, while this one observes the fingerprint
cannot be cheaply refreshed. Landing them together gives the layer both a reason
to be current and a way to become current.

### Acceptance

- `maintain snapshot --project-only` bills zero bytes and opens no connection.
- The resulting baseline's `warehouse` block is byte-identical to the prior one.
- `cache_updated_at` and `warehouse_from` are preserved, not refreshed.
- A subsequent `maintain check` against an unchanged warehouse reports no schema,
  volume, or grain drift.

spascarella:
## What changes
- Adds maintain snapshot --project-only.
- Rebuilds project-side baseline data from dbt/semantic files while carrying the prior warehouse snapshot forward unchanged.
- Preserves the original cache_updated_at and warehouse_from values so warehouse staleness remains visible.
- Skips warehouse dialect loading and adapter access for project-only snapshots.
- Rejects --project-only when combined with warehouse-targeting flags such as --path, --connector, --scope, --project, or --dataset.
- Marks the snapshot envelope to say the warehouse baseline was carried forward, with zero warehouse bytes billed.
- Updates command docs, agent guidance, and maintain skill docs.
- Adds regression coverage for no-adapter/no-dialect execution, byte-identical warehouse carry-forward, preserved timestamps, rejected warehouse flags, and unchanged maintain check drift results.
## Fix
Routine dbt project refactors, such as moving model files or renaming the dbt project without changing warehouse objects, no longer require a full warehouse remap just to refresh project fingerprints. This keeps the baseline current for the transform/semantic layers without spending on warehouse profiling or laundering the warehouse evidence timestamp.

## Related issue
Closes #281


marcociav-exmergo (approved):
Nice utility, LGTM

PR #421, opened 2026-09-03
maintain verify: report output columns that are entirely NULL

### LINKED ISSUE #227: `maintain verify`: report output columns that are entirely NULL
**Depends on:** #224

### Problem

A column that is 100% NULL in a built relation is almost never intended. It is the
visible symptom of a join that matched nothing, a rename that missed, a `case`
whose branches never fire, or a source column that stopped arriving.

The diagnostic value is unusually high because the signal is unambiguous: a wholly
NULL column in a table with rows is a defect roughly every time.

### Proposal

For each selected model, report any column with a null fraction of 1.0 in a
relation that has rows. Where the column comes from a joined relation, name the
join, since a fully NULL join-side column and a failed join are the same event.

Report a near-miss band separately and more quietly (a column above some high
threshold but not at 1.0), since that is frequently legitimate.

### Why it is worth its own issue

It is the cheapest finding in the whole sweep, it needs no SQL analysis beyond
knowing which relation a column came from, and on its own it catches a failure
class that otherwise requires the caller to notice something in output they were
not looking at.

### Acceptance

- A model whose join matched nothing reports every column sourced from the
  join side, and names the join.
- A column that is legitimately optional and mostly NULL is not reported at 1.0
  unless it truly is 1.0.
- An empty relation reports "no rows" once rather than reporting every column.
- The null-fraction scan is priced with the rest of the gated half.

spascarella:
## What changes
- Adds grain drift detection for empty, mostly NULL, and fully NULL columns.
- Reports fully NULL joined columns as high severity (95%) and includes the related join metadata when available.
- Includes NULL-fraction scans in billed connector cost estimates and confirmed spend accounting.
- Updates maintain/grain and billed-handshake tests for the new scan behavior.
## Fix
- Surfaces failed joins that previously appeared only as all-NULL projected columns.
- Prevents empty relations from silently passing grain checks.
- Keeps cost preflight accurate now that grain checks include NULL-fraction aggregate scans.
## Related issue
Closes #227

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,58 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   `AGENTS.md`, `references/command-contract.md`, and both the `transform` and
   `maintain` skills document the new payload, the third class, and the offer.
 
+- **A column that is wholly or almost wholly NULL is now a grain finding, and a
+  fully NULL column names the join that could explain it** ([#227]). A column
+  that is 100% NULL in a relation that has rows is the visible symptom of a join
+  that matched nothing, a rename that missed, a `CASE` whose branches never
+  fire, or a source column that stopped arriving. Nothing in the sweep was
+  looking at it, so the caller had to notice it in output they were not reading.
+
+  `maintain grain`, and `maintain check` which sweeps every axis, now survey the
+  null fraction of every column of every in-scope relation and emit three new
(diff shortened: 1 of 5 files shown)


marcociav-exmergo (approved):
Approved thanks @spascarella, just some merge conflicts on the CHANGELOG.md

PR #423, opened 2026-09-03
SCRUM-1235-relationship conflicts and hosted+local composition

### LINKED ISSUE #408: Support Ossie in explore semantic and use-project mapping
Route `explore semantic list` through the Ossie backend and fold catalog links
into map, profile, relationships, and diagram flows, including composite declared
edges.

## Acceptance criteria

source relations are annotated and stale annotations clear; explicit
relationship edges are declared edges carrying full composite pairs; dataset
key declarations reach the grain channel; PII linkage stays conservative; the
Ossie backend passes the shipped semantic-backend conformance contract.

## Detail

Fold the catalog into `explore map/profile/relationships/diagram --use-project`.
Relations receive exposure annotations that clear on reread. Ossie relationships
are solid declared edges with full composite pairs and provenance. Profiling may
compare declared keys with evidence; verification may report a contradictory
finding but never mutates or lowers an authored declaration. PII gets only exact
relation plus direct-column links.

**Composite edges need a shape change, and it is a prerequisite rather than a
detail.** Ossie relationships carry column arrays on both sides. The cache
relationship type is already list-shaped and already uses the same direction
convention, but the neutral join shape in between carries single column strings,
so a composite relationship cannot pass through it. Widen the neutral shape to
column lists rather than adding a second shape beside it: it has one consumer,
the type it feeds is already list-shaped, and a parallel shape means two
edge-resolution paths where the point of a neutral view is one. The dbt path and
its tests move with it.

Relationship columns need not be declared fields. Upstream requires only
non-empty arrays, and a valid document routinely names a physical foreign key
column that no field entry declares. Endpoint validation runs against cached
warehouse columns, never against the declared field list.

Dataset key declarations reach the grain channel through the format's tier-1
`definitions()`, not through the semantic catalog. Without that route an
Ossie-only repository has key declarations that influence nothing.

Subclass the shipped semantic-backend conformance contract. It asserts
provenance, read idempotency, that the declared dimension scope is honest, and
that the physical column map never reaches the payload. It is close to free and
it is the standard the other backends are already held to.

Subclass `DeclaringProjectContract` here as well. It asserts that a declared
unique key and a declared join actually arrive through `definitions()`, which is
the route this issue opens, and it is opt-in precisely so a format claims it only
once it has declarations to show.

## Constraints

  already has a tier.
  dependencies.

## Dependencies

Depends on #407.

emnarahmouni-exmergo:
Closes #408 (relationship composition for the native Ossie
semantic layer). Three pieces were still open after the earlier
Ossie-setup work landed:

- **Conflict detection**: two declarations that join the same two
  datasets on different columns used to fold into two silent, unlabeled
  edges. `explore map` and `explore relationships` now surface every
  such disagreement as a `DeclaredRelationshipConflict` in a new
  `conflicts` field, naming each declaration's columns and source, with
  a note pointing at it. Every declaration is still kept as its own
  edge; dex reports the disagreement rather than picking a winner.

- **Hosted + local composition**: passing both `--use-project` and
  `--use-hosted-semantic-layer` used to read only the local project,
  the hosted branch never ran. Both are now read and unioned: semantic
  models, metrics, dimensions and measures merge by name (local wins a
  name both declare), and an entity declared on both sides merges its
  per-model roles and re-derives its summary `type` from the wider set.
  Relationship-edge extraction composes the same way. Each side is read
  independently, so a vendor missing one deployment (Ossie has no
  hosted counterpart) degrades to whichever side answered instead of
  losing an already-successful read.

- **Grain channel + Mermaid fix**: `SemanticLayer` gains a
  `declared_keys()` capability so a layer that is never a transformation
  project (Ossie) still reaches `explore profile --use-project`'s grain
  detection, added vendor-agnostically per
  `test_no_command_carries_a_vendor_branch`. Separately, the Mermaid
  edge label was dropping the parent side of a composite join
  (`(product_id, variant_id)` rendered as one unpaired list); it now
  pairs every column.

## Testing

- New unit tests for the conflict detector and for
  `_compose_semantic_catalogs` (name-union, entity role merge, local
  precedence).
- New `tests/ossie/test_duckdb_e2e.py`: a real DuckDB warehouse plus a
  native Ossie document exercised through the actual CLI
  (`relationships`, `map`, `diagram`, `--verify`) covering a
  single-column pair, a composite pair, a same-endpoint conflict, and
  an opaque query-backed dataset.
- Full `tests/explore/` + `tests/ossie/` suite passes (977 passed, 30
  skipped); the 5 remaining failures are pre-existing Windows-only
  environment issues (encoding, symlink privilege, path handling,
  known upstream schema drift) unrelated to this change.
- Manually verified end-to-end against a live DuckDB warehouse: conflict
  flagged in `explore relationships`, composite edge and conflict
  persisted through `explore map`, composite label rendered correctly
  in `explore diagram`, and both declared joins confirmed with
  `orphan_fraction: 0.0` under `--verify`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,57 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   context have a declared home instead of being dropped or smuggled into a
   neighbouring field.
 
+- **A semantic layer's own declared dataset keys now reach grain detection**
+  ([#408]). Ossie is never the transformation project `engine.project_format()`
+  resolves, so its `primary_key`/`unique_keys` declarations had no route to
+  `explore profile --use-project`'s grain channel at all: only a dbt project's
+  own tier-1 `definitions()` fed it. `SemanticLayer` gains a `declared_keys()`
+  capability, the same shape `declared_relationships()` already established;
+  every backend but Ossie's returns nothing, because a dbt-backed layer's keys
+  already reach grain through the project route and stating them twice would
+  only ever double-count. `explore commands.py` calls this capability
+  unconditionally rather than branching on which vendor is configured, honoring
+  the same architectural rule `test_no_command_carries_a_vendor_branch` already
+  enforces elsewhere. A repository with both dbt and Ossie gets both sets of
+  keys, additively, not a choice between them.
+
+- **Declared relationships that disagree about which columns join the same two
+  datasets are now flagged rather than silently doubled** ([#408]). Two
+  declarations naming the same dataset pair with different column pairs used
+  to fold into two separate, unlabeled edges with nothing to say they
+  contradict each other. `explore map` and `explore relationships` now surface
+  every such pair as a `DeclaredRelationshipConflict` in the new `conflicts`
+  field, naming every disagreeing declaration and its columns and source, with
+  a note pointing at the structured field. Every declaration is still kept as
+  its own edge: dex reports the disagreement rather than picking a winner.
+
+- **`--use-project` and `--use-hosted-semantic-layer` now compose instead of
+  one silently winning** ([#408]). Passing both used to read only the local
+  project's semantic layer; the hosted read, and its "cannot add map exposure
+  annotations" note, never ran. Both are now read and unioned: semantic
+  models, metrics, dimensions and measures merge by name (the local entry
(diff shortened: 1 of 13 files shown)


marcociav-exmergo (approved):
LGTM

PR #425, opened 2026-09-03
SCRUM-1236-Reach tier 2 for native Ossie: maintain snapshot and drift

### LINKED ISSUE #409: Add the Ossie snapshot channel and free semantic drift checks
Reach tier 2 for `OssieProject` and persist Ossie definition fingerprints in the
existing snapshot shape, extended additively for what it cannot hold today.

## Acceptance criteria

identify additions, removals, edits, dangling relations, missing columns,
broken relationships, and affected metrics; composite relationships and dataset
keys are recorded; existing snapshots load unchanged; an unreadable document is
never reported as an empty layer.

## Detail

Reach tier 2 and persist Ossie fingerprints in the **existing** snapshot shape.
The semantic-layer, semantic-model, and metric definition shapes are already
documented as format-neutral, with the definition path explicitly optional for
formats whose definitions are objects rather than files, and they already record
the physical column behind each field with computed expressions mapped to
nothing. Datasets and metrics fit as they are.

Two things the existing shape cannot hold: composite relationship tuples and
dataset key declarations. Add them as optional fields with defaults and **do not
bump the snapshot schema version**: the supported-version set holds exactly one
value, so a bump hard-refuses every committed dbt baseline for a change that
concerns only Ossie.

The real hazard is semantic rather than structural. A baseline written before
these fields carries them empty, and
empty is indistinguishable from "this layer declares no relationships", so
relationship drift would report clean. Record how the semantic side was captured,
the way `warehouse_from` records the warehouse side, and have the comparison say
the relationship axis could not be checked rather than reporting no drift.

**Do not add a parallel vendor section.** It would double comparison logic for
data the existing shape mostly already fits.

The model-DAG channel for a format with no model DAG returns document paths and
content hashes with a note saying the layer is narrower than a dbt one. That
return type is documented as one a non-dbt format fills in and its collections
default empty, so this is the shape used as intended.

Report definition added/removed/changed, dangling sources, absent direct fields,
invalid relationship endpoints and keys, and affected metrics where that impact
is known. Every finding names the document, the definition, the physical
evidence, and the affected metrics. Never interpret an unreadable document as an
empty layer: an empty layer compared against an empty layer reads as "no drift"
rather than "this could not be checked".

## Constraints

  already has a tier.
  dependencies.

## Dependencies

Depends on #407.

emnarahmouni-exmergo:
Completes issue #409: native Ossie documents reach `MaintainProject` tier 2,
so `maintain snapshot` and `maintain check` get a real drift baseline for a
semantic vendor that is not dbt.

- **`OssieSemanticLayer` implements `transform_layer()` and
  `semantic_layer()`** — the same two methods `DbtProject` already has.
  `transform_layer()` is file hashes only (Ossie declares no build step);
  `semantic_layer()` fingerprints named definitions, each with a content hash
  and the physical column behind it. Neither raises, matching tier 1's
  degrade-gracefully contract rather than dbt's "raise if no project" tier-2
  behavior, since every Ossie failure mode is an ordinary user-facing state
  (bad file, missing extra), not a wiring mistake.

- **The snapshot shape gains two things it couldn't hold before, both
  additive, neither bumping `SNAPSHOT_SCHEMA_VERSION`**: `SemanticModelDef`
  gets `relation` (for a format with no build step to name a `model_ref`
  through) and `keys` (every declared unique key, one column list per
  declaration regardless of arity); a new `SemanticLayerSnapshot.relationships`
  list carries full composite ordered column pairs.

- **The hazard the issue named directly, guarded explicitly**: an old
  baseline's empty `relationships`/`keys` is indistinguishable from "this
  layer declares none," which would report a false clean bill instead of "not
  checked." `SemanticLayerSnapshot.relationships_and_keys_captured` (default
  `False`) gates the two new finding classes — `broken_relationship` and the
  key-column analogue of `dangling_reference` — and gates relationship
  added/removed/changed detection on *both* sides of a comparison, so a
  pre-#409 baseline never reads every current relationship as freshly added.
  dbt's own snapshots leave the flag `False`: it has no composite-relationship
  or multi-column-key concept at the semantic-model level, and that's an
  honest narrower answer, not a regression.

- **`maintain`'s two layers are now read independently** (`_read_layers` in
  `maintain/commands.py`), through the same seam #408 added on the explore
  side: the semantic half comes from whichever format answers
  `semantic.vendor` (a `SEMANTIC_PROJECT_FORMATS` table lookup, not a name
  check), rather than always from `project.format`. A repository with no dbt
  project at all and `semantic.vendor: ossie` gets a semantic baseline even
  though the transform half has nothing to answer with; a repository that
  keeps dbt for its models with Ossie's semantics beside it gets both, each
  degrading on its own.

## Testing

- Extended `TestOssieProject` with `MaintainProjectContract` and
  `SemanticProjectContract`, plus the required fixture hooks — verifies tier
  2, the empty-project shape, the transform/definitions model agreement, and
  a full snapshot JSON round trip.
- 9 new unit tests (`tests/maintain/test_semantic_relationships_and_keys.py`)
  driving `semantic_free_drift` directly: the `relation`-based column-match
  fallback, missing-key-column detection (and its silence when not
  captured), `broken_relationship` for a vanished model and for a broken
  column pair, and the added/removed/changed guard against an uncaptured
  baseline.
- 4 end-to-end tests (`tests/ossie/test_maintain_e2e.py`) against a real
  DuckDB warehouse with **no dbt project at all** — the hardest case the
  routing fix exists for: `maintain snapshot` succeeds and persists Ossie's
  keys/relationships, `maintain check` flags a relationship broken by a
  dropped column, and schema/volume findings still run independently of the
  missing transform project.
- Full regression sweep: `tests/ossie/`, `tests/maintain/`, `tests/explore/`,
  `tests/adapters/` — 1836 passed, 36 skipped. The 5 remaining failures are
  the same pre-existing Windows-only environment issues already documented
  (encoding, symlink privilege, path handling, upstream schema drift),
  unrelated to this change.
- Ruff check/format clean on every touched file, no em-dashes, and a direct
  grep confirms no new violation of `test_no_command_carries_a_vendor_branch`
  (the `SEMANTIC_PROJECT_FORMATS` lookup is a table membership check, not a
  vendor-name comparison).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,49 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   context have a declared home instead of being dropped or smuggled into a
   neighbouring field.
 
+- **Native Ossie documents reach maintain's tier 2, so `maintain snapshot` and
+  `maintain check` get a real drift baseline for a semantic vendor that is not
+  dbt** ([#409]). `OssieSemanticLayer` now answers `transform_layer()` (file
+  hashes only; Ossie declares no build step) and `semantic_layer()` (named
+  definitions, each with a content hash and the physical column behind it),
+  the same two methods `DbtProject` already implements.
+
+  The snapshot shape gained two things it could not hold before, both additive
+  and both defaulted so a committed `.dex/snapshot.json` from before this
+  change still loads: `SemanticModelDef.relation` (the semantic model's own
+  physical relation, for a format with no build step to name a `model_ref`
+  through) and `.keys` (every declared unique key, one list of columns per
+  declaration regardless of arity), plus a new `SemanticLayerSnapshot.
+  relationships` list carrying full composite ordered column pairs. None of
+  it bumps `SNAPSHOT_SCHEMA_VERSION`.
+
+  The hazard the issue named directly: an old baseline's empty `relationships`
+  and `keys` is indistinguishable from "this layer declares none", which would
+  make relationship and key drift report a false clean bill instead of "not
+  checked". `SemanticLayerSnapshot.relationships_and_keys_captured` (default
+  `False`) is the guard, the same role `warehouse_from` already plays for the
+  warehouse side: `semantic_free_drift`'s two new finding classes,
+  `broken_relationship` (a relationship whose model or column pair no longer
+  resolves) and the key-column analogue of `dangling_reference`, run only when
+  the *current* read set it; relationship added/removed/changed detection
+  (folded into the existing generic diff) runs only when *both* the baseline
+  and the current read did, so a baseline that never captured relationships
+  never reads every current one as freshly added. dbt's own snapshots leave
+  the flag `False`: it has no composite-relationship or multi-column-key
+  concept at the semantic-model level to capture, and that is an honest
+  narrower answer, not a regression.
+
+  The two layers are read independently now (`maintain/commands.py`'s
+  `_read_layers`), through the same seam #408 added on the explore side: the
+  semantic half comes from whichever format answers `semantic.vendor`
+  (`SEMANTIC_PROJECT_FORMATS`, a table lookup rather than a name check), not
+  always from the configured `project.format`. A repository with no dbt
+  project at all and `semantic.vendor: ossie` gets a semantic baseline even
+  though the transform half has nothing to answer with, and a repository that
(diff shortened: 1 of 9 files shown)


marcociav-exmergo (approved):
LGTM

PR #426, opened 2026-09-04
SCRUM-1237-Verify Ossie keys through governed grain checks

### LINKED ISSUE #410: Verify Ossie keys and relationships through governed warehouse checks
Reuse the existing cost-safe verification for Ossie primary/unique keys and
explicit relationships.

## Acceptance criteria

composite keys are checked as composites; billed connectors require existing
confirmation/budget/ledger behavior; the result distinguishes declaration,
cached evidence, and new measurement; failed declarations are findings, never
auto-edits.

## Detail

Reuse the current governed warehouse machinery to verify primary and unique keys
and relationship containment and fanout. Composite keys must be measured as
composites, and the dataset unique-key declaration is an array of arrays, which
is a shape no existing declaration path carries; check what the composite grain
machinery can express before promising it. Metered connectors must use the
confirmation, budget, and ledger behavior already required. The result
distinguishes declaration, cached evidence, and new measurement. A failure is a
finding, never an automatic rewrite.

## Constraints

  already has a tier.
  dependencies.

## Dependencies

Depends on #408 and #409.

emnarahmouni-exmergo:
Completes issue #410: Ossie's declared primary/unique keys and explicit
relationships are verified through the existing governed warehouse
machinery, reused rather than rebuilt.

- **Research first, since the issue explicitly says to reuse what exists.**
  Relationship verification was already fully working for Ossie: #408
  routed its composite `DeclaredRelationship` through the neutral
  `Relationship`/`probe_candidates`/`verify_relationships` path, which
  already handles N-ary composite joins, the cost guard, and the
  confirmation/budget/ledger handshake identically to any other format —
  already proven by `test_duckdb_e2e.py`. The `unique_keys: array of arrays`
  shape the issue warns to check was also already handled correctly:
  `ossie/catalog.py`'s `_keys()` already decomposes it into one
  `DeclaredKey`/`DeclaredCompositeKey` per declaration, and `drift.py`'s
  `_declared_composites` already accumulates multiple composites per
  identifier.

- **The one real gap: routing.** `grain_plan` — the one code path in the
  whole tree that actually probes a *declared* composite key against the
  live warehouse (`declared_grain_not_unique`) — read declared keys
  exclusively from `engine.project_format().definitions()`, which is never
  Ossie. So no matter what a repository's Ossie document declared, it never
  reached this check.

- **The fix**: a new `_composed_definitions(engine)` in `maintain/commands.py`
  folds a differing semantic vendor's own declared keys into the transform
  project's definitions, additively — the identical pattern #408's
  `_fold_semantic_layer_keys` and #409's `_semantic_layer` already
  established for their own consumers. Wired into `grain_drift()`, `check()`,
  and `reconcile()` (the last one for correct advisory text, even though
  Ossie has no write tier to act on regardless).

- **Nothing else needed changing.** The declaration/cached-evidence/new-
  measurement distinction already exists on both axes (`RelationshipKind`
  +`verified` for relationships, the `key_lost_uniqueness`/
  `declared_grain_not_unique` code split for keys). The advisory-only
  behavior for a failed declaration was already an existing, held invariant
  (`reconcile.py`'s advisory table), and Ossie has no write tier at all to
  auto-edit into even if something tried.

## Testing

- 4 new end-to-end tests (`tests/ossie/test_maintain_grain_e2e.py`) against
  a real DuckDB warehouse, no dbt project at all:
  - a declared composite key that is actually broken fires
    `declared_grain_not_unique` via both `maintain grain` and the aggregate
    `maintain check` sweep;
  - a declared composite key that is genuinely unique stays silent (and free
    on DuckDB, no confirmation needed);
  - `maintain reconcile` treats the failure as `kind: advisory` with no
    `plan_id`, never an edit.
- Full regression sweep: `tests/ossie/`, `tests/maintain/`, `tests/explore/`
  — 1207 passed, 30 skipped. The 5 remaining failures are the same
  pre-existing Windows-only environment issues already documented
  (encoding, symlink privilege, path handling, upstream schema drift),
  unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,28 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   each degrading on its own rather than one absence hiding the other's
   presence.
 
+- **`maintain grain`/`maintain check` verify Ossie's declared primary and
+  unique keys, composite ones included** ([#410]). `grain_plan` read declared
+  composite keys exclusively from `engine.project_format().definitions()`,
+  which is never Ossie, so its `primary_key`/`unique_keys` declarations never
+  reached the one code path that actually probes a declared grain against the
+  live warehouse (`declared_grain_not_unique`), no matter how the connector
+  was configured. A new `_composed_definitions` folds in a differing semantic
+  vendor's own keys additively, the same seam #408's `_fold_semantic_layer_keys`
+  and #409's `_semantic_layer` already use, so a repository with no dbt
+  project at all and `semantic.vendor: ossie` now gets its declared grain
+  checked as the never-measured declaration it is (`declared_grain_not_unique`,
+  never the demoted "was unique, no longer is" `key_lost_uniqueness` a
+  measurement it never earned would read as), and priced and
+  confirmation-gated through the identical billed handshake every other
+  connector already goes through. Relationship
+  verification needed no change: #408 already routed Ossie's composite
+  declarations through the same neutral `Relationship`/`--verify` path any
+  other format's do. Nothing here proposes an edit; a failed declaration
+  surfaces only as a finding, the existing behavior for every format and,
+  for Ossie specifically, also true because it has no write tier to edit at
+  all.
+
 - **A semantic layer's own declared dataset keys now reach grain detection**
   ([#408]). Ossie is never the transformation project `engine.project_format()`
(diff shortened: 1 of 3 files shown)


marcociav-exmergo (approved):
LGTM

Issue #411, opened 2026-09-01
Add native Ossie authoring plans


marcociav-exmergo:
Reach tier 3 and the placement protocol, and implement `semantic ossie
define|update|plan` using a format-neutral document edit kind and the existing
plan/apply stale-edit protection.

## Acceptance criteria

define/update namespace guards work; the implementation explicitly chooses
either a tested surgical editor that preserves comments and formatting or
semantic-content preservation with documented reformatting; only configured
native documents are within the editing surface; dbt authoring is unchanged;
the format passes the shipped editable-tier and placement contracts.

## Detail

Depends on #409 structurally, not by preference: the editable tier extends the
maintain tier, so the snapshot channel has to exist first.

Reach tier 3 and the placement protocol, and add `semantic ossie
define|update|plan` with a format-neutral document edit kind. Reuse the existing
edits payload, stored plan, diff, source-hash pinning, stale-edit refusal, and
atomic apply. Define rejects existing names; update rejects missing names; plan
accepts mixed atomic edits.

**Be honest about how much of the dbt path is reusable.** The dbt semantic plan
path loads a dbt project, runs dbt-specific reference and mode checks, warns
about the time spine, and gates on a shadow project parse. What is genuinely
shared is plan construction and the apply spine. This is a sibling validator over
that spine, not a reuse of the dbt planner, and the dbt path's external parse gate
has no Ossie equivalent.

The editable surface is only configured, repository-confined Ossie documents,
declared through the placement protocol so confinement is structural. Choose and
test one preservation contract: a surgical textual editor that retains comments,
anchors, ordering, and formatting, or semantic-content preservation with a clear
warning that formatting may change. YAML parse/re-serialize alone cannot promise
byte preservation.

Add the safety-spine entries here: authoring writes nothing outside the
configured documents, and no plan is stored when validation fails. The spine is
the constitutional test and it currently has no Ossie entry.

Non-goals: dbt-to-Ossie generation or synchronization, direct writes, and dbt
parser validation.

## Constraints

Two constraints govern this work wherever it touches the engine, and an issue
that satisfies its own tests while breaking either is implemented wrong:

- **Ossie is a project format, not a set of vendor branches.** The project
  adapter is tiered and structurally checked, and every channel this work needs
  already has a tier.
- **Ossie does not depend on MetricFlow and dbt does not depend on Ossie.**
  Neither direction may import the other's reader, schema, or optional
  dependencies.

## Dependencies

[...]

PR #428, opened 2026-09-04
SCRUM-1239-Validate Ossie authoring against cached physical evidence

### LINKED ISSUE #412: Add Ossie authoring validation against cached physical context
Validate source relations, direct fields, keys, and relationship endpoints using
the current cache without opening a warehouse.

## Acceptance criteria

unknown evidence is a note; invalid known references are refused; computed
expressions and query-backed sources are not over-validated; no plan is stored
on failure; no connection is opened.

## Detail

At plan time validate source relations, direct field columns, key declarations,
and relationship endpoint and key references against cached exploration evidence.
Known-invalid references refuse storage. Missing or unprofiled evidence is a
named note. Computed expressions, quoted identifiers, and query-backed sources
remain unverified rather than over-validated. This command must not open a
connection or scan a warehouse.

## Constraints

  already has a tier.
  dependencies.

## Dependencies

Depends on #408 and #411.

emnarahmouni-exmergo:
Closes : #412: `semantic ossie define|update|plan` validates the
prospective document set against cached exploration evidence (`.dex/cache.json`)
before storing a plan, without opening a warehouse connection.

- **New `ossie/cache_validation.py`** checks every dataset's source relation,
  direct (bare-identifier) field columns, declared primary/unique keys, and
  relationship endpoint columns against the cache. It deliberately reuses the
  exact same `normalize_relation`/`match_identifier`/`column_reference`/
  `select_expression` primitives the read catalog already uses, so a
  reference this rejects at plan time is checked by the identical rule
  `explore semantic list` already applies when reading.

- **The known-invalid vs. unknown split is structural, not heuristic.** A
  relation is refused only when it's absent from a namespace the cache
  *completely inventoried*; a column is refused only when it's missing from a
  relation the cache actually *profiled*. A new `CacheProvenance.
  inventory_namespaces` field carries that provenance, populated by reusing
  the `observed_namespaces` bookkeeping issue #149 already established for
  carry-forward detection, not a new concept. Everything else, an unprofiled
  relation, a namespace never inventoried, a query-backed or quoted source, a
  computed field expression, is a note. No plan is stored on a refusal.

- **Wiring is minimal**: `transform/commands.py`'s `semantic_ossie()` reads
  the cache via `readable_cache(engine.store)` (a store read, never
  `connect.open_adapter`) and passes it into `validate_plan()`, which raises
  before any plan is stored when the cache contradicts a reference. Notes
  flow through the existing `Result.notes` mechanism with no new plumbing.

## Testing

- 16 tests in `tests/ossie/test_authoring.py` (8 pre-existing from #411, 8 new
  for this issue): absent-cache note, unprofiled-columns note, a relation
  confirmed missing from a completely-inventoried namespace refuses the plan,
  a column confirmed missing from a profiled relation refuses the plan
  (parametrized across field/key/relationship-endpoint), computed and
  opaque-source references are left unvalidated, and a hard guarantee test
  that monkeypatches `connect.open_adapter` to raise if cache validation ever
  tries to open a connection.
- Full `tests/ossie/` + `tests/explore/` sweep: 1013 passed, only the 5
  pre-existing Windows-only environment failures already documented on prior
  PRs in this series (encoding, symlink privilege, path handling, upstream
  schema drift), unrelated to this change.

## Notes

Stacked on `SCRUM-1238-issue-411` per #412's own dependency on #411 (the
authoring/plan machinery this validates didn't exist before that branch).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,27 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Added
 
+- **Native Ossie authoring plans are validated against cached exploration
+  evidence before they are stored, with no warehouse connection opened**
+  ([#412]). `semantic ossie define|update|plan` now checks each dataset's
+  source relation, direct field columns, declared keys, and relationship
+  endpoint columns against `.dex/cache.json`, reusing the exact same
+  `normalize_relation`/`match_identifier`/`column_reference` primitives the
+  read catalog already uses, so a reference dex would call invalid while
+  reading is checked by the identical rule at plan time.
+
+  The distinction the issue asks for is structural, not a judgment call: a
+  relation is refused only when it is absent from a namespace the cache
+  completely inventoried (a new `CacheProvenance.inventory_namespaces`,
+  populated from the same observed-namespace bookkeeping issue #149 already
+  established for carry-forward, not a new concept); a column is refused only
+  when it is missing from a relation the cache actually profiled. Everything
+  else, an unprofiled relation, a namespace never inventoried, a query-backed
+  or quoted source, a computed field expression, is a note, never a refusal:
+  absence of evidence is not evidence of absence. A refused reference stores
+  no plan, exactly like every other validation failure this command already
+  has.
+
 - **Native Ossie semantic documents can now be authored as reviewable plans**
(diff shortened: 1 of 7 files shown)


marcociav-exmergo (approved):
LGTM

PR #431, opened 2026-09-05
BigQuery measures a TIMESTAMP column's month gap in dates, and the table keeps its aggregates

### LINKED ISSUE #430: BigQuery: temporal continuity renders `TIMESTAMP_DIFF(…, MONTH)`, which BigQuery rejects, and the whole table degrades to metadata-only
### What happened

On BigQuery, `explore map` requests temporal continuity for every non-PII `TIMESTAMP` column at the `month` grain, and the adapter renders it as `TIMESTAMP_DIFF(later, earlier, MONTH)`. BigQuery's `TIMESTAMP_DIFF` supports `MICROSECOND` through `DAY` only, so the statement is rejected at job insert with a 400. Because the continuity subqueries are spliced into the same flat `SELECT` as the batch's other aggregates, the `BadRequest` fallback in `column_aggregates` then degrades the **entire table** to metadata-only — every column loses null fraction, distinct count, min/max, key detection and grain, not just the temporal fractions — and records `aggregate profiling failed and was skipped` in `data_quality`.

Expected: either a valid month-grain expression for `TIMESTAMP` (`DATETIME_DIFF` and `DATE_DIFF` both accept `MONTH`, so `DATETIME` and `DATE` columns are unaffected), or a fallback that drops the optional temporal fractions and keeps the core aggregates when the batch fails.

Impact seen: our billing-export table (28 columns, three `TIMESTAMP` columns the classifier leaves at `None`) has had no aggregate profile on any nightly `explore map` since 1.6.6 arrived on 2026-08-16. The BigQuery Jobs API shows its aggregate statement running as a job every night before that and never after — a 400 at insert creates no job, so nothing downstream of the note changes: the run still returns `ok`. Verified the generator is identical in 1.9.1 and 1.9.2 (`temporal_units_for` and `_date_diff_expr` diffed), so this is present from 1.6.6 (#206) onward.

### Steps to reproduce

Without a client, on the installed wheel:

```python
from exmergo_dex_core.adapters import bigquery as bq
from exmergo_dex_core.adapters.base import ColumnMeta

col = ColumnMeta(name="t", data_type="TIMESTAMP", nullable=True, ordinal=1)
adapter = bq.BigQueryAdapter.__new__(bq.BigQueryAdapter)
sql, _ = bq.BigQueryAdapter._build_aggregate_sql(
    adapter, "p.d.t", [col], set(), set(), set(), set(), {"t"}
)
assert "TIMESTAMP_DIFF(" in sql and ", MONTH)" in sql  # passes on 1.9.1 and 1.9.2
```

Live: any BigQuery table with one `TIMESTAMP` column whose preliminary PII flag is `None`, then `explore map`. The dataset's `data_quality` carries the note below and every column's aggregates are `None`.

Suggested fixes, any of which would do: render the month grain for `TIMESTAMP` as `DATE_DIFF(DATE(later), DATE(earlier), MONTH)`; or drop `month` from the `TIMESTAMP` unit set on this adapter; and, separately, retry the batch without the temporal expressions before degrading the whole table to metadata-only, since one optional fraction taking every column's core aggregates with it is the part that hurt. Related: #206 (the feature) and #315 — the same adapter path has no live `TIMESTAMP` coverage either, which is how this shipped green.

### Connector

BigQuery

### dex-core version

1.9.2 (also reproduced against the 1.9.1 wheel; first observed live on 1.6.6)

### Environment

Linux container (python:3.12), invoked through the CLI (`explore map --use-project`) by a FastAPI service on a nightly schedule.

### Relevant output

```shell
aggregate profiling failed and was skipped: POST https://bigquery.googleapis.com/bigquery/v2/projects/<project>/jobs?prettyPrint=false: TIMESTAMP_DIFF does not support the MONTH date part when the argument is TIMESTAMP type at [1:6355]

Location: US
Job ID: <redacted>
```

catincloudlabs:
# BigQuery measures a TIMESTAMP column's month gap in dates, and the table keeps its aggregates

Closes #430.

## The defect

On BigQuery, `explore map` requests temporal continuity for every non-PII `TIMESTAMP` column at
the `month` grain, and the adapter rendered it as `TIMESTAMP_DIFF(period, prev_period, MONTH)`.
BigQuery's `TIMESTAMP_DIFF` supports `MICROSECOND` through `DAY` only, so the statement was refused
at job insert with a 400:

```
TIMESTAMP_DIFF does not support the MONTH date part when the argument is TIMESTAMP type
```

The continuity subqueries are spliced into the same flat `SELECT` as the batch's other aggregates,
so the `BadRequest` fallback in `column_aggregates` degraded the **whole batch** to metadata-only.
Every column lost its null fraction, distinct count, min and max, key detection and grain, not only
the temporal fractions, and `data_quality` recorded `aggregate profiling failed and was skipped`.
The run still returned `ok`. A 400 at insert creates no job, so nothing downstream of that note
changes.

Present since 1.6.6 shipped the feature (#206); `temporal_units_for` and `_date_diff_expr` are
identical in 1.9.1 and 1.9.2. `DATETIME` and `DATE` columns were never affected because
`DATETIME_DIFF` and `DATE_DIFF` both accept `MONTH`.

## The fix

`_date_diff_expr` renders the month gap on a `TIMESTAMP` column as
`DATE_DIFF(DATE(period), DATE(prev_period), MONTH)`. Both operands are already
`TIMESTAMP_TRUNC(..., MONTH)` periods, so their UTC dates are month boundaries and the diff is exact.
Day and hour keep `TIMESTAMP_DIFF`; `DATETIME` and `DATE` render byte-identically to before.

## Verification

- New test builds the aggregate statement for one `TIMESTAMP`, one `DATETIME` and one `DATE` column
  and pins the diff family per unit. The `TIMESTAMP` arm fails on `main` (it is the reproducer from
  the issue) and passes here; the other two arms pin that nothing else moved. The existing BigQuery
  temporal test still passes, since day and hour still use `TIMESTAMP_DIFF`.
- The assembled statement still passes `assert_select_only` in the bigquery dialect for all three
  types.
- `uv run pytest` in `packages/dex-core`: `2872 passed, 202 skipped, 23 failed`. The 23 are the same 23 that fail on an untouched `main` checkout on this Windows machine (demo path-separator assertions and `transform/test_dev_target.py`), so none is from this change. The new test fails on `main` with the fix stashed and passes with it.
- `uvx ruff@0.15.19 check .` and `format --check .` clean; the em-dash gate clean.
- `uvx pytest evals`: 6 wrapper tests fail identically on an untouched `main` checkout on this
  Windows machine, so they are unrelated to this change.
- Live, since this path has no `TIMESTAMP` coverage upstream: a BigQuery dry run of `TIMESTAMP_DIFF(period, prev_period, MONTH)` over `TIMESTAMP_TRUNC(..., MONTH)` periods is refused with the issue's exact message, and the same dry run of `DATE_DIFF(DATE(period), DATE(prev_period), MONTH)` validates. On literals, `2025-11-01` to `2026-03-15` at MONTH gives `4`, and the day arm still gives `3` for `02-27` to `03-02`. Both reproducible from `bq query --dry_run` with no table.

## Not in this PR

Retrying the batch without the optional temporal expressions before degrading to metadata-only.
One optional fraction taking every column's core aggregates with it is the part that hurt, but
that is a change to the fallback contract, not to the expression that broke it, and it costs a
second statement and a second cost-gate pass. The issue leaves it open for a maintainer's call.

Related: #315, since the same adapter path has no live `TIMESTAMP` coverage, which is how this
shipped green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -195,6 +195,28 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ### Fixed
 
+- **BigQuery `explore map` profiles a table with a `TIMESTAMP` column again**
+  ([#430]). Temporal continuity asked for a month-grain gap on every non-PII
+  `TIMESTAMP` column as `TIMESTAMP_DIFF(period, prev_period, MONTH)`, and
+  BigQuery's `TIMESTAMP_DIFF` stops at `DAY`, so the statement was refused at
+  job insert. The continuity subqueries ride the same flat `SELECT` as the
+  batch's other aggregates, and the `BadRequest` fallback in `column_aggregates`
+  degrades the whole batch, so one optional fraction took every column's null
+  fraction, distinct count, min and max, key detection and grain with it, and
+  recorded `aggregate profiling failed and was skipped` in `data_quality`. The
+  run still returned `ok`. Present since 1.6.6 shipped the feature ([#206]);
+  `DATETIME` and `DATE` columns were never affected because their diff
+  families accept `MONTH`.
+
+  The month gap on a `TIMESTAMP` column is now
+  `DATE_DIFF(DATE(period), DATE(prev_period), MONTH)`. Both operands are
+  already `TIMESTAMP_TRUNC(..., MONTH)` periods, so their UTC dates are month
+  boundaries and the diff is exact; day and hour keep `TIMESTAMP_DIFF`, and
+  the other two temporal types render as before. Whether the batch should
+  retry without the optional temporal expressions before degrading to
+  metadata-only is left open on the issue: it is a change to the fallback
+  contract, not to the expression that broke it.
+
 - **`maintain` stops refusing an engine that was never given a repo root.**
   `check`, `grain`, and `reconcile` each read the project format without
   guarding the construction, and the shipped dbt format correctly refuses to
--- a/packages/dex-core/src/exmergo_dex_core/adapters/bigquery.py
+++ b/packages/dex-core/src/exmergo_dex_core/adapters/bigquery.py
@@ -142,6 +142,13 @@ def _date_trunc_expr(qcol: str, unit: str, data_type: str) -> str:
 
 
 def _date_diff_expr(unit: str, later: str, earlier: str, data_type: str) -> str:
+    # TIMESTAMP_DIFF stops at DAY: BigQuery refuses MONTH (and every coarser
+    # part) for TIMESTAMP arguments, at job insert, so the whole aggregate
+    # statement failed and the table degraded to metadata-only (issue #430).
+    # Both operands here are TIMESTAMP_TRUNC(..., MONTH) periods, so their UTC
+    # dates are month boundaries and DATE_DIFF at MONTH is exact.
+    if unit == "month" and "TIMESTAMP" in data_type.upper():
+        return f"DATE_DIFF(DATE({later}), DATE({earlier}), MONTH)"
     return f"{_diff_family(data_type)}({later}, {earlier}, {unit.upper()})"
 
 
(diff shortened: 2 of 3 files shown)


marcociav-exmergo (approved):
Awesome, thanks @catincloudlabs.
I'm prepping a release that includes this fix

Issue #357, opened 2026-08-26
The shared PII gate screens filter clauses with MetricFlow-specific regexes, so a non-dbt backend inherits a gate that silently screens nothing


marcociav-exmergo:
## Problem

`requested_dimension_refs` lives in the **neutral** package `__init__.py`, is
shared by both backends, and screens filter clauses with two MetricFlow-specific
regexes:

```python
_DIMENSION_REF = re.compile(r"(?:Time)?Dimension\(\s*['\"]([^'\"]+)['\"]")
_ENTITY_REF = re.compile(r"Entity\(\s*['\"]([^'\"]+)['\"]")
```

That is correct for dbt, whose `where` clauses are Jinja calls. It is a fail-open
for anything else. A semantic layer whose filters are JSON objects rather than
Jinja calls matches neither regex, so `refs` contains the `group_by` tokens only
and **every dimension referenced in a filter goes unscreened**.

The failure is invisible. The query succeeds, and the notes disclose nothing,
because nothing was requested: `unadjudicated_refs` can only report on refs the
extractor found. So a third backend inherits a gate that silently does nothing on
half its input, and the first sign of it is a disclosure that never comes.

Together with #347 this is the second independent structural fail-open
in the same gate, and both have the same shape: **the authoritative screening
quietly does not run.** That is the class #172 was filed about.

This is latent today, since both shipped backends are dbt. It is worth fixing
before a second format rather than during it, because "we will fix the gate when
the second backend lands" means shipping a backend whose gate is decorative, and
the review that would catch it is the one least likely to look at shared code.

## Fix

Move filter-reference extraction behind the `SemanticBackend` seam: each backend
knows its own filter dialect and reports the dimension and entity refs a query
would touch. The neutral layer keeps the screening policy (evidence beats names,
silence never clears, name heuristic as the fail-closed floor) and stops owning
the parsing.

A backend that cannot extract refs from its filter dialect must say so through the
capability declaration and refuse filtered queries, not pass them.

## Related vendor vocabulary in shared code

Lower stakes, same direction, worth cleaning in the same pass:

- `SemanticQuery`'s docstring promises `where` is "the Jinja filter dialect ...
  verbatim on both backends", which is a dbt contract written into the neutral
  seam
- the `metric_time` special cases in `_split_grain` and `_group_by_names`
- `_GRAINS`, covered by #355

PR #446, opened 2026-09-09
SCRUM-1292-Fix classify_content misreading MetricFlow filter grammar

### LINKED ISSUE #445: `classify` reads MetricFlow's own filter grammar as a repository-controlled macro call, so every filtered metric classifies executable
## Problem

`transform/classify.py` reports MetricFlow's own filter grammar as content the
repository controls. The same engine already knows better, one module away.

A metric whose measure carries a filter is the ordinary shape of a semantic
definition:

```yaml
metrics:
  - name: live_sessions
    label: Live sessions
    type: simple
    type_params:
      measure:
        name: session_count
        filter: "{{ Dimension('session__is_deleted') }} = false"
```

`classify_content` reads `Dimension` as a macro call and takes the executable
verdict. `metricflow_dialect.filter_refs`, in the same installed package, reads
the identical string as a dimension reference:

```
classify_content:       executable
signals:                [('macro_call', 'Dimension')]
metricflow filter_refs: ['session__is_deleted']
same metric, no filter: declarative
```

Reproduction, run against 1.12.0 to produce the output above:

```python
from exmergo_dex_core import host, metricflow_dialect

METRIC = """metrics:
  - name: live_sessions
    label: Live sessions
    type: simple
    type_params:
      measure:
        name: session_count
        filter: "{{ Dimension('session__is_deleted') }} = false"
"""

FILTER = "{{ Dimension('session__is_deleted') }} = false"

verdict = host.classify_content(METRIC, path="models/live_sessions.yml")
print("classify_content:      ", verdict.artifact_class.value)
print("signals:               ", [(s.signal, s.detail) for s in verdict.signals])
print("metricflow filter_refs:", metricflow_dialect.filter_refs([FILTER]))

plain = METRIC.replace(
    '\n        filter: "{{ Dimension(\'session__is_deleted\') }} = false"', ""
)
print(
    "same metric, no filter:",
    host.classify_content(plain, path="models/live_sessions.yml").artifact_class.value,
)
```

Versions: `exmergo-dex-core` 1.12.0, `dbt-core` 1.12.3, `dbt-duckdb` 1.11.0,
`duckdb` 1.5.5, Python 3.11.14.

## Why this is a defect rather than a preference

The classifier's own docstring gives the rationale for the signal:

> A macro call is a signal because its meaning lives in code the repository
> controls, which is the fact a host offline in a disposable checkout is
> actually asking about.

That is exactly what is not true of `Dimension`, `TimeDimension` and `Entity`.
They are MetricFlow's rendering vocabulary, resolved by the semantic layer
against definitions the project already declares. They are not dispatch into a
macro the repository wrote, and no local macro of those names is consulted. So
the verdict is reached on a premise that does not hold for these callees, and
`_DECLARATIVE_CALLEES` reads as a list that simply had not met a semantic filter
yet:

```python
_DECLARATIVE_CALLEES = frozenset(
    {"ref", "source", "var", "config", "this", "target", "doc", "env_var"}
)
```

`doc` is in that set with a reason recorded beside it: flagging the most common
jinja call in a `schema.yml` would make the signal useless. A metric filter is
the same argument one layer up, and it is the most common jinja call in a
semantic YAML.

There is already a module that owns this grammar, added when #357 moved
MetricFlow-specific regexes out of shared code:

```python
_DIMENSION_REF = re.compile(r"(?:Time)?Dimension\(\s*['\"]([^'\"]+)['\"]")
_ENTITY_REF = re.compile(r"Entity\(\s*['\"]([^'\"]+)['\"]")
```

`classify.py` imports `..dbt_project` and `..edits` and does not import it. The
classifier shipped after that move, in #442, so this looks like a module that
was written without the vocabulary rather than a decision to ignore it.

## Consequence for a host

Content classification is the gate a host applies before it writes a change
offline. Under the shipped verdict, a semantic metric that carries a filter is
never `declarative`, so a host whose policy admits declarative content and holds
everything else for a person holds **every filtered metric**, which is most of
them. A metric with no filter and an edit to a semantic model both classify
`declarative` and pass, so the effect is narrow and precise: the gate refuses the
one shape a semantic layer exists to express.

The verdict is the engine's and a host should not second-guess it, which is why
this is a request to fix the classifier rather than a note about working around
it. A host that special-cased these callees itself would be keeping a second,
private opinion about what dbt renders, which is the thing the classification
contract exists to prevent.

## Suggested shape

Let `classify.py` consult `metricflow_dialect` for the callees that module owns,
so one engine has one answer about what `Dimension('...')` means. A filter clause
that references a dimension the project declares is declarative; the signal stays
worth reporting where the callee is genuinely unknown.

Two boundaries worth keeping while doing it:

- **Silence must still not read as safe.** Content that does not parse stays
  `UNKNOWN`, and a filter that cannot be scanned stays a signal.
- **A document carrying both a filter and a hook is still executable.** The
  stronger verdict rule should be untouched; what changes is only whether a bare
  dimension reference is evidence of repository-controlled execution.

## Acceptance

- A metric whose measure carries `{{ Dimension('...') }}`, `{{ TimeDimension('...', '...') }}`
  or `{{ Entity('...') }}` classifies `declarative`, with the reference still
  reported as a signal if that is the chosen shape.
- The same metric plus a `post-hook` classifies `executable`.
- A metric filter calling a macro the project defines still classifies
  `executable`, so the narrowing is to the vendor's vocabulary and not to filters
  in general.
- Unparseable content still classifies `UNKNOWN` and never `declarative`.
- A conformance vector covers the filtered metric, since the shipped
  `classification` vectors currently do not.

## Adjacent, lower stakes, same contract

Raised here rather than separately because it is a question about intent rather
than a defect, and it is one field away.

`Grounding.completeness` stays `complete` for a plan whose edit references a
model that does not exist. `ref('no_such_model')` is reported as a dependency
with no `relation`, `unresolved` stays empty, and the summary reads `complete`.
That is consistent with the shipped `grounding-unresolved` vector, where
`unresolved` means a reference the engine read and could not name statically
(`ref(var('which'))`), and a name that resolves to nothing is a different
condition.

The per-dependency information a host needs is all present, so nothing is
missing; the question is whether `completeness` is meant to be readable on its
own. A host that gates on it alone admits a plan whose build has nothing to run,
and the field's name invites exactly that reading. If the intent is that
`completeness` describes only static nameability, a sentence in its docstring
saying so would close it and no code needs to change.

emnarahmouni-exmergo:
Closes #445: `classify_content` read MetricFlow's own filter grammar
(`Dimension`/`TimeDimension`/`Entity`) as an unrecognized macro call, so any
metric with a filter classified `executable` instead of `declarative`, the
narrowing is real and specific, not a general softening of the classifier.

- **Root cause**: `_jinja_signals`' callee loop had no branch for these three
  names, so they fell through to the generic `macro_call` signal, which is in
  `EXECUTABLE_SIGNALS`. But these names dispatch into the semantic layer's own
  resolution against definitions the project already declares, not into a
  macro the repository wrote , the same distinction `_DECLARATIVE_CALLEES`
  already draws for `ref`/`source`/`var`/etc.

- **Fix**: `metricflow_dialect.py` (already the owner of this grammar since
  #357) exports `FILTER_CALLEES = frozenset({"Dimension", "TimeDimension",
  "Entity"})`. `classify.py` consults it rather than re-deriving the names.
  A match produces a new `semantic_ref` signal , still reported, per the
  module's own "a signal is reported even when it did not decide the class"
  principle — but `semantic_ref` is not in `EXECUTABLE_SIGNALS`, so it no
  longer flips the verdict.

- **Scoped twice, through review**: the exception only applies when the call
  sits inside a `filter` key's value, *and* that key is under `metrics[...]`.
  Two false-safe boundaries this closes:
  - A repository macro that happens to be named `Dimension`, called from a
    `description` or any other field, still classifies `macro_call`/executable.
  - A `filter` key outside `metrics[...]` (`models[].config.filter`, say)
    names no MetricFlow grammar at all and still classifies
    `macro_call`/executable.

  Both were caught in review and are now regression-tested.

- **Untouched, by design**: a filter calling a macro the project actually
  defines still classifies executable (the narrowing is to the vendor's
  vocabulary, not to filters in general); a document with both a filter and a
  `post-hook` is still executable (execution still outranks everything else in
  `_verdict`); unparseable content is still `UNKNOWN`, never `declarative`.

- **Adjacent, lower-stakes item from the same issue**: `Grounding.
  completeness`'s docstring now says explicitly that "complete" means every
  reference was named statically, not that every named reference resolves to
  something that exists — a documentation clarification only, no behavior
  change, as the issue itself specified.

## Testing

- 8 new tests in `tests/transform/test_classify.py`: all three callees
  declarative, still executable beside a hook, a project-defined macro in a
  filter still executable, unparseable content still unknown, and the two
  false-safe boundary regressions from review (macro named `Dimension`
  outside a filter; `filter` key outside `metrics[...]`).
- New `classify-filtered-metric` conformance vector in
  `host/vectors/`, exercised by the existing
  `test_each_classification_vector_still_classifies_the_way_it_says`
  (its expected output was generated by running the real classifier, not
  hand-written).
- `tests/transform/test_classify.py` + `tests/test_host_contract.py`: 57
  passed. Wider `tests/transform/` + `tests/test_host_contract.py` sweep:
  747 passed, 0 failed.

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dex",
   "description": "Analytics engineering for Claude Code and any agent: data warehouse exploration, dbt transformation and semantic modeling, and schema-drift maintenance on dbt.",
-  "version": "1.12.0",
+  "version": "1.12.1",
   "author": {
     "name": "Exmergo, Inc.",
     "email": "support@exmergo.com",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,34 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`classify_content` read a metric filter's `Dimension`/`TimeDimension`/
+  `Entity` calls as an unrecognized macro, so a metric with a filter was
+  never `declarative`** ([#445]). Those three names are MetricFlow's own
+  filter grammar: they dispatch into the semantic layer's resolution against
+  definitions the project already declares, not into a macro the repository
+  wrote, so treating them as an unknown `macro_call` reached the executable
+  verdict on a premise that never held for them. `classify.py` now consults
+  `metricflow_dialect.FILTER_CALLEES`, the same module `#357` already made the
+  owner of this grammar, rather than re-deriving it. The reference is still
+  reported (a new `semantic_ref` signal) so a host applying its own policy
+  keeps the evidence; a filter calling a macro the project actually defines is
+  untouched and still classifies executable, and a document carrying both a
+  filter and a hook is still executable, since execution still outranks
+  everything else in `_verdict`. The exception is scoped to the value of a
+  `filter` key under `metrics[...]` specifically, not the callee name
+  everywhere it appears and not every `filter` key: a repository macro that
+  happens to be named `Dimension` and is called from a description or any
+  other field is still repository-controlled code and still classifies
+  `macro_call`/executable, and so is a `filter` key outside `metrics[...]`
+  (`models[].config.filter`, say), which names no MetricFlow grammar at all.
+  A new `classify-filtered-metric` conformance vector covers it. Also
+  clarified, no behavior change: `Grounding.
+  completeness` docstring now says explicitly that "complete" means every
+  reference was named statically, not that every named reference resolves to
+  something that exists.
+
 ## [1.12.0] - 2026-09-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+## [1.12.1] - 2026-09-09
+
 ### Fixed
 
 - **`classify_content` read a metric filter's `Dimension`/`TimeDimension`/
--- a/packages/dex-core/src/exmergo_dex_core/host/vectors/classify-filtered-metric.json
+++ b/packages/dex-core/src/exmergo_dex_core/host/vectors/classify-filtered-metric.json
@@ -1,0 +1,25 @@
+{
+  "name": "classify-filtered-metric",
+  "contract": "classification",
+  "description": "A metric filter naming a dimension through MetricFlow's own grammar. Dimension('...') dispatches into the semantic layer's resolution against definitions the project already declares, not into a macro the repository wrote, so the verdict is declarative and the reference is still reported as a signal.",
+  "payload": {
+    "path": "models/marts/live_sessions.yml",
+    "content": "metrics:\n  - name: live_sessions\n    label: Live sessions\n    type: simple\n    type_params:\n      measure:\n        name: session_count\n        filter: \"{{ Dimension('session__is_deleted') }} = false\"\n"
+  },
+  "expect": {
+    "artifact_class": "declarative",
+    "result": {
+      "path": "models/marts/live_sessions.yml",
+      "artifact_class": "declarative",
+      "basis": "content",
+      "parsed": true,
+      "signals": [
+        {
+          "signal": "semantic_ref",
+          "where": "metrics[0].type_params.measure.filter: line 1",
+          "detail": "Dimension(session__is_deleted)"
+        }
+      ]
+    }
+  }
+}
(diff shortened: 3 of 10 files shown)


marcociav-exmergo (approved):
LGTM, thanks @emnarahmouni-exmergo!

Preparing release 1.12.1

PR #447, opened 2026-09-09
SCRUM-1293-maintain verify: report column contract drift on built relations

### LINKED ISSUE #230: `maintain verify`: report a built relation whose columns diverge from its declared contract
**Depends on:** #224, #214

### Problem

A model with a `schema.yml` entry declaring its columns can build with a different
set, and nothing compares the two. Columns documented but absent, and columns
present but undocumented, are both invisible.

### Proposal

For each selected model with declared columns, compare the declared set to the
built relation's actual columns and report both directions. Where a type is
declared, compare that too.

Deliberately a warning rather than a refusal, and deliberately two-directional: an
undocumented column is a documentation gap, and a documented column that does not
exist is a real defect that will break anyone reading the docs to write a query.

#214 is the same comparison at plan time, before the build. Both are worth
having, since a caller who edits a model by hand never goes through `transform
plan`. They should share the comparison logic.

### Acceptance

- A model missing a declared column reports it.
- A model with an undeclared extra column reports it, at lower severity.
- A model with no `schema.yml` entry reports nothing, and the absence of a contract
  is itself reported once at summary level rather than per model.
- Free on every connector, since both sides come from metadata.

emnarahmouni-exmergo:
Closes #230

- `maintain verify` now reports a built relation whose columns diverge from
  its declared schema.yml contract: a missing declared column (`column_missing`,
  high), an undeclared extra column (`column_undeclared`, low), and a type
  mismatch where both sides declare one (`column_type_mismatch`, medium). A
  model with no `columns:` entry reports nothing per model; how many such
  models were considered is named once at the summary level (#230's
  acceptance criteria).
- Shares its comparison logic with #214's plan-time check rather than
  duplicating it: `dbt_project.column_contract_divergence` is the one
  set-math primitive both `transform plan`'s authored-SELECT-list comparison
  and this built-relation comparison call, differing only in where each
  side's names and types come from.
- Suppressed the same way build-status and no-relation findings already are
  when the project doesn't compile, the warehouse is unreachable, or there's
  no dbt project.
- `maintain verify <object>` now actually scopes the column contract to the
  object named, instead of computing it for every model and discarding the
  rest: the plan-level read is narrowed up front, and the finding's
  identifier is now the model name (matching every other finding class in
  this module) instead of the warehouse-qualified relation name, which
  previously meant a scoped request silently returned zero column findings.

## Test plan
- [x] `tests/maintain/test_verify.py` — 58 passed (10 pure unit tests for
      `column_contract_plan`/`column_contract_findings`, 4 end-to-end tests
      including the new object-scoping regression test, plus the
      pre-existing compile-failure suppression test updated for the new key)
- [x] `tests/transform/test_plan.py -k column` — #214's existing 6 tests
      still pass unchanged after refactoring it onto the shared primitive
- [x] Full regression sweep: `tests/maintain/ tests/transform/` — 997 passed
- [x] `ruff check` / `ruff format` clean on all touched files
- [x] Zero em-dashes in code, tests, and CHANGELOG

### DIFF
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "dex",
   "description": "Analytics engineering for Claude Code and any agent: data warehouse exploration, dbt transformation and semantic modeling, and schema-drift maintenance on dbt.",
-  "version": "1.12.1",
+  "version": "1.12.2",
   "author": {
     "name": "Exmergo, Inc.",
     "email": "support@exmergo.com",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,37 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **`maintain verify` reports a built relation whose columns diverge from its
+  declared schema.yml contract** ([#230]). For every selected model that
+  declares columns, the model's actual built columns (read free, from
+  `adapter.table_metadata`, the same schema-only lookup `missing_relation_
+  findings` already uses) are compared against the compiled manifest's own
+  declared set, in both directions, plus a type check where a type is
+  declared. A documented column the relation does not have is a real defect
+  (`column_missing`, high severity); an undocumented column is a
+  documentation gap (`column_undeclared`, low); a type that disagrees ranks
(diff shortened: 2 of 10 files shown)


marcociav-exmergo (approved):
LGTM!

PR #449, opened 2026-09-09
`explore map` carries the `pii_overrides` mismatch warning that `explore profile` already emits

### LINKED ISSUE #448: `explore map` should carry the `pii_overrides` mismatch warning that `explore profile` already emits
### Problem

`explore profile` warns when a `pii_overrides` entry matches no column of the profiled set — `_override_mismatches(datasets, config.pii_overrides)` feeds its result's `warnings=` (`explore/commands.py`, the only call site, inside `profile`; read at 1.11.0 and again at 1.12.1, line 767). `explore map` profiles the same datasets through `profile_mod.profile(...)` but builds its own `warnings` list — carry-forward, `--verify` and overlap warnings only — and never calls `_override_mismatches`. The cleared count (`_override_notes`) lands in `notes`, which says how many overrides applied, not which ones applied to nothing.

So a host that runs `map` on a schedule and `profile` only by hand — ours: `explore map --use-project` nightly, `explore profile` never — cannot learn from any scheduled run that a column rename has left an override pointing at nothing. The override silently stops clearing a column that no longer exists, the renamed column comes back under the classifier's default flag, and the one place dex says so is a command nothing scheduled calls. The `relationships`-conflict field that 1.11.0 added to `map` (#408) is the model here: the map is where a project-level disagreement gets seen because the map is what runs.

### Proposed solution

Carry the same warnings on `map`'s result: after the datasets are composed, `warnings.extend(_override_mismatches(datasets, config.pii_overrides))` before the `MapResult` is built, so both arms (the exact-entry phrasing `pii_overrides entry '<table.column>' matches no column of <table>` and the pattern-entry phrasing `pii_overrides pattern entry (column_name=…, scope=…) matches no column named …`) reach `map`'s `warnings[]` exactly as they reach `profile`'s. `datasets` and `config` are both in scope at that point. Same text on both commands, so a host matching structurally on the `pii_overrides` prefix reads one contract.

A `MapResult` field naming the orphaned entries (the way `conflicts` names disagreeing declarations) would be better still for a host, but the one-line `warnings` extension is what closes the gap and needs no new shape.

### Alternatives considered

- Scheduling `explore profile` beside the nightly `map`: doubles the billed scan for a warning the map already has every input to produce.
- Recomputing the mismatch in the host from `.dex/config.yml` and the cache's columns: a second copy of dex's rule, one that drifts the first time `_override_mismatches` learns a new entry form (it already has two).
- Reading `notes` for the cleared count: a count that moves is not a name, and it does not move when the orphaned entry was the only one.

### Area

Explore

### dex-core version

1.12.1 (read in the wheel; `map` unchanged in this respect since at least 1.11.0)

catincloudlabs:
## What this changes

`_override_mismatches` warns when an exact `pii_overrides` entry names a column its table does not have, or a pattern entry's scope matches profiled tables and none of them carries the named column. `profile` was its only caller (`explore/commands.py`, the `warnings=` argument of its result). `map` profiles the same datasets but built its own `warnings[]` from carry-forward, `--verify` and overlap warnings alone, so a host that schedules `map` and runs `profile` by hand had no scheduled run that could say a column rename had left an override pointing at nothing. The override silently stopped clearing, the renamed column came back under the classifier's default flag, and the one command that said so was the one nothing scheduled.

### The fix

`map` now extends its warnings with the same call, over the composed set (fresh and carried profiles alike, so a cache hit does not hide it), placed after the overlap warning and before `MapResult` is built. Both entry forms reach `map`'s `warnings[]` with the text `profile` uses, so a host matching on the `pii_overrides` prefix reads one contract. One statement plus a comment in `explore/commands.py`.

### The tests

Three tests on the `tpch_names_duckdb` fixture in `tests/explore/test_explore.py`: an exact entry naming a column the table lacks warns on `map` and clears nothing (the `hosts.name` flag survives in the cache); a pattern entry naming a column no matched table carries warns; an entry naming a real column produces no such warning and clears the column with its `pii_overridden` audit trail. The two positive arms fail on the previous tree, which is the reproducer from the issue.

### Not in this change

A `MapResult` field naming the orphaned entries, the way `conflicts` names disagreeing declarations. The issue calls it the better shape for a host; it is a larger downstream-visible change than the one-line extension that closes the gap, and is left for a maintainer's call. `explore relationships` also profiles the full inventory and builds its own `warnings[]` without this call; it is not the command a host schedules, so it is left as it was rather than widened past the issue.

Under the criterion in `.github/release.yml`, this qualifies as `downstream-visible`: a consumer that stores or compares `map`'s envelope will see warnings it did not before, and only when an override is already orphaned. I cannot apply the label from a fork, so it is stated here.

## Related issues

Closes #448.

## Checklist

- [x] Eval scoring-core tests pass (`uvx pytest evals`): 36 pass locally; the six `evals/tests/test_wrapper.py` cases that monkeypatch `PosixPath` cannot run on Windows (`cannot instantiate 'PosixPath' on your system`), fail identically on `main` here, and are untouched by this change, so the ubuntu CI job is the arbiter for them
- [x] If a safety path is touched, the spine still holds: read-only against data, cost-guarded, PII flagged not surfaced, propose-don't-impose (no safety path is touched; the change adds a warning string and reads no data)
- [x] PR title is ready for release notes; applicable release-note labels and upgrade guidance are included (label stated above, cannot be applied from a fork; no upgrade step)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,6 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
   the `pii_overrides` prefix reads one contract. No new result field: the
   issue names a `MapResult` field for orphaned entries as the better shape
   for a host, and that is a larger downstream-visible change than this one.
-  Downstream-visible in the sense of the release-notes label: a consumer
-  that stores or compares `map`'s envelope will see warnings it did not
-  before, and only when an override is already orphaned.
 
 ## [1.12.1] - 2026-09-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,29 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`explore map` now carries the `pii_overrides` mismatch warning that
+  `explore profile` already emitted** ([#448]). `_override_mismatches` warns
+  when an exact entry names a column its table does not have, or a pattern
+  entry's scope matches profiled tables and none carries the named column;
+  `profile` was its only caller, and `map` built its own `warnings[]` from
+  carry-forward, `--verify` and overlap warnings alone. A host that schedules
+  `map` and runs `profile` by hand therefore had no scheduled run that could
+  say a column rename had left an override pointing at nothing: the override
+  silently stopped clearing, the renamed column came back under the
+  classifier's default flag, and the one command that said so was the one
+  nothing scheduled. `map` now extends its warnings with the same call over
+  the composed set (fresh and carried profiles alike), placed after the
+  overlap warning and before the result is built, so both entry forms reach
+  `map`'s `warnings[]` with the text `profile` uses and a host matching on
+  the `pii_overrides` prefix reads one contract. No new result field: the
+  issue names a `MapResult` field for orphaned entries as the better shape
+  for a host, and that is a larger downstream-visible change than this one.
+  Downstream-visible in the sense of the release-notes label: a consumer
+  that stores or compares `map`'s envelope will see warnings it did not
+  before, and only when an override is already orphaned.
+
 ## [1.12.1] - 2026-09-09
 
 ### Fixed
--- a/packages/dex-core/src/exmergo_dex_core/explore/commands.py
+++ b/packages/dex-core/src/exmergo_dex_core/explore/commands.py
@@ -2329,6 +2329,12 @@ def map(
     )
     if overlap_warning:
         warnings.append(overlap_warning)
+    # The same sentence `profile` emits when a `pii_overrides` entry names a
+    # column the profiled table does not have (issue #448). `map` is the
+    # command a host schedules, so this is where a rename that orphaned an
+    # override gets seen; over the composed set, fresh and carried alike, so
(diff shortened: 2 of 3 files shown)


catincloudlabs:
Self-review before anyone else's, two findings, both addressed in 42c0b98.

1. The comment beside the new statement claims the warning is computed over the composed set so a cache hit does not hide it, and no test asserted that. The scheduled case is exactly the one where it matters: the second nightly `map` re-scans nothing. Added `test_map_still_warns_when_every_profile_is_a_cache_hit`, which runs `map` twice with an orphaned exact entry and checks that the second run, at `profiled_count` 0, carries the same single warning from the carried profiles.

2. The CHANGELOG entry ended with two sentences about the downstream-visible release label. That is a note to the reviewer rather than release history, and it already sits in this body; removed from the entry.

Not changed on purpose: `explore relationships` also builds its own `warnings[]` without this call. It profiles the full inventory too, but it is not the scheduled command the issue is about, so widening past the issue is left as a separate call.

marcociav-exmergo (approved):
LGTM! Thanks @catincloudlabs.
We're going to add this to our next release in an hour or so

PR #469, opened 2026-09-11
SCRUM-1079-maintain verify: report a built relation whose grain is not unique

### LINKED ISSUE #229: `maintain verify`: report a built relation whose grain is not unique
**Depends on:** #224, soft-blocked by **#168**

### Problem

A model whose intended grain is one row per entity, and which has duplicates, is
wrong in a way that propagates: every downstream join fans out and every downstream
sum is inflated. dbt's `unique` test catches this **when someone declared it**,
which in practice is a minority of models.

### Proposal

For each selected model, determine the intended grain (from a declared `unique`
test, a declared primary entity in the semantic layer, or the profiling heuristic)
and verify it holds in the built relation. Report `grain_broken` with the duplicate
count when it does not.

Where the grain could not be determined at all, say so. An unknown grain is a
different and less alarming state than a broken one, and conflating them is how a
finding loses its meaning.

**#168** is a soft blocker: the composite-pair cap can crowd the true grain out of
the candidate set, and a wrong grain produces a wrong verdict here. Worth fixing
first, or at minimum noting the interaction so a composite-grained model is
reported as unknown rather than broken.

### Acceptance

- A model with a declared unique key that has duplicates is reported with a count.
- A model whose grain is a composite the heuristic did not find is reported as
  unknown, never as broken.
- A model with a proven unique grain reports nothing.
- The finding carries the `exact` flag honestly when the verdict rests on an
  approximate distinct count.

emnarahmouni-exmergo:
## Summary
- `maintain verify` now reports a built relation whose grain is not unique: for every selected model, the intended grain is determined from a declared `unique` test, a declared composite `unique_combination_of_columns` (checked only when the model has no single-column test), or a semantic model's declared primary entity (checked only when neither dbt test exists), each verified with a real `exact_distinct_counts`/`distinct_combination_counts` scan against the built relation.
- A model with no declaration at all falls to a free naming heuristic: an `id`/`<entity>_id`-shaped column is checked with an approximate distinct count first (`adapter.column_aggregates`), escalating to an exact scan only when already near-unique (mirroring `explore/profile.py`'s own escalation pattern) — a column nowhere near unique is reported straight from the cheap approximate count instead of paying for an exact scan to confirm what's already obvious.
- A broken grain reports `grain_broken` with the duplicate count; a model whose true grain the heuristic couldn't find (a composite with no declaration, say) is named once in the summary notes as unknown, never reported broken; a proven unique grain reports nothing.
- The finding's `exact` flag is honest: `True` for every declared check and every heuristic check the near-unique escalation reached, `False` only for the one case a verdict rests on the raw approximate count alone.
- Wired into `maintain verify`'s existing suppression machinery (`_grain_contract` wrapper in `commands.py`), consistent with build-status/row-population/column-contract: suppressed the same way when the project doesn't compile, the warehouse is unreachable, or there's no dbt project.

## Acceptance criteria
- [x] A model with a declared unique key that has duplicates is reported with a count.
- [x] A model whose grain is a composite the heuristic did not find is reported as unknown, never as broken.
- [x] A model with a proven unique grain reports nothing.
- [x] The finding carries the `exact` flag honestly when the verdict rests on an approximate distinct count.

## Test plan
- [x] 12 new pure unit tests covering `grain_plan`'s declared-source priority (unique test > composite > primary entity) and scope filtering, plus `grain_findings`'s declared-check, heuristic-escalation, approximate-verdict, disqualified-candidate, unknown, and ambiguous-relation paths
- [x] 4 new end-to-end tests through the real DuckDB `maintain_repo` fixture: the primary-entity path (real semantic YAML fallback), the declared-unique-test path (constructed manifest test node, outranking the primary entity for the same model), the clean/proven-unique path, and the naming-heuristic path (a plain warehouse table with no dbt test or semantic declaration)
- [x] `tests/maintain/test_verify.py` — 77/77 passed
- [x] Full regression sweep: `tests/maintain/ tests/transform/ tests/explore/` — 1849 passed, 30 skipped (pre-existing, unrelated), 0 failed
- [x] `ruff check` / `ruff format` clean on all touched files
- [x] Zero em-dashes in code, tests, and CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,28 @@ tag releases both in lockstep, so entries below are keyed by the engine version.
 
 ## [Unreleased]
 
+### Added
+
+- **`maintain verify` reports a built relation whose grain is not unique**
+  ([#229]). For every selected model, the intended grain is determined from a
+  declared `unique` test, a declared composite `unique_combination_of_columns`
+  (checked only when no single-column test exists for the model), or a
+  semantic model's declared primary entity (checked only when neither dbt
+  test exists), each verified with a real distinct-count scan against the
+  built relation. A model with no declaration at all falls to a free naming
+  heuristic instead: an `id`/`<entity>_id`-shaped column is checked with an
+  approximate distinct count first, escalating to an exact scan only when
+  already close to unique, so a column nowhere near unique is reported
+  straight from the cheap approximate count rather than paying for an exact
+  scan to confirm what is already obvious. A broken grain reports
+  `grain_broken` with the duplicate count; a model whose true grain the
+  heuristic could not find (a composite with no declaration, say) is named
+  once in the summary notes as unknown, never reported broken; a proven
+  unique grain reports nothing. The finding's `exact` flag is honest about
+  which of the two: `True` for every declared check and every heuristic
+  check the near-unique escalation reached, `False` only for the one case a
+  verdict rests on the approximate count alone.
+
 ## [1.12.2] - 2026-09-09
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


marcociav-exmergo (approved):
LGTM

